PR #41, opened 2020-12-21
Feature/13606 vets website linting rules

timwright12:
The linting rules had fallen about 6 months out of sync with vet-website, so just cleaning that up by merged each patch. I had to manually add a couple as well.


billfienberg (approved):
Good call

PR #79, opened 2021-04-09
removed Circle CI

timwright12:


U-DON (approved):

theodur:
Should we also delete `docker-compose.circle.yml` and `circle.eslint.json`?

timwright12:
@theodur yeah nice callout, I'll do that

PR #93, opened 2021-04-29
Run lint/security/unit checks while building v2

jhonnyoddball:


Jenkinsfile
```diff
@@ -17,28 +16,107 @@ node('vetsgov-general-purpose') {
...
 
     try {
       parallel (
         failFast: true,
 
+        buildDev: {
+          if (commonStages.shouldBail()) { return }
+          envName = 'vagovdev'
+
+          shouldBuild = !contentOnlyBuild || envName == params.cmsEnvBuildOverride
```
billfienberg, line 39: I wonder if this logic needs to be changed for the content build.

Jenkinsfile
```diff
@@ -17,28 +16,107 @@ node('vetsgov-general-purpose') {
...
+  stage('Main') {
+    def contentOnlyBuild = params.cmsEnvBuildOverride != 'none'
+    def assetSource = contentOnlyBuild ? ref : 'local'
 
     try {
       parallel (
         failFast: true,
 
+        buildDev: {
+          if (commonStages.shouldBail()) { return }
```
billfienberg, line 36: I wonder if it would be helpful to log why a build stopped. 

For example: 
`echo "Bailing because commonStages.shouldBail() was true"`

Jenkinsfile
```diff
@@ -17,28 +16,107 @@ node('vetsgov-general-purpose') {
...
+          envName = 'vagovdev'
+
+          shouldBuild = !contentOnlyBuild || envName == params.cmsEnvBuildOverride
+          if (!shouldBuild) { return }
```
billfienberg, line 40: I wonder if it would be helpful to log why a build stopped. 

For example: 
`echo "Bailing because of content-only build"`

PR #99, opened 2021-05-03
Add `content-build` repo to `install-repos.sh` script

indytechcook:
Add `content-build` repo to `install-repos.sh` script

### DIFF
--- a/script/install-repos.sh
+++ b/script/install-repos.sh
@@ -10,3 +10,9 @@ if [ ! -d ../vets-website ]; then
 else
   echo "Repo vets-website already cloned."
 fi
+
+if [ ! -d ../content-build ]; then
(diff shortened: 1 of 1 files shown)


script/install-repos.sh
```diff
@@ -9,4 +9,10 @@ if [ ! -d ../vets-website ]; then
   git clone --single-branch --depth 1 https://github.com/department-of-veterans-affairs/vets-website.git ../vets-website
 else
   echo "Repo vets-website already cloned."
-fi
\ No newline at end of file
+fi
+
+if [ ! -d ../content-build ]; then
```
timwright12, line 14: @indytechcook are you seeing an error pop up for this? I wouldn't think we'd need it since it's the current repo

indytechcook: Yeah, that's true.  I had thought I needed it for a build in the CMS and it turns out I don't.

PR #100, opened 2021-05-04
Implement VAMC data layer for links within accordions

brangi:
## Description
Issue: https://github.com/department-of-veterans-affairs/va.gov-team/issues/21556

Screen Shot 2021-05-03 at 7 09 08 PM


src/site/facilities/health_service.drupal.liquid
```diff
@@ -25,8 +25,21 @@
...
-                            {% if facility != empty %}<li class="vads-u-margin-bottom--2">
-                                <a href="{{ facility.entityUrl.path }}">{{ facility.title }}</a></li>{% endif %}
+                            {% if facility != empty %}
+                                <li class="vads-u-margin-bottom--2">
+                                    <a
+                                        onclick="recordEvent(
+                                            {
+                                                'event': 'nav-accordion-link-click',
+                                                'accordion-parent-label': '{{ serviceTaxonomy.name }}',
+                                                'accordion-child-label': '{{ serviceTaxonomy.fieldAlsoKnownAs }}',
```
mpelzsherman, line 35: If `serviceTaxonomy.fieldAlsoKnownAs` is null, do we need to substitute the string "undefined"? Or does that happen automatically?

mpelzsherman (approved):
Please check the comment re: "undefined".

We should probably add some unit tests but that can be done later.

brangi:
> Please check the comment re: "undefined".
> 
> We should probably add some unit tests but that can be done later.

alright, I can check that

timwright12 (approved):

src/site/facilities/health_service.drupal.liquid
```diff
@@ -25,8 +25,21 @@
...
+                                                'accordion-section-label': '{{ sectionName }}'
+
+                                            })"
+                                        href="{{ facility.entityUrl.path }}">{{ facility.title }}
+                                    </a>
+                                </li>
+                            {% endif %}
```
mpelzsherman, line 42: Should we create a tech-debt ticket to convert these to `va-accordion`?

PR #109, opened 2021-05-06
exclude file downloads from certain paths

timwright12:
Should exclude broken files from download


jhonnyoddball (approved):

indytechcook (approved):
I think this is fine as long as we checked a few examples to make sure they weren't in the content build HTML/CSS.

PR #141, opened 2021-05-13
[Preview server] Extract non-node query out of node query

ncksllvn:
## Description
This PR fixes an issue where the preview is very slow and even times out often. It fixes it by extracting the non-node elements (eg sidebars) into its own variable that is cached into memory, so that future node queries don't need to ask for that. 

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/5228
## Testing done
- [x] Confirmed preview server works locally and renders fast by using a variety of NodeIds so that nodes are not cached

## Screenshots
image

image

## Acceptance criteria
- [x] Preview server doesn't time out

## Definition of done

### DIFF
--- a/script/preview.js
+++ b/script/preview.js
@@ -3,10 +3,14 @@
 require('isomorphic-fetch');
 const Raven = require('raven');
 const commandLineArgs = require('command-line-args');
+const fs = require('fs-extra');
 const path = require('path');
 const express = require('express');
 const proxy = require('express-http-proxy');
 const jsesc = require('jsesc');
+const {
+  nonNodeQueries,
+} = require('../src/site/stages/build/drupal/individual-queries');
 const createPipeline = require('../src/site/stages/preview');
 
 const getDrupalClient = require('../src/site/stages/build/drupal/api');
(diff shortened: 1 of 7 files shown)


script/preview.js
```diff
@@ -89,6 +102,59 @@ if (process.env.SENTRY_DSN) {
...
+    const freshNonNodeContent = { data: {} };
+
+    const queries = Object.entries(nonNodeQueries());
+    for (const [queryName, query] of queries) {
+      console.time(queryName);
+
+      // eslint-disable-next-line no-await-in-loop
+      const json = await drupalClient.query({ query });
+      Object.assign(freshNonNodeContent.data, json.data);
+      console.timeEnd(queryName);
```
ncksllvn, line 131: No parallelization - this is slow, but i think that is fine because we def don't want to end up swarming the CMS again.

oseasmoran73 (approved):

kelsonic (approved):

mpelzsherman:
I tested on a few nodes, LGTM. It does seem a little faster. Thanks!

mpelzsherman (approved):

PR #156, opened 2021-05-14
Copy build from Jenkins to Github Action

oseasmoran73:
## Description
As part of ongoing migration to GitHub, following PR addresses the following:
- Create build for respective envs [vagov, vagovstaging, vagovprod]
- Checks if broken links are detected
- Notify channel based on severity

Reference to ticket: https://github.com/department-of-veterans-affairs/va.gov-team/issues/24013

## Testing done

- Tested against github-actions
- Tested against vsp-github-action repo for slack bot
- Verification that communication between slack channel and bot are displaying message

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -206,7 +206,7 @@ jobs:
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
@@ -214,42 +214,42 @@ jobs:
 
       - name: Get Slack app token
         uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           ssm_parameter: /devops/github_actions_slack_socket_token
           env_variable_name: SLACK_APP_TOKEN
 
       - name: Get Slack bot token
         uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           ssm_parameter: /devops/github_actions_slack_bot_user_token
           env_variable_name: SLACK_BOT_TOKEN
           
       - name: Get va-vsp-bot token
         uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           ssm_parameter: /devops/VA_VSP_BOT_GITHUB_TOKEN
           env_variable_name: VA_VSP_BOT_GITHUB_TOKEN
 
       - name: Get Docker token
         uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           ssm_parameter: /devops/dockerhub/vavfsbot_password
           env_variable_name: VA_VFS_BOT_DOCKER_TOKEN
 
       - name: Login to Docker Hub
         uses: docker/login-action@v1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           username: "vavfsbot"
           password: ${{ env.VA_VFS_BOT_DOCKER_TOKEN }}
 
       - name: Checkout VSP actions
         uses: actions/checkout@v2
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         with:
           repository: department-of-veterans-affairs/vsp-github-actions
           ref: refs/heads/main
@@ -260,7 +260,7 @@ jobs:
       # TODO: Change to correct channel_id when ready
       - name: Notify Slack about broken links
         uses: ./.github/actions/vsp-github-actions/slack-socket
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK != '0' || failure() }}
         env:
           SSL_CERT_DIR: /etc/ssl/certs
         with:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -202,6 +202,10 @@ jobs:
         id: get-broken-link-info
         run: node ./script/github-actions/check-broken-links.js ${{ matrix.buildtype }}
 
+      - name: test failure
+        if: ${{ failure() }}
+        run: echo "this still execs"
+
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -232,7 +232,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: department-of-veterans-affairs/vsp-github-actions
-          ref: refs/heads/attachment
+          ref: refs/heads/main
           token: ${{ env.VA_VSP_BOT_GITHUB_TOKEN }}
           persist-credentials: false
           path: ./.github/actions/vsp-github-actions
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -204,7 +204,7 @@ jobs:
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
-        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' || failure() }}
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' || always() }}
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,51 @@ jobs:
...
+          ssm_parameter: /devops/github_actions_slack_socket_token
+          env_variable_name: SLACK_APP_TOKEN
+      - uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
+        with:
+          ssm_parameter: /devops/github_actions_slack_bot_user_token
+          env_variable_name: SLACK_BOT_TOKEN
+      - uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
+        with:
+          ssm_parameter: /devops/VA_VSP_BOT_GITHUB_TOKEN
+          env_variable_name: VA_VSP_BOT_GITHUB_TOKEN
```
U-DON, line 224: Naming each step.

```suggestion
      - name: Get Slack app token
        uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
        with:
          ssm_parameter: /devops/github_actions_slack_socket_token
          env_variable_name: SLACK_APP_TOKEN

      - name: Get Slack bot token
        uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
        with:
          ssm_parameter: /devops/github_actions_slack_bot_user_token
          env_variable_name: SLACK_BOT_TOKEN
          
      - name: Get va-vsp-bot token
        uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
        with:
          ssm_parameter: /devops/VA_VSP_BOT_GITHUB_TOKEN
          env_variable_name: VA_VSP_BOT_GITHUB_TOKEN
```

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,51 @@ jobs:
...
+          SSL_CERT_DIR: /etc/ssl/certs
+        with:
+          slack_app_token: ${{ env.SLACK_APP_TOKEN }}
+          slack_bot_token: ${{ env.SLACK_BOT_TOKEN }}
+          attachments: ${{ steps.get-broken-link-info.outputs.SLACK_ATTACHMENTS }}
+          message: "Broken links detected!"
+          blocks: ${{ steps.get-broken-link-info.outputs.SLACK_BLOCKS }}
+          channel_id: "U01SPJ10WBC"
+
+      - name: Max broken links exceeded
```
U-DON, line 250:
```suggestion
      - name: Fail from broken links
```

U-DON: Actually, could the workflow fail by having `check-broken-links.sh` script exit with error? I think the Slack steps can still run if you check for failure.

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,51 @@ jobs:
...
+        uses: actions/checkout@v2
+        with:
+          repository: department-of-veterans-affairs/vsp-github-actions
+          ref: refs/heads/attachment
+          token: ${{ env.VA_VSP_BOT_GITHUB_TOKEN }}
+          persist-credentials: false
+          path: ./.github/actions/vsp-github-actions
+
+      # TODO: Change to correct channel_id when ready
+      - name: Slack notify
```
U-DON, line 237:
```suggestion
      - name: Notify Slack about broken links
```

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,51 @@ jobs:
         id: get-broken-link-info
         run: node ./script/github-actions/check-broken-links.js ${{ matrix.buildtype }}
 
-      - name: Slack notification
-        run: echo Slack Message ${{ steps.get-broken-link-info.outputs.SLACK_MESSAGE }}. Outputting color for testing ${{ steps.get-broken-link-info.outputs.SLACK_COLOR }}
-        if: ${{ (steps.get-broken-link-info.outputs.SLACK_COLOR != '' || steps.get-broken-link-info.outputs.SLACK_MESSAGE != '') && github.ref == 'refs/heads/master' }} 
+      - name: Configure AWS credentials
```
U-DON, line 205: Everything after this point should probably check if the broken link check failed.

```yml
if: ${{ failure() && ... }}
```

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,51 @@ jobs:
...
+        with:
+          ssm_parameter: /devops/github_actions_slack_bot_user_token
+          env_variable_name: SLACK_BOT_TOKEN
+      - uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
+        with:
+          ssm_parameter: /devops/VA_VSP_BOT_GITHUB_TOKEN
+          env_variable_name: VA_VSP_BOT_GITHUB_TOKEN
+
+      # TODO: Change ref (branch) to main upon merge of branch in vsp-github-action repo
+      - name: Checkout department-of-veterans-affairs/vsp-github-actions
```
U-DON, line 227:
```suggestion
      - name: Checkout VSP actions
```

.github/workflows/continuous-integration.yml
U-DON: The rest of the conditional would probably be:

```yml
(
  steps.get-broken-link-info.outputs.SLACK_ATTACHMENTS != '' ||
  steps.get-broken-link-info.outputs.SLACK_BLOCKS != ''
) && github.ref == 'refs/heads/master'
```

But that's rather verbose to add to every step... Probably want to store that in a variable, if possible.

script/github-actions/check-broken-links.js
```diff
@@ -17,17 +17,17 @@ if (fs.existsSync(reportPath)) {
   const shouldFail =
     brokenLinks.isHomepageBroken ||
     brokenLinks.brokenLinksCount > maxBrokenLinks;
-  const color = shouldFail ? 'danger' : 'warning';
-
+  const color = shouldFail ? '#D33834' : '#FFCC00'; // danger or warning, needs to be in hex
   const heading = `@cmshelpdesk ${brokenLinks.brokenLinksCount} broken links found in the ${envName} build on ${BRANCH_NAME} \n\n${SERVER_URL}\n\n`;
-  const message = `${heading}\n${brokenLinks.summary}`;
+  const slackBlocks = `[{"type": "section","text": {"type": "plain_text","text": "${heading}"}}]`;
```
U-DON, line 22: Just to match the spacing of `slackAttachments`...

```suggestion
  const slackBlocks = `[{"type": "section", "text": {"type": "plain_text", "text": "${heading}"}}]`;
```

U-DON (commented):
Mostly nitpicks, but the main thing would be to set the conditional on steps after the broken link check. Would be great if we can make it less painful to read/write that conditional too...

script/github-actions/check-broken-links.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-console */
+// https://cms-nc3gdbj3c2p9pizhf1sm8czjvtfeg1ik.demo.cms.va.gov -- Tugboat CMS with 2 broken links
```
oseasmoran73, line 2: Not sure where to store this information, but figured this would be a nice spot for the time being

script/github-actions/check-broken-links.js
```diff
@@ -37,7 +41,17 @@ if (fs.existsSync(reportPath)) {
     // continue merging.
     return;
   }
-  if (color === 'danger') {
+
+  /*
+   * Only emit this variable if ran against master branch or during Content Release.
+   * Meets the following condition: blocks & attachments & IS_PROD_BRANCH
+   */
+  console.log(`::set-output name=NOTIFY_SLACK::1`);
```
oseasmoran73, line 49: Not inside `if (shouldFail)` because we still want to emit notification on warnings, not just failures

.github/workflows/continuous-integration.yml
oseasmoran73: I consolidated all of those conditions into `steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1'` where 1 is true, 0 is false. 

Additionally max links were to be exceeded it would throw an Error so I added `failure()` to still emit slack message.

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,69 @@ jobs:
...
+        with:
+          ssm_parameter: /devops/dockerhub/vavfsbot_password
+          env_variable_name: VA_VFS_BOT_DOCKER_TOKEN
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' || failure() }}
+        with:
+          username: "vavfsbot"
+          password: ${{ env.VA_VFS_BOT_DOCKER_TOKEN }}
```
U-DON, line 246: Is this something we should do in general (other jobs)?

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,69 @@ jobs:
...
+        with:
+          ssm_parameter: /devops/VA_VSP_BOT_GITHUB_TOKEN
+          env_variable_name: VA_VSP_BOT_GITHUB_TOKEN
+
+      - name: Get Docker token
+        uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' || failure() }}
+        with:
+          ssm_parameter: /devops/dockerhub/vavfsbot_password
+          env_variable_name: VA_VFS_BOT_DOCKER_TOKEN
```
U-DON, line 239:
```suggestion
          env_variable_name: VA_VFS_BOT_DOCKERHUB_PASSWORD
```

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,69 @@ jobs:
         id: get-broken-link-info
         run: node ./script/github-actions/check-broken-links.js ${{ matrix.buildtype }}
 
-      - name: Slack notification
-        run: echo Slack Message ${{ steps.get-broken-link-info.outputs.SLACK_MESSAGE }}. Outputting color for testing ${{ steps.get-broken-link-info.outputs.SLACK_COLOR }}
-        if: ${{ (steps.get-broken-link-info.outputs.SLACK_COLOR != '' || steps.get-broken-link-info.outputs.SLACK_MESSAGE != '') && github.ref == 'refs/heads/master' }} 
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' || failure() }}
```
U-DON, line 207: I see that we're notifying not only on errors but also for warnings about broken links.

It seems like we can eliminate the `failure()` condition. If failures cause the step to get skipped without a job status check, maybe we can try `always() && steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1'`. [Docs seem to suggest we might need the job status check](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions).

> If your `if` expression does not contain any of the status functions it will automatically result with `success()`. 

Unless we're intentionally notifying on failure in general, but the Slack message is currently tailored for broken links.

```suggestion
        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' }}
```

U-DON: In other words, we don't run the rest of these steps whenever there's any general failure (even unrelated to broken links). But we do want to run them if we've determined that we need to notify Slack, failure or not.

.github/workflows/continuous-integration.yml
```diff
@@ -201,6 +202,69 @@ jobs:
...
+          env_variable_name: SLACK_BOT_TOKEN
+          
+      - name: Get va-vsp-bot token
+        uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
+        if: ${{ steps.get-broken-link-info.outputs.NOTIFY_SLACK == '1' || failure() }}
+        with:
+          ssm_parameter: /devops/VA_VSP_BOT_GITHUB_TOKEN
+          env_variable_name: VA_VSP_BOT_GITHUB_TOKEN
+
+      - name: Get Docker token
```
U-DON, line 234:
```suggestion
      - name: Get Docker Hub password
```

.github/workflows/continuous-integration.yml
oseasmoran73: Only if using docker/dockerhub from my understanding. Had to add due to Slack Bot limited unauthenticated pull [slack](https://dsva.slack.com/archives/C01V94K0QNS/p1621978945065700)

.github/workflows/continuous-integration.yml
oseasmoran73: I don't think `always()` is the right approach. It will work for all token/authentication portion, but it causes a misleading message when it gets specifically to slack notification.

Currently, the slack bot requires the following `message: "Broken links detected!"` where "Broken links detected" can be anything. The Slack Bot first checks if `blocks` has been passed and if it doesn't it defaults to `message` (otherwise it acts as backup text -- usually in form of pop up what I have seen). 

If there are no warnings or error, it will still fire to the channel that broken links are detected (or any message passed). With failure, it will ensure nothing gets fired, where as always, the slackbot will need to send something (which my thought process is no news is good news in this case). 

Thoughts?

.github/workflows/continuous-integration.yml
U-DON: In the same thread, I see that we won't need to log in anymore. Switching to the new image should bypass the throttling issue.

.github/workflows/continuous-integration.yml
oseasmoran73: Tested workflow and working as intended

U-DON (approved):

PR #165, opened 2021-05-19
Fix for links must have discernible text on TOC in build

timwright12:
## Description
[Nightwatch sitemap a11y scan ](http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fcontent-build/detail/feature%2Fnightwatch-content-a11y-tests/9/pipeline/)is picking up an error for `Links must have discernible text` on https://www.va.gov/albany-health-care/policies/

I tracked this down to a rogue H2 coming from the CMS, so I added a check in the TOC generator to test for the presence of and ID before creating the link.

### Before
before

### After
after


Cgfor3 (approved):
LGTM

PR #168, opened 2021-05-20
Added tests for 4 liquid.filter functions

gescobardev:
## Description
Created tests for the following functions:

1. `benefitTerms()`
2. `findCurrentPathDepth()`
3. `hashReference()`
4. `fileExt()`

Additionally, I combed through `liquid.js` and found that the following tests are not currently being utilized and have removed them:

- `humanizeTime()`
- `unixFromDate()`
- `modulo()`
- `breakTerms()`
- `facilityIds()`
- `widgetFacilitiesList()`
- `eventSorter()`
- `getPagerPage()`
- `isPitt()`
- `isPastDate()`

## Acceptance criteria
- [ ] All tests must pass
- [ ] Please verify that the functions I listed above and removed, are indeed not in use.

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -266,12 +266,12 @@ module.exports = function registerFilters() {
   };
 
   liquid.filters.hashReference = str => {
-    if (str === null) return null;
+    if (!str) return null;
     return str
       .toString()
       .toLowerCase()
-      .split(' ')
-      .join('-');
+      .trim()
+      .replace(/\s+/g, '-');
   };
 
   // We might not need this filter, refactor
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -22,8 +22,6 @@ module.exports = function registerFilters() {
   liquid.filters.humanizeDate = dt =>
     moment(dt, 'YYYY-MM-DD').format('MMMM D, YYYY');
 
-  liquid.filters.humanizeTime = dt => moment(dt).format('LT');
-
   liquid.filters.humanizeTimestamp = dt =>
     moment.unix(dt).format('MMMM D, YYYY');
 
@@ -154,8 +152,6 @@ module.exports = function registerFilters() {
       .replace(/PM/g, 'p.m.');
   };
 
-  liquid.filters.unixFromDate = data => new Date(data).getTime();
-
   liquid.filters.currentTimeInSeconds = () => {
     const time = new Date();
     return Math.floor(time.getTime() / 1000);
@@ -164,8 +160,6 @@ module.exports = function registerFilters() {
   liquid.filters.numToWord = numConvert => converter.toWords(numConvert);
 
   liquid.filters.jsonToObj = jsonString => JSON.parse(jsonString);
-
-  liquid.filters.modulo = item => item % 2;
 
   liquid.filters.genericModulo = (i, n) => i % n;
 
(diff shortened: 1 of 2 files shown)


src/site/filters/liquid.js
```diff
@@ -174,11 +174,14 @@ module.exports = function registerFilters() {
 
   liquid.filters.fileSize = data => `${(data / 1000000).toFixed(2)}MB`;
 
-  liquid.filters.fileExt = data =>
-    data
+  liquid.filters.fileExt = data => {
+    if (data === null) return null;
```
mpelzsherman, line 178: Thanks for adding this! I think it's even better to check for any "falsy" value for data, e.g. `undefined`.
We usually just do `if (!data)` for these sorts of checks. Another option might be to see if `data` has a `toString()` property. EIther way works for me.

gescobardev: That makes sense and I like `if (!data)`.
I will update

src/site/filters/liquid.unit.spec.js
```diff
@@ -7,6 +7,176 @@ import featuredContentData from '../layouts/tests/vet_center/fixtures/featuredCo
 
 registerFilters();
 
+describe('benefitTerms', () => {
+  it('returns null when null is passed', () => {
+    expect(liquid.filters.benefitTerms(null)).to.eq(null);
+  });
```
mpelzsherman, line 13: I prefer to add a space between each test case. Could you add those please? (Sorry, I know it's a pain. 🙏 )

src/site/filters/liquid.unit.spec.js
```diff
@@ -7,6 +7,176 @@ import featuredContentData from '../layouts/tests/vet_center/fixtures/featuredCo
...
+    expect(liquid.filters.hashReference(null)).to.eq(null);
+  });
+  it('returns an empty string when an empty array is passed', () => {
+    expect(liquid.filters.hashReference([])).to.eq('');
+  });
+  it('returns string with spaces replaced by "-" ', () => {
+    expect(liquid.filters.hashReference('testing one two three')).to.eq(
+      'testing-one-two-three',
+    );
+  });
```
mpelzsherman, line 162: Does it collapse multiple spaces into one space? Does it matter? Might want to check with @ethanteague ?

src/site/filters/liquid.unit.spec.js
```diff
@@ -7,6 +7,176 @@ import featuredContentData from '../layouts/tests/vet_center/fixtures/featuredCo
...
+  it('returns null when undefined is passed', () => {
+    expect(liquid.filters.fileExt(undefined)).to.eq(null);
+  });
+  it('returns null when empty string is passed', () => {
+    expect(liquid.filters.fileExt('')).to.eq(null);
+  });
+  it('returns an empty string when an empty array is passed', () => {
+    expect(liquid.filters.fileExt([])).to.eq('');
+  });
```
mpelzsherman, line 177: I'm not seeing a happy path test here? How about `foo.txt`? Also try `foo.bar.txt`.

src/site/filters/liquid.unit.spec.js
gescobardev: No worries! I'll add them

src/site/filters/liquid.unit.spec.js
gescobardev: Thank you for catching that!
I have now added spacing between tests, removed the list of unused liquid filter functions and added 5 more tests to `fileExt()`

mpelzsherman (changes requested):
Please add a test for `hashReference` to confirm that it handles multiple spaces correctly.

I think the correct behavior is to condence multiple spaces into a single dash. So:

`"foo<space><space>bar" => "foo-bar"`

Thanks!

gescobardev:
@mpelzsherman I have added about 4 more tests to `hashReference`

mpelzsherman (approved):

PR #171, opened 2021-05-20
Vet center - "Nearby locations" should be retrieved from Lighthouse

brangi:
resolves

### DIFF
--- a/src/site/layouts/vet_center_locations_list.drupal.liquid
+++ b/src/site/layouts/vet_center_locations_list.drupal.liquid
@@ -48,16 +48,7 @@
                     {% endif %}
 
                     {% if fieldNearbyVetCenters.length > 0 %}
-                        <h2 class="vads-u-font-size--xl vads-u-margin-top--3 medium-screen:vads-u-margin-top--5 vads-u-margin-bottom--2p5
-                 medium-screen:vads-u-margin-bottom--3" id="other-near-locations">
-                            Other nearby Vet Centers
-                        </h2>
-                        {% for entityVetCenter in fieldNearbyVetCenters %}
-                            {% include "src/site/includes/vet_center_address_phone_image.liquid" with vetCenter = entityVetCenter.entity %}
-                        {% endfor %}
-                        {% for entityVetCenter in fieldNearbyMobileVetCenters %}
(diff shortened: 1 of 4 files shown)


src/site/stages/build/drupal/graphql/vetCenterLocations.graphql.js
```diff
@@ -17,6 +17,8 @@ fragment vetCenterLocationsFragment on NodeVetCenterLocationsList {
   fieldNearbyVetCenters {
     entity {
       ... on NodeVetCenter {
+        fieldFacilityLocatorApiId
+        entityPublished
```
mpelzsherman, line 21: I don't see where `entityPublished` is being used.

src/site/layouts/vet_center_locations_list.drupal.liquid
```diff
@@ -48,16 +48,10 @@
...
-                        {% for entityVetCenter in fieldNearbyVetCenters %}
-                            {% include "src/site/includes/vet_center_address_phone_image.liquid" with vetCenter = entityVetCenter.entity %}
-                        {% endfor %}
-                        {% for entityVetCenter in fieldNearbyMobileVetCenters %}
-                            {% include "src/site/includes/vet_center_address_phone_image.liquid" with vetCenter = entityVetCenter.entity %}
-                        {% endfor %}
+                        {% include 'src/site/includes/vet_center_nearby.liquid' with
+                            nearbyVetCenters = fieldNearbyVetCenters
+                            mainVetCenterPhone = fieldOffice.entity.fieldPhoneNumber
+                        %}
```
mpelzsherman, line 54: Looks like we are always using the vet_center_nearby widget, regardless of whether the data is published in Drupal. Is that OK?

brangi: yes, it's easier to manipulate the array and create the logic inside the widget than here

src/site/stages/build/drupal/graphql/vetCenterLocations.graphql.js
brangi: https://github.com/department-of-veterans-affairs/vets-website/pull/17297

mpelzsherman (approved):
I’ll approve the PR but eventually I’d like to see those redundant templates removed so we just have one way of rendering Vet Centers.

PR #172, opened 2021-05-21
Added Liquid filter to check for and updated duplicate ID values in the WYSIWYG editor

timwright12:
## Description

We have a reported section 508 scan issue of a duplicate ID value the in WYSIWYG content output on https://va.gov/health-care/health-needs-conditions/military-sexual-trauma/ - this PR adds a liquid filter to check for duplicate IDs in the WYSIWYG content and makes them unique.

## Testing Done

Successful pipeline build and spot checking content pages.

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -763,11 +763,12 @@ module.exports = function registerFilters() {
     let searchedIdValue;
     let idList;
 
-    try {
+    // If there's no content, don't run this because Cheerio will throw and error
+    if (replaced !== null) {
       // Load the DOM fragement into Cheerio so we can parse it.
       const $ = cheerio.load(replaced);
 
-      // Loop through all the IDs.
+      // Loop through all the IDs in the WYSIWYG output.
       $('[id]').each(function() {
         searchedIdValue = $(this).attr('id');
         idList = $(`[id="${searchedIdValue}"]`);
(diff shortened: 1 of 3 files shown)


timwright12:
Seeing if we can get this fixed in the CMS before adding the filter

Cgfor3 (approved):
👍

zacharymorel (approved):

src/site/filters/liquid.js
```diff
@@ -753,6 +755,40 @@ module.exports = function registerFilters() {
...
+          });
+          // Return the cleaned HTML.
+          replaced = $.html();
+        }
+      });
+    } // if()
+
+    return replaced;
+  };
+
```
mpelzsherman, line 791: Please add a test for any new filter functions (see `src/site/filters/liquid.unit.spec.js`).

mpelzsherman:
Based on the Slack convo, it sounds like the team correctly concluded that this PR isn’t needed. We appreciate the initiative here, but in general, I think issues like this should probably go through our normal grooming process.

mpelzsherman (changes requested):
In general, we should not be doing this sort of complex logic to fix issues with CMS data.

I think we should close this PR.

src/site/filters/liquid.js
timwright12: We're only merging here if this become a recurring problem in the WYSIWYG, but CMS has opened a ticket for it so I'm pulling this back into draft for the time being https://github.com/department-of-veterans-affairs/va.gov-cms/issues/5421

PR #198, opened 2021-05-27
#23667: Increase tinyliquid timeout to 20 minutes

alexandec:
## Description
We use tinyliquid to render templates into HTML. Tinyliquid's default 2-minute timeout is insufficient with high node counts, likely because metalsmith runs many tinyliquid engines in parallel. At 48k nodes the following error would cause the build to fail both locally and on CI:

```
Step 21 of 54 start: Apply layouts C
/Users/chris/Documents/code/content-build/src/site/stages/build/index.js:243
    if (err) throw err;
             ^
Error: Timeout.
    at Timeout.<anonymous> (/Users/chris/Documents/code/content-build/node_modules/tinyliquid/lib/index.js:79:16)
```

Increasing the timeout to 20 minutes resolves the error.

## Testing done
- Successful 48k node content build (no timeout observed)
- Instrument tinyliquid to output timeout value, ensuring it is set correctly
- No regressions in existing build

## Acceptance criteria
- [x] Current and 48k node content builds are successful (no timeouts)

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -7,6 +7,10 @@ const moment = require('moment-timezone');
 // Relative imports.
 const phoneNumberArrayToObject = require('./phoneNumberArrayToObject');
 
+// The default 2-minute timeout is insufficient with high node counts, likely
+// because metalsmith runs many tinyliquid engines in parallel.
+const TINYLIQUID_TIMEOUT_MINUTES = 20;
+
 function getPath(obj) {
   return obj.path;
 }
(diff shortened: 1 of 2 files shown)


jhonnyoddball (approved):
LGTM

kelsonic (approved):
LGTM, one note: we should add a unit test for `liquid.run` [here](https://github.com/department-of-veterans-affairs/content-build/blob/6945cd46a56174c8e53182b9adac935bc4df4d8a/src/site/filters/liquid.unit.spec.js)

mpelzsherman (approved):

PR #179, opened 2021-05-25
Ensure pagination does not show when only 1 page

kelsonic:
# Description

**Issue:** https://github.com/department-of-veterans-affairs/va.gov-team/issues/18571

This PR tweaks the pagination template to **not show when there is only 1 page.**

## Screenshots

**Before:** https://www.va.gov/resources/tag/claims-and-appeals-status/

www va gov_resources_tag_claims-and-appeals-status_

**After:**

www va gov_resources_tag_claims-and-appeals-status_ (1)

## Acceptance Criteria

- [x] Do not show the pagination template when it only has 1 page

PR #200, opened 2021-05-28
Revert "Ensure pagination does not show when only 1 page"

kelsonic:
Reverts department-of-veterans-affairs/content-build#179


zacharymorel (approved):
Why the reversion?

PR #205, opened 2021-06-01
Local builds do not re-download cached assets 

mpelzsherman:
## Description

Keep Drupal assets in `.drupalAssets`. Do not delete cache between builds. Do not re-download assets if already in cache. If desired, we can add a yarm command to clear the asset cache.

## Acceptance criteria
- [x] Local builds do not re-download cached assets 

## Testing done
Ran `yarn build --pull-drupal` twice.

Assets downloads went from ~20 minutes to 102ms for the 2nd run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ src/site/tests/html/
 # Files to ignore
 ._*
 .cache
+.drupalAssets
 .config
 .DS_Store
 .env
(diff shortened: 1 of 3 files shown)


timwright12 (approved):
Checked locally and the s3 upload, seem like it works as expected

PR #215, opened 2021-06-03
add lang attribute to tableOfContents links

youngfreezyVA:
this PR adds the lang attribute to tableOfContents links for screen readers


src/site/stages/build/plugins/modify-dom/add-id-to-subheadings.js
```diff
@@ -37,7 +37,9 @@ function createUniqueId(headingEl, headingOptions) {
 module.exports = {
   modifyFile(fileName, file) {
     let idAdded = false;
-
+    const langAttribute = localStorage.getItem('redux')
```
youngfreezyVA, line 40: moved this PR to draft. the `adaptLinksWithLangCode` function should already be taking care of this i realized, but stumbled upon a way to access the language code inside a drupal template

PR #224, opened 2021-06-07
#25074: Add manual garbage collection at regular interval

alexandec:
## Description
Parts of the content build rapidly allocate and deallocate large amounts of data. V8's automatic garbage collection handles this well in most cases, but it can fall behind. This causes spikes in heap and non-heap (buffer) allocation, resulting in very large peak memory use at high node counts.

To address the problem, this PR:
- triggers manual garbage collection at a 10-second interval to keep memory use in check, particularly during the Apply Layouts steps
- adds 1ms pauses to allow garbage collection during two especially memory-intensive steps
- adds logging for total memory use (RSS) to the log, helping track changes  over time

See [#25074](https://app.zenhub.com/workspaces/vsp---frontend-tools-5fc9325744944e0015ed1861/issues/department-of-veterans-affairs/va.gov-team/25074) and [#24022](https://github.com/department-of-veterans-affairs/va.gov-team/issues/24022) for more information and benchmarks.

## Testing done
- Benchmark memory use locally and on CI with a variety of GC intervals
- Manual checks to ensure no regressions

## Screenshots
- See above tickets for graphs, memory use logs, and other benchmarks

## Acceptance criteria
- [x] Manual garbage collection is performed at a regular interval
- [x] Total memory use (RSS) is logged to allow monitoring over time

## Definition of done

### DIFF
--- a/src/site/stages/build/index.js
+++ b/src/site/stages/build/index.js
@@ -271,10 +271,10 @@ function build(BUILD_OPTIONS) {
       // If this isn't a watch, just output the normal "end of build" information
       if (global.verbose) {
         smith.printSummary();
+        smith.printPeakMemory();
       }
 
       smith.endGarbageCollection();
-      smith.printPeakMemory();
 
       console.log('The Metalsmith build has completed.');
 
--- a/src/site/stages/build/index.js
+++ b/src/site/stages/build/index.js
@@ -75,6 +75,10 @@ function build(BUILD_OPTIONS) {
 
   registerLiquidFilters();
 
+  // Start manual garbage collection to limit large spikes in memory use.
+  // This prevents running out of memory on CI.
+  smith.startGarbageCollection();
+
   // Set up Metalsmith. BE CAREFUL if you change the order of the plugins. Read the comments and
   // add comments about any implicit dependencies you are introducing!!!
   //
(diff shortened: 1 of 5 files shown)


src/site/stages/build/silversmith.js
```diff
@@ -76,45 +98,80 @@ module.exports = () => {
...
       });
   };
 
+  smith.startGarbageCollection = function startGarbageCollection() {
+    if (global.gc) {
+      garbageCollectionInterval = setInterval(() => {
+        const memBefore = process.memoryUsage();
+        global.gc();
+        const memAfter = process.memoryUsage();
+        printGarbageCollectionStats(memBefore, memAfter);
```
theodur, line 135: Since we're using manual garbage collection in the default `build` command, should we only output garbage collection stats if `global.verbose` is true, to be consistent with when we output memory related stats?

src/site/stages/build/index.js
```diff
@@ -265,6 +272,10 @@ function build(BUILD_OPTIONS) {
       if (global.verbose) {
         smith.printSummary();
       }
+
+      smith.endGarbageCollection();
+      smith.printPeakMemory();
```
theodur, line 277: Same here, should we only output this if `global.verbose` is enabled?

theodur (approved):

src/site/stages/build/silversmith.js
alexandec: Good point, thanks! I made that change, and the one you mentioned below as well

PR #227, opened 2021-06-08
Copy archive step to GitHub Actions workflow

oseasmoran73:
## Description

Consider making the correction to compress the archive when adapting this to GHA.

Reference [24188](https://app.zenhub.com/workspaces/vsp---frontend-tools-5fc9325744944e0015ed1861/issues/department-of-veterans-affairs/va.gov-team/24188)

## Acceptance criteria
- [x] The archive step is in the GitHub Actions workflow after prearchive and before deploy.
- [x] Archives for each build type are uploaded to S3 at the paths corresponding to the ref and build type.
- [x] Each archive has a BUILD.txt with the correct information.

## Definition of done


.github/workflows/continuous-integration.yml
```diff
@@ -322,14 +332,69 @@ jobs:
...
-        run: tar -zxvf ${{ matrix.buildType }}.tar.gz
+        run: tar -jxvf build/${{ matrix.buildType }}.tar.bz2
 
       - name: Prearchive
         run: node ./script/prearchive.js --buildtype=${{ matrix.buildtype }}
+
+      - name: Compress ${{ matrix.buildtype }} build w/ urls updated
+        run: tar -jcvf ${{ matrix.buildType }}.tar.bz2 build/${{ matrix.buildType }}
+
+      - name: Upload build
```
oseasmoran73, line 353: With prearchive steps and urls updated, need to reupload to artifact so archive can use correct paths

.github/workflows/continuous-integration.yml
```diff
@@ -205,16 +205,26 @@ jobs:
...
+          BRANCH_NAME=$(echo "${GITHUB_REF#refs/heads/}")
+          RUN_ID=${{ github.run_id }}
+          REF=${{ github.sha }}
+          BUILDTIME=$(date +%s)
+          EOF
+
       - name: Compress ${{ matrix.buildtype }} build
         if: ${{ github.ref == 'refs/heads/master' && success() }}
-        run: tar -zcvf ${{ matrix.buildType }}.tar.gz build
+        run: tar -jcvf ${{ matrix.buildType }}.tar.bz2 build/${{ matrix.buildType }}
```
oseasmoran73, line 220: Changed what folder got compressed to play nicely w/ prearchive & archive

.github/workflows/continuous-integration.yml
```diff
@@ -205,16 +205,26 @@ jobs:
...
 
       - name: Upload build
         uses: actions/upload-artifact@v2
         if: ${{ github.ref == 'refs/heads/master' && success() }}
         with:
           name: ${{ matrix.buildtype }}
-          path: ${{github.workspace}}/content-build/${{ matrix.buildType }}.tar.gz
+          path: ${{ github.workspace }}/content-build/${{ matrix.buildType }}.tar.bz2
```
U-DON, line 227: I'm inclined to use relative paths from the working directory where possible. I think that would help narrow down the scope in our own mental models when thinking about the operations taking place. Like this could just be:

```suggestion
          path: ${{ matrix.buildType }}.tar.bz2
```

.github/workflows/continuous-integration.yml
```diff
@@ -322,14 +332,69 @@ jobs:
...
         run: node ./script/prearchive.js --buildtype=${{ matrix.buildtype }}
+
+      - name: Compress ${{ matrix.buildtype }} build w/ urls updated
+        run: tar -jcvf ${{ matrix.buildType }}.tar.bz2 build/${{ matrix.buildType }}
+
+      - name: Upload build
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.buildtype }}
+          path: ${{ github.workspace }}/content-build/${{ matrix.buildType }}.tar.bz2
```
U-DON, line 357:
```suggestion
          path: ${{ matrix.buildType }}.tar.bz2
```

.github/workflows/continuous-integration.yml
```diff
@@ -205,16 +205,26 @@ jobs:
...
       - name: Compress ${{ matrix.buildtype }} build
         if: ${{ github.ref == 'refs/heads/master' && success() }}
-        run: tar -zcvf ${{ matrix.buildType }}.tar.gz build
+        run: tar -jcvf ${{ matrix.buildType }}.tar.bz2 build/${{ matrix.buildType }}
 
       - name: Upload build
         uses: actions/upload-artifact@v2
         if: ${{ github.ref == 'refs/heads/master' && success() }}
         with:
           name: ${{ matrix.buildtype }}
```
U-DON, line 226: This should probably have the suffix. When I was testing in `vets-website`, I found it was nice to have it automatically have the right extension when I downloaded the artifact. I have that change coming up over there too.

```suggestion
          name: ${{ matrix.buildtype }}.tar.bz2
```

.github/workflows/continuous-integration.yml
```diff
@@ -322,14 +332,69 @@ jobs:
...
       - name: Prearchive
         run: node ./script/prearchive.js --buildtype=${{ matrix.buildtype }}
+
+      - name: Compress ${{ matrix.buildtype }} build w/ urls updated
+        run: tar -jcvf ${{ matrix.buildType }}.tar.bz2 build/${{ matrix.buildType }}
+
+      - name: Upload build
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.buildtype }}
```
U-DON, line 356: I'm wary about uploading to the same artifact.

https://github.com/actions/upload-artifact#uploading-to-the-same-artifact

> Warning: Be careful when uploading to the same artifact via multiple jobs as artifacts may become corrupted

> Each artifact behaves as a file share. Uploading to the same artifact multiple times in the same workflow can overwrite and append already uploaded files

U-DON: I wonder if the `prearchive` step should just be rolled into the `build` step. And since the E2E tests will expect a build with the non-prearchived URLs, maybe the `build` step could actually generate two artifacts when `matrix.buildType == 'vagovprod'`. One artifact could just be `vagovprod.tar.bz2` and the other could be `test-build.tar.bz2` or something like that.

U-DON (commented):
This looks good so far! I actually have been wondering ahead of this if we could reduce the number of times we pack and unpack when passing the build between jobs.

I mainly want to point your attention to the part about combining the prearchive step into the build step. And if we go with that, the `vagovprod` build step could generate an extra artifact with an un-prearchived build just for the E2E tests.

.github/workflows/continuous-integration.yml
U-DON: This would have two benefits:
1. Avoids uploading to the same path already created artifact. Even if we tried to avoid name collision by renaming, having more artifacts adds noise.
2. Saves time by only compressing and decompressing once in one job. The pre-prearchived builds aren't needed anywhere else, except for `vagovprod`, which needs to send a pre-prearchived version of itself to E2E tests.

oseasmoran73: Yeah that was my big worry as well, but did not like the idea of having something like `vagov.build.tar.bz2` and `vagov.prearchive.tar.bz2`. 

I would actually be in favor as I initially when writing this PR down seemed to make sense to combine prearchive and build. The compress/decompress seemed like a lot of overhead.

So I will change the following:
1. Get rid of prearchive step and combine with build
2. Create different tarball for prearchive URLS and for E2E

👍

.github/workflows/continuous-integration.yml
```diff
@@ -270,66 +296,48 @@ jobs:
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-gov-west-1
+          role-to-assume: ${{ env.AWS_FRONTEND_NONPROD_ROLE }}
+          role-duration-seconds: 900
+          role-session-name: vsp-frontendteam-githubaction
 
-      - name: Extract ${{ matrix.buildtype }}
-        run: tar -zxvf ${{ matrix.buildType }}.tar.gz
+      - name: ls
+        run: ls -la
```
U-DON, line 340: Do we still need this? Seems like this was for debugging?

.github/workflows/continuous-integration.yml
```diff
@@ -205,16 +205,42 @@ jobs:
...
+          RUN_ID=${{ github.run_id }}
+          REF=${{ github.sha }}
+          BUILDTIME=$(date +%s)
+          EOF
+
+      - name: Compress vagovprod build for E2E
+        if: ${{ github.ref == 'refs/heads/master' && matrix.buildType == 'vagovprod' }}
+        run: tar -jcvf test-build.tar.bz2 build/vagovprod
+
+      - name: Upload build for E2E
```
U-DON, line 222:
```suggestion
      - name: Upload artifact vagovprod build for E2E
```

.github/workflows/continuous-integration.yml
```diff
@@ -205,16 +205,42 @@ jobs:
...
-      - name: Upload build
+      - name: Prearchive
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: node ./script/prearchive.js --buildtype=${{ matrix.buildtype }}
+
+      - name: Compress ${{ matrix.buildtype }} build w/ urls updated
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: tar -jcvf ${{ matrix.buildType }}.tar.bz2 build/${{ matrix.buildType }}
+
+      - name: Upload ${{ matrix.buildtype }} for archive
```
U-DON, line 238:
```suggestion
      - name: Upload artifact ${{ matrix.buildtype }} build for archive
```

.github/workflows/continuous-integration.yml
```diff
@@ -205,16 +205,42 @@ jobs:
...
+          name: test-build.tar.bz2
+          path: content-build/test-build.tar.bz2
+          retention-days: 1
 
-      - name: Upload build
+      - name: Prearchive
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: node ./script/prearchive.js --buildtype=${{ matrix.buildtype }}
+
+      - name: Compress ${{ matrix.buildtype }} build w/ urls updated
```
U-DON, line 234: Not sure if prearchive will always just be updating URLs. Can keep it general, I suppose.

```suggestion
      - name: Compress prearchived ${{ matrix.buildtype }} build
```

U-DON (approved):
Looks good! All I have are nits about step names.

.github/workflows/continuous-integration.yml
oseasmoran73: No we do not and it was for debugging 😅  removing in next commit!

PR #229, opened 2021-06-08
Fix - On this page navigation should have a unique accessible label

brangi:
Resolves - https://github.com/department-of-veterans-affairs/va.gov-team/issues/24998

### DIFF
--- a/src/site/layouts/vet_center.drupal.liquid
+++ b/src/site/layouts/vet_center.drupal.liquid
@@ -19,7 +19,7 @@
                         </div>
                     {% endif %}
 
-                    <nav id="table-of-contents">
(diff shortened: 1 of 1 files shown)


mpelzsherman (approved):
This satisfies the AC. However, instead of fixing it in just this one place, we should address this issue everywhere on the site.

Instead, we should fix it in this plugin: src/site/stages/build/plugins/modify-dom/add-id-to-subheadings.js. But I'm ok with doing that in a new ticket.

P.S. @cvalarida @bkjohnson - The component-library version of this widget, https://design.va.gov/storybook/?path=/docs/components-va-on-this-page--default
needs to add  `aria-labelledby` tags and analytics support.

PR #231, opened 2021-06-09
add aria-labelledby="on-this-page" to on this page elements

youngfreezyVA:
adds `aria-labelledby="on-this-page"` to on this page elements
resolves this accessibility bug: https://github.com/department-of-veterans-affairs/va.gov-team/issues/25830


mpelzsherman (approved):
At some point we should convert all of these to the new design system component.

PR #253, opened 2021-06-14
News release page now preview-able

gescobardev:
## Description
Issue is linked below
To preview use - http://localhost:3002/preview?nodeId=2805

## Acceptance criteria
- [ ] All tests pass

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -778,7 +778,11 @@ module.exports = function registerFilters() {
 
   //* Sorts event dates (fieldDatetimeRangeTimezone) starting with the most upcoming event.
   //* Also sorts press releases (fieldReleaseDate) from newest to oldest.
-  liquid.filters.eventOrPressReleasesDateSorter = (dates, dateKey, reverse) => {
+  liquid.filters.sortByDateKey = (
+    dates,
+    dateKey = 'fieldDatetimeRangeTimezone',
+    reverse = false,
+  ) => {
     if (!dates) return null;
     return dates.sort((a, b) => {
       const start1 = moment(a[dateKey].value);
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -776,13 +776,15 @@ module.exports = function registerFilters() {
     });
   };
 
-  //* Sorts event dates (fieldDatetimeRangeTimezone) from newest to oldest.
-  liquid.filters.eventDateSorter = dates => {
+  //* Sorts event dates (fieldDatetimeRangeTimezone) starting with the most upcoming event.
+  //* Also sorts press releases (fieldReleaseDate) from newest to oldest.
+  liquid.filters.eventOrPressReleasesDateSorter = (dates, dateKey, reverse) => {
     if (!dates) return null;
     return dates.sort((a, b) => {
-      const start1 = a.fieldDatetimeRangeTimezone.value;
-      const start2 = b.fieldDatetimeRangeTimezone.value;
-      return start1 - start2;
+      const start1 = moment(a[dateKey].value);
+      const start2 = moment(b[dateKey].value);
+
+      return reverse ? start2 - start1 : start1 - start2;
     });
   };
 
--- a/src/site/filters/liquid.unit.spec.js
+++ b/src/site/filters/liquid.unit.spec.js
@@ -5,6 +5,7 @@ import registerFilters from './liquid';
 import vetCenterData from '../layouts/tests/vet_center/fixtures/vet_center_escanaba_data';
 import featuredContentData from '../layouts/tests/vet_center/fixtures/featuredContentData.json';
 import eventListingMockData from '../layouts/tests/vamc/fixtures/eventListingMockData.json';
+import pressReleasesMockData from '../layouts/tests/vamc/fixtures/pressReleasesMockData.json';
 
 const _ = require('lodash');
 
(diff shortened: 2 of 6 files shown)


src/site/filters/liquid.js
```diff
@@ -771,13 +771,15 @@ module.exports = function registerFilters() {
     });
   };
 
-  //* Sorts event dates (fieldDatetimeRangeTimezone) from newest to oldest.
-  liquid.filters.eventDateSorter = dates => {
+  //* Sorts event dates (fieldDatetimeRangeTimezone) starting with the most upcoming event.
+  //* Also sorts press releases (fieldReleaseDate) from newest to oldest.
+  liquid.filters.eventOrPressReleasesDateSorter = (dates, dateKey, reverse) => {
```
mpelzsherman, line 776: This function could be used to sort anything with dates. How about calling it just `sortByDateKey`?

Also, `dateKey` is now a required field. You might want to use `fieldDatetimeRangeTimezone` as the default.

```suggestion
  liquid.filters.sortByDateKey = (dates, dateKey = 'fieldDatetimeRangeTimezone', reverse = false) => {
```

src/site/filters/liquid.unit.spec.js
```diff
@@ -53,6 +53,49 @@ const eventsMockData = [
   },
 ];
 
+const pressReleasesMockData = [
```
mpelzsherman, line 56: This is kinda big. Move to a fixture file?

src/site/filters/liquid.unit.spec.js
```diff
@@ -135,18 +178,22 @@ describe('filterUpcomingEvents', () => {
...
+    expect(liquid.filters.eventOrPressReleasesDateSorter(null)).to.eq(null);
   });
 
   it('returns null when empty string is passed', () => {
-    expect(liquid.filters.eventDateSorter('')).to.eq(null);
+    expect(liquid.filters.eventOrPressReleasesDateSorter('')).to.eq(null);
   });
 
-  it('returns an array of upcoming events in date and time order starting with most recent to least recent', () => {
+  it('returns an array of upcoming events in date order starting with most recent to least recent', () => {
```
mpelzsherman, line 190: Probably not a bad idea to leave "and time" in there, or just say "date/time order".

src/site/filters/liquid.unit.spec.js
```diff
@@ -166,6 +213,57 @@ describe('eventDateSorter', () => {
...
   });
+
+  it('returns an array of press releases in date order from newest to oldest', () => {
+    expect(
+      liquid.filters.eventOrPressReleasesDateSorter(
+        pressReleasesMockData,
+        'fieldReleaseDate',
+        false,
+      ),
+    ).to.have.deep.members([
```
mpelzsherman, line 224: Can we reference this array as `pressReleasesMockData` rather than duplicating it here?

src/site/filters/liquid.unit.spec.js
```diff
@@ -166,6 +213,57 @@ describe('eventDateSorter', () => {
...
+        },
+        fieldIntroText:
+          'Veterans Affairs (VA) Pittsburgh Healthcare System is connecting women Veterans with VA and non-VA services during a resource fair on Friday, Aug. 16, from 10 a.m. to 1 p.m. ',
+        fieldReleaseDate: {
+          value: '2019-08-12T19:12:40',
+        },
+        title: 'Women Veterans resource fair spotlights VA and community care',
+      },
+    ]);
+  });
```
mpelzsherman, line 266: If we make `fieldDatetimeRangeTimezone` the default, can we add a test where we don't provide a `dateKey` param?

src/site/filters/liquid.js
mpelzsherman: We should probably make this function accept an object rather than a list of params, so you can pass any combination in any order.

This is generally a good idea once a function starts getting above 2 params, esp. if they aren't all required.

mpelzsherman (changes requested):
see comments

src/site/filters/liquid.js
gescobardev: This is true, I can rename.

src/site/filters/liquid.unit.spec.js
gescobardev: Yeah, I thought about that too. Will do

src/site/filters/liquid.js
gescobardev: Hmm I'm not so sure about passing `fieldDatetimeRangeTimezone` as a default value. In my opinion, if we set a default value, it should generally always work. In this case, `fieldDatetimeRangeTimezone` will work for events but will not work for sorting press releases. What are your thoughts about it?

src/site/filters/liquid.unit.spec.js
gescobardev: True, I will re-add

src/site/filters/liquid.js
mpelzsherman: Just because it doesn't always work, doesn't mean we shouldn't have a default. `fieldDatetimeRangeTimezone` appears in many places, including `src/site/includes/date.drupal.liquid`, so it seems like a good default value to me.

gescobardev: Having a default value implies that that argument is optional to pass in - meaning that if I don't pass anything in, it will run as intended every time. I guess the way I am seeing it is, we are only using `sortByDateKey` in 2 places where the key passed in can be either `fieldDatetimeRangeTimezone ` for events or `fieldReleaseDate` for press releases.

I can add the default though 😁, maybe I'm just missing something

src/site/filters/liquid.unit.spec.js
gescobardev: This array isn't an exact clone, it represents the correct date order of press releases. 
In testing it out I realized the this doesn't quite check for that order the way I originally thought so i'll fix that. 
But thats initially why I had it like that b/c it is not quite a duplicate.

mpelzsherman: OK, thanks for the explanation. It makes the test a bit verbose but I guess that's OK.

src/site/filters/liquid.js
mpelzsherman: > Having a default value implies that that argument is optional to pass in - meaning that if I don't pass anything in, it will run as intended every time.

Not necessarily. Setting a parameter default just gives the parameter a default value, which I think will be the same most of the time. It doesn't guarantee that the function will "run as intended every time".

I anticipate we might use this function in more places over time as we continue to migrate code out of the build.

But this is a minor point. We can always add the default later.

mpelzsherman (approved):

src/site/filters/liquid.unit.spec.js
gescobardev: Yes, I can add a test for this use case.

src/site/filters/liquid.unit.spec.js
```diff
@@ -53,6 +54,49 @@ const eventsMockData = [
...
+//     fieldIntroText:
+//       'VA Pittsburgh Healthcare System was named a 2020 “LGBTQ Healthcare Equality Leader” by the Human Rights Campaign Foundation (HRC). The designation is the eighth time in as many years and is listed in the 13th edition of the Healthcare Equality Index (HEI). ',
+//     fieldReleaseDate: {
+//       value: '2020-09-14T11:40:46',
+//     },
+//     title:
+//       'VA Pittsburgh named ‘LGBTQ Healthcare Equality Leader’ for 8th year',
+//   },
+// ];
+
```
mpelzsherman, line 99: Let's remove the commented-out code...

gescobardev: Oops, will remove

src/site/filters/liquid.js
gescobardev: thank you for your explanation and thoughts on it, I have added the default value.

PR #268, opened 2021-06-17
Changes made to operating status page

gescobardev:
## Description
Resolves the issue linked below

It was discovered that the missing content/data was, in fact, not there in the first place (operating statuses needed to be manually input into each facility- this became a separate issue that CMS will handle on their end)

I made changes to handle instances where the the `Facility operating statuses` jumplink and `Situation updates and information` h2 tags will not display if there is no content to show.

Use Wilkes-Barre to preview- http://localhost:3002/preview?nodeId=2369

## Acceptance criteria
- [ ] Both `Facility operating statuses` jumplink and `Situation updates and information` h2 tags will not display if there is no content to show.

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -879,4 +879,10 @@ module.exports = function registerFilters() {
   liquid.filters.isFirstPage = paginator => {
     return !paginator || paginator.prev === null;
   };
+
+  liquid.filters.isFieldSituationUpdates = fieldBannerAlert => {
+    return _.find(fieldBannerAlert, e => {
+      return e.entity.fieldSituationUpdates.length > 0;
+    });
+  };
 };
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -880,8 +880,8 @@ module.exports = function registerFilters() {
     return !paginator || paginator.prev === null;
   };
 
-  liquid.filters.isFieldSituationUpdates = fieldBannerAlert => {
-    return _.find(fieldBannerAlert, e => {
+  liquid.filters.hasFieldSituationUpdates = fieldBannerAlert => {
+    return _.some(fieldBannerAlert, e => {
       return e.entity.fieldSituationUpdates.length > 0;
     });
   };
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -882,8 +882,8 @@ module.exports = function registerFilters() {
 
   liquid.filters.hasContentAtPath = (rootArray, path) => {
     for (let i = 0; i < rootArray.length; i++) {
-      const hasLength = _.get(rootArray[i], path)?.length > 0;
-      if (hasLength) return hasLength;
+      const hasContent = _.get(rootArray[i], path)?.length > 0;
+      if (hasContent) return hasContent;
     }
     return false;
   };
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -880,9 +880,11 @@ module.exports = function registerFilters() {
     return !paginator || paginator.prev === null;
   };
 
-  liquid.filters.hasFieldSituationUpdates = fieldBannerAlert => {
-    return _.some(fieldBannerAlert, e => {
-      return e.entity.fieldSituationUpdates.length > 0;
-    });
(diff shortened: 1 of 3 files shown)


src/site/layouts/vamc_operating_status_and_alerts.drupal.liquid
```diff
@@ -7,151 +7,135 @@
...
+                  </p>
+                {% endfor %}
+              {% endif %}
+            </section>
+          {% endif %}
+
+        </article>
+      </div>
+    </div>
+  </main>
```
mpelzsherman, line 138: It's hard to distinguish the logic changes from the formatting changes here.

In the future, please push up a formatting PR first, then make your logic changes in a separate commit. That will make it much easier to review. (I'll do the same.)

src/site/filters/liquid.js
```diff
@@ -879,4 +879,10 @@ module.exports = function registerFilters() {
   liquid.filters.isFirstPage = paginator => {
     return !paginator || paginator.prev === null;
   };
+
+  liquid.filters.isFieldSituationUpdates = fieldBannerAlert => {
+    return _.find(fieldBannerAlert, e => {
+      return e.entity.fieldSituationUpdates.length > 0;
+    });
+  };
```
mpelzsherman, line 887: I'm not quite clear what this function does, but it looks like the intent is to return `true` if the `fieldBannerAlert` object contains any `fieldSituationUpdates`. Would `hasFieldSituationUpdates` be a better name?

Adding a test would also help clarify make the function's intended behavior.

BTW, if I am correct, you could probably just use the native Javascript function [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) instead of Lodash `_find`.

src/site/layouts/vamc_operating_status_and_alerts.drupal.liquid
gescobardev: you got it, I will do that next time

src/site/filters/liquid.js
mpelzsherman: Also: please check our existing filters to see if you can use one of the already-existing functions to meet this need, rather than creating a specific one for this purpose.

Any time a filter function includes a specific field in its name, that's kind of a "code smell" that we should use a generic function instead that takes the field name as an argument, so we don't end up with dozens of redundant functions.

gescobardev: Yes, `some()` is actually what I was looking for to use in this situation, thanks for that suggestion! I've never used it before but it works much better than `_.find()`. Lodash has `_.some()`, so I have used that. I added tests and have changed the function name as well.

gescobardev: Sounds good to me! I usually check the filters to see if we already have something we can use but I couldn't find anything with the keywords I was searching. 
This makes a lot of sense. I can also change my tests to fit this new filter was well.

src/site/filters/liquid.js
```diff
@@ -879,4 +879,12 @@ module.exports = function registerFilters() {
   liquid.filters.isFirstPage = paginator => {
     return !paginator || paginator.prev === null;
   };
+
+  liquid.filters.hasContentAtPath = (rootArray, path) => {
+    for (let i = 0; i < rootArray.length; i++) {
+      const hasContent = _.get(rootArray[i], path)?.length > 0;
+      if (hasContent) return hasContent;
```
mpelzsherman, line 886: Looking good! This can probably be made a bit shorter using some().

gescobardev: Very true! some() just slipped my mind.
changes made.

brangi (approved):

src/site/filters/liquid.js
```diff
@@ -879,4 +879,9 @@ module.exports = function registerFilters() {
   liquid.filters.isFirstPage = paginator => {
     return !paginator || paginator.prev === null;
   };
+
+  liquid.filters.hasContentAtPath = (rootArray, path) => {
+    const hasContent = e => _.get(e, path)?.length > 0;
+    return rootArray.some(hasContent);
+  };
```
mpelzsherman, line 886: Beautiful!

mpelzsherman (approved):

PR #269, opened 2021-06-17
Make images optional on VAMC pages (Main, Location, News Stories)

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/25599

### DIFF
--- a/script/preview.js
+++ b/script/preview.js
@@ -293,6 +293,9 @@ app.get('/preview', async (req, res, next) => {
       isScriptContext: true,
     });
 
+    const drupalAddressUrl = DRUPALS.PUBLIC_URLS[options['drupal-address']];
+    const drupalSite = drupalAddressUrl || 'prod.cms.va.gov';
+
     const files = {
       'generated/file-manifest.json': {
         path: 'generated/file-manifest.json',
@@ -302,9 +305,7 @@ app.get('/preview', async (req, res, next) => {
         ...fullPage,
         isPreview: true,
         headerFooterData: headerFooterDataSerialized,
-        drupalSite:
-          DRUPALS.PUBLIC_URLS[options['drupal-address']] ||
-          options['drupal-address'],
+        drupalSite,
       },
     };
 
--- a/src/site/components/image.drupal.liquid
+++ b/src/site/components/image.drupal.liquid
@@ -1,3 +1,3 @@
 {% if url and alt %}
-  <img class="{{ class }}" src="{{ url }}" alt="{{ alt }}" />
+  <img class="{{ class }}" src="{{ url }}" alt="{{ alt }}" title="{{ title }}" width="{{ width }}" height="{{  height }}"/>
 {% endif %}
--- a/src/site/components/image.drupal.liquid
+++ b/src/site/components/image.drupal.liquid
@@ -1,3 +1,19 @@
-{% if url and alt %}
-  <img class="{{ class }}" src="{{ url }}" alt="{{ alt }}" title="{{ title }}" width="{{ width }}" height="{{  height }}"/>
+{% assign height = image.derivative.height %}
+{% if height == empty%}
+  {% assign height = "100%" %}
+{% endif %}
+
+{% assign width = image.derivative.width %}
+{% if width == empty%}
+  {% assign width = "100%" %}
+{% endif %}
+
+{% if image.derivative.url and image.alt %}
+  <img class="{{ class }}"
+       src="{{ image.derivative.url }}"
+       alt="{{ image.alt }}"
+       title="{{ image.title }}"
+       width="{{ width }}"
+       height="{{ height }}"
+  />
 {% endif %}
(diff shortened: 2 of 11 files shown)


script/preview.js
```diff
@@ -302,9 +305,7 @@ app.get('/preview', async (req, res, next) => {
         ...fullPage,
         isPreview: true,
         headerFooterData: headerFooterDataSerialized,
-        drupalSite:
-          DRUPALS.PUBLIC_URLS[options['drupal-address']] ||
-          options['drupal-address'],
+        drupalSite,
```
mpelzsherman, line 308: These changes fix the "Edit this page in the CMS" links so they work in our local environments as well as in production.

brangi (approved):

src/site/components/image.drupal.liquid
```diff
@@ -0,0 +1,19 @@
+{% assign height = image.derivative.height %}
+{% if height == empty%}
+  {% assign height = "100%" %}
```
timwright12, line 3: Height and width values are pixel-based and don't take a unit, so `100%` wouldn't be a valid value: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-height

> The intrinsic height of the image, in pixels. Must be an integer without a unit.

src/site/components/image.drupal.liquid
```diff
@@ -0,0 +1,19 @@
...
+{% endif %}
+
+{% assign width = image.derivative.width %}
+{% if width == empty%}
+  {% assign width = "100%" %}
```
timwright12, line 8: Same deal with the width attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width

> The intrinsic width of the image in pixels. Must be an integer without a unit.

src/site/components/image.drupal.liquid
```diff
@@ -0,0 +1,19 @@
...
+{% endif %}
+
+{% if image.derivative.url and image.alt %}
+  <img class="{{ class }}"
+       src="{{ image.derivative.url }}"
+       alt="{{ image.alt }}"
+       title="{{ image.title }}"
```
timwright12, line 15: @1Copenut what are your thoughts about removing this title attribute? I think it atleast needs a conditional so we don't end up with `title=""`

timwright12 (changes requested):
Needs a check with accessibility and one HTML update

src/site/includes/tests/facilityListing.drupal.unit.spec.js
```diff
@@ -39,8 +41,26 @@ describe('Facility Listing', () => {
...
+  describe('Bad images', () => {
+    it('does not render image section if alt tag is missing', async () => {
+      const missingAltImageData = cloneDeep(data.main);
+      missingAltImageData.entity.fieldMedia.entity.image.alt = '';
+
+      container = await renderHTML(layoutPath, missingAltImageData);
+      expect(container.querySelector('section-image')).not.to.exist;
+    });
+
+    it('does not render image section if alt tag is missing', async () => {
```
gescobardev, line 58: Spelling mistake- I believe you meant to say "does not render image section if url is missing"

src/site/components/image.drupal.liquid
1Copenut: We should stick to the `alt` attribute for verbally describing images. Adding a `title` can create unexpected experiences with screen readers.

1Copenut (changes requested):
Please see my comment https://github.com/department-of-veterans-affairs/content-build/pull/269/files/1f14481d2f01dbbd10ff79a2f605a9e807355fa0#r656377111 for requested change.

src/site/components/image.drupal.liquid
mpelzsherman: @1Copenut - one argument in favor us using `title` in addition to `alt` is that it creates a tooltip for sighted users.

See this [CodePen](https://codepen.io/mpelzsherman/pen/ZEeNPXy) for an example.

But if you feel the `title` tag screen reader issue outweighs the benefits of the tooltip, that's OK by me!

1Copenut: Thanks @mpelzsherman for the Codepen link. I do feel the `alt` attribute is the right approach for adding a spoken description of the image, and removing the `title` attribute. The tooltips can't be zoomed in, and haven't been tested for consistency across screen readers and browsers.

If there's text that's important, I'd opt to put it on the page so it can be zoomed and consumed by all users, with a semantic relationship to the image.

mpelzsherman: Sounds good. I mean these image `title` tooltips have been around since the days of `<blink>`, so I don't think they will be particularly missed. 😄

src/site/components/image.drupal.liquid
mpelzsherman: Thanks for pointing that out. Removing!

timwright12 (approved):

1Copenut (approved):
LGTM! I talked with @mpelzsherman about `alt=""` being permissible for decorative images. He's going to add that to the Liquid template.

gescobardev (approved):

PR #283, opened 2021-06-23
Create headerFooter in preview server

theodur:
## Description
Currently, when using the preview server locally, we need to run a full content build prior so that `headerFooter.json` is created and a symlink to `file-manifest.json` is made. This PR updates the preview server script, so we don't need to run a content build just to create/reference those two files.

The `create-header-footer` plugin is added to the preview server's Metalsmith pipeline, so now we don't need to fetch that file from the `generated` directory, which would cause an error on the preview server if it didn't exist. Instead `headerFooter.json` is built on demand when viewing a page in the preview server.

A symlink to `generated/file-manifest.json` will be created in the preview script if one doesn't already exist. This avoids the need to run a full content build, since the content build makes a symlink to vets-website's `generated` directory to acomplish this as well.

## Testing done
Tested locally on `/preview` and `/diff`.

## Acceptance criteria
- [ ] A full content build doesn't need to be executed to use the preview server locally.
- [ ] The preview server functions as it did before.

## Definition of done

### DIFF
--- a/script/preview.js
+++ b/script/preview.js
@@ -196,7 +196,9 @@ function fetchAllPageData(nodeId) {
           return resp.json();
         }
         throw new Error(
-          `HTTP error when fetching manifest: ${resp.status} ${resp.statusText}`,
+          options.buildtype !== ENVIRONMENTS.LOCALHOST
+            ? `HTTP error when fetching manifest: ${resp.status} ${resp.statusText}`
+            : 'file-manifest.json is missing. Try running "yarn build" in vets-website.',
         );
       },
     ),
--- a/script/preview.js
+++ b/script/preview.js
@@ -74,7 +74,7 @@ if (options.buildpath === null) {
   options.buildpath = `build/${options.buildtype}`;
 }
 
-// Create symlink to 'vets-website/generated' if it doesn't exist
+// Create symlink to 'vets-website/generated' if one doesn't exist
 // so we don't have to run the content build
 if (
   options.buildtype === ENVIRONMENTS.LOCALHOST &&
--- a/script/preview.js
+++ b/script/preview.js
@@ -23,6 +23,7 @@ const HOSTNAMES = require('../src/site/constants/hostnames');
 const DRUPALS = require('../src/site/constants/drupals');
 const bucketsContent = require('../src/site/constants/buckets-content');
 const singlePageDiff = require('./preview-routes/single-page-diff');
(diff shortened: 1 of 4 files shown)


theodur:
One small caveat I noticed is if you don't have a prior content build created, the image in the header and footer won't be shown, since the `img` directory isn't present in the build directory. This will only happen on localhost, since the `href`'s for images in non-local environments reference the s3 domains.

script/preview.js
```diff
@@ -196,7 +196,9 @@ function fetchAllPageData(nodeId) {
           return resp.json();
         }
         throw new Error(
-          `HTTP error when fetching manifest: ${resp.status} ${resp.statusText}`,
+          options.buildtype !== ENVIRONMENTS.LOCALHOST
+            ? `HTTP error when fetching manifest: ${resp.status} ${resp.statusText}`
+            : 'file-manifest.json is missing. Try running "yarn build" in vets-website.',
```
mpelzsherman, line 201: 🚀

mpelzsherman:
The original goal of the [ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/25793) was to avoid having to run a **vets-website** build to do a `yarn preview` in content-build. We haven't really solved that problem here, but that's OK. Generating header and footer data in `content-build` is better. Thanks!

theodur:
@mpelzsherman Maybe there is some misunderstanding. The ticket didn't mention avoiding the need to run the vets-website build specifically. It only mentioned not needing to run "`yarn build`" to generate the headerFooter.json file. That file is created in the content build, **not** the vets-website build. The vets-website build doesn't have anything to do with creating that file. This PR makes it so you don't need to run the content build just to create the headerFooter.json file and make a symlink to the file manifest in vets-website, which takes 4-5 times longer than the vets-website build.

mpelzsherman:
> @mpelzsherman Maybe there is some misunderstanding. The ticket didn't mention avoiding the need to run the vets-website build specifically. It only mentioned not needing to run "`yarn build`" to generate the headerFooter.json file. That file is created in the content build, **not** the vets-website build. The vets-website build doesn't have anything to do with creating that file. This PR makes it so you don't need to run the content build just to create the headerFooter.json file and make a symlink to the file manifest in vets-website, which takes 4-5 times longer than the vets-website build.

Yes, sorry for my confusion. It looks like this PR will be a HUGE time saver! Thanks so much!

mpelzsherman (approved):

k80bowman (approved):

PR #299, opened 2021-06-28
Update alt with aria-label

kelsonic:
# Description

This PR resolves an a11y issue with the `Download (PDF)` links on CLPs.


mpelzsherman (approved):
LGTM

PR #304, opened 2021-06-30
Update Vet Centers centralized content logic to handle new format

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/26674

The format of some Vet Centers centralized content fields changed unexpectedly. This PR updates the logic to handle it.


src/site/filters/liquid.js
```diff
@@ -745,33 +745,33 @@ module.exports = function registerFilters() {
     }
     /* eslint-disable camelcase */
```
brangi, line 746: we can remove this one now ?

src/site/layouts/vet_center.drupal.liquid
```diff
@@ -5,172 +5,172 @@
...
+            <h1>{{ entityLabel }}</h1>
+          {% endif %}
+
+          {% if fieldIntroText != empty %}
+            <div class="va-introtext">
+              <p>{{ fieldIntroText }}</p>
+            </div>
+          {% endif %}
+
+          <nav id="table-of-contents" aria-labelledby="on-this-page">
```
timwright12, line 22: Missing the reference ID for this aria-labelledby "on-this-page". Probably just add it to to the child H2

src/site/layouts/vet_center.drupal.liquid
```diff
@@ -5,172 +5,172 @@
...
-                        %}
+            <div class="usa-width-two-thirds vads-u-display--block vads-u-width--full">
+
+              <div>
+                <div class="vads-c-facility-detail">
+                  <section class="vads-facility-detail">
+                    <h3 class="vads-u-font-size--lg vads-u-margin-top--0 vads-u-line-height--1 vads-u-margin-bottom--1">
+                      Main Location</h3>
+                    <div data-widget-type="expandable-operating-status-{{ fieldFacilityLocatorApiId }}"
+                         facilityId="{{ fieldFacilityLocatorApiId }}"
```
timwright12, line 40: These need to be `data-*` attributes to be valid HTML: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

timwright12 (changes requested):

brangi (approved):

src/site/filters/liquid.js
mpelzsherman: Yeah! Good catch!

src/site/layouts/vet_center.drupal.liquid
mpelzsherman: This is a very common issue throughout the entire site. Thanks for pointing it out. cc @kelsonic @gescobardev

src/site/layouts/vet_center.drupal.liquid
mpelzsherman: Changing these attributes would require changes to the React widgets in vets-website.
I get that it's not valid HTML, but I'm not aware of any actual bugs around this.

BTW - The changes you are seeing here are mostly a result of auto-formatting the file, and are not related to the original ticket.

timwright12 (approved):

PR #316, opened 2021-07-02
Bump timeout for a11y tests that are timing out

timwright12:
## Description
Extending the timeout for daily accessibility tests because they are sometimes failing with the previous 20 second timeout limit.


U-DON (approved):
Hopefully we don't need to go right to `ludicrouslySlow`.

PR #317, opened 2021-07-02
Format phone number links in Vet Center health service accordions

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/26688


brangi (approved):
LGTM

PR #332, opened 2021-07-07
Updated file path search for drupal assets

timwright12:
## Description
File path regex was failing to download on assets from other sites that were also using Drupal. Reverting to a previous functional state.


acrollet (approved):
Change seems to make sense and a build succeeded on tugboat:

Screen Shot 2021-07-07 at 2 55 08 PM

kelsonic (approved):
lgtm!

PR #333, opened 2021-07-07
Move CMS Link checking to Broken Link Step

indytechcook:
## Description

Move CMS Link checking to Broken Link plugin. 

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/5755

## Testing done

* Ran the content build against prod Drupal.
* Ran the build against local Drupal with a cms URL added and verified it was in the broken link JSON file.

## Definition of done


alexandec:
Thanks for doing this! Should be able to remove `plugins/check-cms-urls.js` as well

src/site/stages/build/plugins/modify-dom/check-broken-links/helpers/getBrokenLinks.js
```diff
@@ -37,8 +38,9 @@ function getBrokenLinks(file, allPaths, isBrokenLink = _isBrokenLink) {
     }
 
     const isBroken = isBrokenLink(target, currentPath, allPaths);
+    const isCMS = isCMSUrl(target, file);
```
swirtSJW, line 41: I think this has to run first.  Otherwise we will get failures on both.  If there

swirtSJW (commented):
This looks ok.  But I am not sure if the intent was kept, mainly because I can't go see the other logic (on my phone).  Originally the cmsurl check would fail the build.  Now it seems like we are just alerting? Or does the same count threshold of 10 still apply?  Fewer than 10 cms urls should be allowed?

I am thinking that if a cms url is found, we should just rewrite it before the link checker runs. Less noise and more fault tolerant.

src/site/stages/build/plugins/modify-dom/check-broken-links/helpers/getBrokenLinks.js
swirtSJW: This was incorrect. Disregard.

swirtSJW: This was incorrect. Disregard.

indytechcook: It doesn't duplicate entires.  It just treats it as another broken link.

swirtSJW: This was incorrect. Disregard.

indytechcook:
> This looks ok. But I am not sure if the intent was kept, mainly because I can't go see the other logic (on my phone). Originally the cmsurl check would fail the build. 

This is the intended purpose.  To not block builds on a single cms URLs.  They should just be treated as another broken link. 

> I am thinking that if a cms url is found, we should just rewrite it before the link checker runs. Less noise and more fault tolerant.

This would be a good follow up.

src/site/stages/build/plugins/modify-dom/check-broken-links/helpers/getBrokenLinks.js
swirtSJW: Sorry.  Stuck phone

alexandec:
This is failing on Jenkins but it should pass if you merge in the latest master

swirtSJW (approved):

alexandec:
On Jenkins there are a lot of broken links reported:
Screen Shot 2021-07-08 at 8 30 53 AM

indytechcook:
Looks like image links are being evaluated.  I didn't see this locally.  Looking

indytechcook:
This does work in CI and local because the Drupal base URL is not replaced in the `convertDrupalFilesToLocal` method due to the `usingAWS` variable.  I could either update the current check Drupal URL step to add items to the broken link json files OR change the `usingAWS` variable to make the assets use relative paths instead of Drupal full paths.

Thoughts?

alexandec:
I'm not sure of all the implications in this case, but it seems like using relative paths is generally a good approach where possible

indytechcook:
image

indytechcook:
@alexandec can you give this another check please

k80bowman (approved):

src/site/stages/build/plugins/modify-dom/check-broken-links/helpers/isCMSUrl.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param {*} link The HREF/SRC value to be validated.
+ * @param {*} file The HTML file from the Metalsmith pipeline.
+ * @param {*} buildOptions The build options.
+ * @returns {boolean} Does this URL point to the CMS.
+ */
+function isCMSUrl(link, file, buildOptions) {
+  if (!link) return false;
+
+  const parsed = url.parse(link);
```
alexandec, line 16: Is there the potential for this to throw and break the build?

alexandec:
Looks good! Just added a question

alexandec (approved):

src/site/stages/build/plugins/modify-dom/check-broken-links/helpers/isCMSUrl.js
indytechcook: I can add a try/catch

indytechcook: This is the pattern as https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/stages/build/plugins/modify-dom/check-broken-links/helpers/isBrokenLink.js#L13 which is run before this code.  The ` if (!link) return false;` was needed as without it exceptions are thrown.

kelsonic (approved):

PR #331, opened 2021-07-07
Sorted VAMC health services by name

gescobardev:
## Description
Resolves Issue link below

To preview you can use - http://localhost:3002/preview?nodeId=7712

## Acceptance criteria
- [ ] VAMC health services should be sorted by name - alphabetically

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #338, opened 2021-07-08
Fixed URL validation for featured content

gescobardev:
## Description
Resolves issue linked below

To preview use Abilene - http://localhost:3002/preview?nodeId=3886

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -905,15 +905,14 @@ module.exports = function registerFilters() {
     return rootArray.some(hasContent);
   };
 
-  liquid.filters.isValidUrl = str => {
+  liquid.filters.hasCharacterOtherThanSpace = str => {
     if (!str) return null;
-    let url;
+    const substring = 'internal:/';
+    let newStr;
 
-    try {
-      url = new URL(str);
-    } catch (e) {
-      return false;
-    }
-    return url.protocol === 'http:' || url.protocol === 'https:';
+    if (str.includes(substring)) {
+      newStr = str.replace(substring, '');
+      return /\S/.test(newStr);
+    } else return /\S/.test(str);
   };
 };
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -906,7 +906,7 @@ module.exports = function registerFilters() {
   };
 
   liquid.filters.hasCharacterOtherThanSpace = str => {
-    if (!str) return null;
+    if (!str) return false;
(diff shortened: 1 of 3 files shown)


src/site/filters/liquid.js
```diff
@@ -905,15 +905,14 @@ module.exports = function registerFilters() {
     return rootArray.some(hasContent);
   };
 
-  liquid.filters.isValidUrl = str => {
+  liquid.filters.hasCharacterOtherThanSpace = str => {
     if (!str) return null;
-    let url;
+    const substring = 'internal:/';
```
gescobardev, line 910: https://github.com/department-of-veterans-affairs/content-build/pull/331

gescobardev: Screen Shot 2021-07-08 at 1 05 40 PM

src/site/filters/liquid.unit.spec.js
```diff
@@ -54,33 +54,41 @@ const eventsMockData = [
...
 
   it('returns null if null is passed', () => {
-    expect(liquid.filters.isValidUrl(null)).to.be.null;
+    expect(liquid.filters.hasCharacterOtherThanSpace(null)).to.be.null;
   });
 
   it('returns null if undefined', () => {
-    expect(liquid.filters.isValidUrl(undefined)).to.be.null;
+    expect(liquid.filters.hasCharacterOtherThanSpace(undefined)).to.be.null;
   });
```
mpelzsherman, line 68: It looks like `hasCharacterOtherThanSpace` sometimes returns `null`, and other times returns a Boolean.

It would be better to always return a Boolean, if possible.

src/site/filters/liquid.js
```diff
@@ -905,15 +905,14 @@ module.exports = function registerFilters() {
     return rootArray.some(hasContent);
   };
 
-  liquid.filters.isValidUrl = str => {
+  liquid.filters.hasCharacterOtherThanSpace = str => {
     if (!str) return null;
```
mpelzsherman, line 909: Can we return `false` here instead?

src/site/filters/liquid.unit.spec.js
gescobardev: Yes, that makes sense. I will changes

src/site/filters/liquid.js
gescobardev: Yes

mpelzsherman (approved):

PR #383, opened 2021-07-20
DO NOT MERGE [self-hosted, asg] to [self-hosted, Snoopy]

ndsprinkle:
# DO NOT MERGE

## Description
Related to department-of-veterans-affairs/va.gov-team#27680.

[self-hosted, asg] to [self-hosted, Snoopy]

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -135,7 +135,7 @@ jobs:
   
   build:
     name: Build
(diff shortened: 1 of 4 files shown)


oseasmoran73:
With #412 , Snoopy will no longer be available and persistent/on-demand runners will be dynamically spun up. We can probably close PR.

PR #394, opened 2021-07-23
Wrap footer in role=contentinfo section for consistent landmarks

Rleahy22:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/7425

Footer announcements under `#announcement-root` exist outside of a landmark, so this moves both the `footer` and `#announcement-root` under a `section` with `role="contentinfo"` This is inside of a `section` because a `footer` automatically gets `role="contentinfo"` unless it is contained within a small subset of tags with `section` being the most appropriate of this subset. see https://www.w3.org/TR/wai-aria-practices/examples/landmarks/HTML5.html

## Testing done
local testing with http://matatk.agrip.org.uk/landmarks/

## Screenshots
Screen Shot 2021-07-23 at 3 00 31 PM

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/site/includes/footer.html
+++ b/src/site/includes/footer.html
@@ -1,7 +1,9 @@
-<div id="announcement-root"></div>
-<footer class="footer" role="contentinfo">
-  <div id="footerNav"></div> <!-- let's try this React thing -->
-</footer>
+<section role="contentinfo">
+  <div id="announcement-root"></div>
+  <footer class="footer">
+    <div id="footerNav"></div> <!-- let's try this React thing -->
+  </footer>
+</section>
 
 <!--
 “To care for him who shall have borne the battle and for his widow, and his orphan.”
--- a/src/site/includes/header.html
+++ b/src/site/includes/header.html
@@ -120,9 +120,14 @@
 
   {% include "src/site/components/fullwidth_banner_alerts.drupal.liquid" %}
 
-  <div data-widget-type="banners" data-homepage-banner-content="{{ homepage_banner.content | escape }}"
+  <div
+    data-widget-type="banners"
+    data-homepage-banner-content="{{ homepage_banner.content | escape }}"
     data-homepage-banner-title="{{ homepage_banner.title | escape }}"
-    data-homepage-banner-type="{{ homepage_banner.type }}" data-homepage-banner-visible="{{ homepage_banner.visible }}">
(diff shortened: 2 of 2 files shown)


zacharymorel (approved):

kelsonic (approved):
lgtm! 🎉

Rleahy22:
I added `role="region"` to the homepage banner since that was also without a landmark
Screen Shot 2021-07-26 at 9 46 53 AM
However if there are multiple regions on a page, which can happen if we have more than one banner, then they should have unique labels. They should really have unique labels either way. The banners are web components so we would need changes there, specifically the banner title's element (`h2`/h3`) should get a unique id which we could reference as the `aria-labelledBy` in the `div` that includes the banner.

kelsonic:
This is awesome, ty @Rleahy22! So the banners that can appear are 1) the [Emergency Banner](https://github.com/department-of-veterans-affairs/component-library/blob/master/src/components/EmergencyBanner/EmergencyBanner.jsx), 2) the [Maintenance Banner](https://github.com/department-of-veterans-affairs/component-library/blob/master/src/components/MaintenanceBanner/MaintenanceBanner.jsx), and then 3) we have modals/sticky announcements that can be found [here](https://github.com/department-of-veterans-affairs/vets-website/blob/5b602f840f45f369fff651c0193b5b68bbd58390/src/platform/site-wide/announcements/config/index.js). Not sure if that helps! 🙂

Rleahy22:
@cvalarida @bkjohnson Would it be possible to add a unique id for each banner via the component library. I plan on merging this pr as is and seeing if we can improve even further upon the banner landmarks in the future. see below comment:

> I added `role="region"` to the homepage banner since that was also without a landmark
> Screen Shot 2021-07-26 at 9 46 53 AM
> However if there are multiple regions on a page, which can happen if we have more than one banner, then they should have unique labels. They should really have unique labels either way. The banners are web components so we would need changes there, specifically the banner title's element (`h2`/`h3`) should get a unique id which we could reference as the `aria-labelledBy` in the `div` that includes the banner.

bkjohnson:
I glanced at the source for those two React components and it doesn't look like it's possible at the moment to have an `id` go anywhere. However, they both use `<AlertBox>` under the hood, and it is possible for you to use [`<va-alert>`](https://design.va.gov/storybook/?path=/docs/components-va-alert--default) and give it an `id`.

PR #401, opened 2021-07-27
Wrap phone numbers with parentheses around the area code

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/27501

### DIFF
--- a/src/site/filters/liquid.unit.spec.js
+++ b/src/site/filters/liquid.unit.spec.js
@@ -1131,6 +1131,13 @@ describe('phoneLinks', () => {
     expect(liquid.filters.phoneLinks(text)).to.equal(expected);
   });
 
+  it('wraps phone numbers with parentheses around the area code', () => {
+    const text = 'Here is a phone number: (123)-456-7890. Pretty cool!';
+    const expected =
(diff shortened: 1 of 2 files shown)


Rleahy22 (approved):
LGTM. Only question is if we want to maintain the parentheses in the display. Good to merge though

kelsonic (approved):

kelsonic:
Just curious but do we also need to support the TTY: 711 number use case?

src/site/filters/liquid.unit.spec.js
```diff
@@ -1131,6 +1131,22 @@ describe('phoneLinks', () => {
     expect(liquid.filters.phoneLinks(text)).to.equal(expected);
   });
 
+  it('wraps phone numbers with parentheses around the area code', () => {
```
gescobardev, line 1134: Can we add another test where the phone number looks like this - `(123) 456-7890` ?

gescobardev (changes requested):
Looks great! I just have one comment about testing.

src/site/filters/liquid.unit.spec.js
mpelzsherman: Good catch, thanks! Adding.

gescobardev (approved):

PR #409, opened 2021-07-30
remove sharableLink usage from accordions

adamwhitlock1:
## Description
Removes two implementations of the SharableLink react widget component specifically when they are used in the expandable content of the `va-accordion` web component.

The reason for needing to remove these is noted in this ticket comment: [#26744](https://github.com/department-of-veterans-affairs/va.gov-team/issues/26744#issuecomment-889268802)

Removing these results only one instance of SharableLink via [q_a.drupal.liquid](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/paragraphs/q_a.drupal.liquid#L31) and this instance works correctly because it is not nested within a web component.

Further work will be done at a later time to refactor this component to work appropriately since it is currently shown on only a single page in an experimental fashion. [coronavirus-veteran-frequently-asked-questions](https://www.va.gov/coronavirus-veteran-frequently-asked-questions/)

## Testing done
Tested on local content build. Verified removal.

## Screenshots

Screen Shot 2021-07-29 at 9 54 29 PM

## Acceptance criteria
- [ ] The SharableLink component is not seen within accordions on the coronavirus-veteran-frequently-asked-questions page.

## Definition of done

### DIFF
--- a/src/site/paragraphs/collapsible_panel.drupal.liquid
+++ b/src/site/paragraphs/collapsible_panel.drupal.liquid
@@ -50,11 +50,6 @@
                             {% include {{ bundleComponent }} with entity = paragraph.entity %}
                         {%  endfor %}
 
-                        <div
-                            data-widget-type="sharable-link"
-                            parentId="{{id}}"
-                            sharableText="{{item.fieldTitle}}"
-                        ></div>
                     </div>
                 </div>
             </va-accordion-item>
--- a/src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
+++ b/src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
@@ -22,11 +22,6 @@
               {% endfor %}
             </div>
 
-            <div
-              data-widget-type="sharable-link"
-              parentId="{{id}}"
-              sharableText="{{item.fieldQuestion}}"
-            ></div>
           </div>
         </div>
       </va-accordion-item>


alexandec:
@cvalarida @bkjohnson any thoughts on this change?

bkjohnson (commented):
I'm not super familiar with the `sharable-link` widget, but getting away from the static-pages React widget injection seems like a step in the right direction as far as I'm concerned.

adamwhitlock1:
@bkjohnson Thanks for the feedback, and yes I would agree that the react widget pattern isn't the best thing to be injecting all around static pages. This particular usage is broken in accordions due to the dom querying that it attempts to do, which wont work for inside the new va-accordion web component.

@kelsonic Do you think this PR would be something that could be easily approved? I know my PM @SKasimow was looking to have the broken instances removed asap.

Since it is just yanking out some usages on a single page it should have minimal impact.

bkjohnson:
However, I'm also not sure if `sharable-link` is working as intended even outside of web components:

image

adamwhitlock1:
> However, I'm also not sure if `sharable-link` is working as intended even outside of web components:
> 
> image
> 
> Screenshot taken from: https://www.va.gov/health-care/about-va-health-benefits/dental-care/dental-insurance/

My understanding of this component is that it is only actually rendered on one particular page [coronavirus-veteran-frequently-asked-questions](https://www.va.gov/coronavirus-veteran-frequently-asked-questions/) since it is an experimental component that is being tested / evaluated for further future use. (I'm fairly new at working on va.gov)

Logic for url based rendering is here: [https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/static-pages/sharable-links/createSharableLink.js#L9](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/static-pages/sharable-links/createSharableLink.js#L9)

bkjohnson:
Gotcha. Good to know!

alexandec (approved):

PR #417, opened 2021-08-03
Github Actions (CI)  -- retry on install dependencies for build

oseasmoran73:
## Description

In an effort to reduce flakiness, adding retry to installation for build.


alexandec (approved):

alexandec:
@oseasmoran73 Looks like this PR is causing the following error: `Error: Cannot find module 'semver'`

PR #421, opened 2021-08-03
Sort satellite vet centers

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/26691


src/site/layouts/vet_center_locations_list.drupal.liquid
```diff
@@ -21,29 +21,29 @@
...
+              If you can’t make it to our
+              {{ fieldOffice.entity.title }}
+              we offer satellite locations that may be closer to you.
+              These satellite facilities provide select services with the same community, care, and
+              confidentiality in a non-medical setting.
+              Call us for more information about these locations.
             </p>
-            {% for entityVetCenter in fieldOffice.entity.reverseFieldOfficeNode.entities | sortObjectsBy: 'title' %}
+            {% assign satteliteVetCenters = fieldOffice.entity.reverseFieldOfficeNode.entities | sortObjectsBy: 'title' %}
+            {% for entityVetCenter in satteliteVetCenters %}
```
gescobardev, line 39: Should change the variable to `satelliteVetCenters` for consistency.

gescobardev:
The way that I understood the ticket was that within the Satellite Locations section, there are 3 possible types of locations that can be listed - outstations, CAPs and mobile vet centers

The order in which they are listed should be as follows - outstations & CAPs (both sorted by alpha) should be shown before any mobile vet centers (sorted by alpha).
There should be 2 groupings within Satellite Locations - outstations/CAPs vs mobile centers.
If thats the case, then it looks like we can look at the `entityBundle` to discern between each of them.

mpelzsherman:
> The way that I understood the ticket was that within the Satellite Locations section, there are 3 possible types of locations that can be listed - outstations, CAPs and mobile vet centers
> 
> The order in which they are listed should be as follows - outstations & CAPs (both sorted by alpha) should be shown before any mobile vet centers (sorted by alpha).
> There should be 2 groupings within Satellite Locations - outstations/CAPs vs mobile centers.
> If thats the case, then it looks like we can look at the `entityBundle` to discern between each of them.

I'm pretty sure that the outstations & CAPs are included in the GraphQL field `fieldNearbyVetCenters`, and the mobile centers are in `fieldNearbyMobileVetCenters`. I will confirm with the CMS team.

mpelzsherman:
> There should be 2 groupings within Satellite Locations - outstations/CAPs vs mobile centers.
> If thats the case, then it looks like we can look at the `entityBundle` to discern between each of them.

Please disregard my earlier comment. 

You are correct that A/C on the ticket states:

"within the "Satellite Locations" list = outstations and community access points are grouped together and sorted by alpha order followed by Mobile Vet Centers, grouped in alpha order"

I would much rather separate out the mobile vet centers in GraphQL if possible.

It does look like the GraphQL query `fieldOffice.entity.reverseFieldOfficeNode` _might_ include some mobile vet centers. It is unclear whether these might be the same as those returned by `fieldNearbyMobileVetCenters`.

 I will check with the CMS team on this. Thanks for raising the issue!

gescobardev:
No problem! 
Yepp, separating them out in graphql would be ideal.
You're right it is confusing why that one mobile vet center isn't returned within `fieldNearbyMobileVetCenters`. 
Just thinking out loud, but could there maybe be a difference between "_nearby_ mobile vet centers" and "_satellite_ mobile vet centers"?

mpelzsherman:
From @kevwalsh in Slack:

> `reverseFieldOfficeNode` is showing the facilities that are “owned” by the parent Vet Center, such as Outstations, or MVCs that they “own”.  `fieldNearbyMobileVetCenters`   is for MVCs that they don’t “own”, but they want to highlight. 

So we need to merge, sort, and dedupe these 2 fields.

mpelzsherman:
Per @davidconlon we can skip deduping for now.

gescobardev (changes requested):

src/site/filters/liquid.js
```diff
@@ -711,6 +711,11 @@ module.exports = function registerFilters() {
     return data.filter(e => _.get(e, filterBy) === valueFilter);
   };
 
+  liquid.filters.rejectBy = (data, filterBy, valueFilter) => {
+    if (!data) return null;
+    return data.filter(e => _.get(e, filterBy) !== valueFilter);
+  };
```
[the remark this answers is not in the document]
mpelzsherman: Done! Please re-review.

src/site/filters/liquid.unit.spec.js
```diff
@@ -1052,36 +1052,55 @@ describe('strip', () => {
...
+describe('rejectBy', () => {
+  const testData = [
+    { class: { abstract: { number: 3 } } },
+    { class: { abstract: { number: 5 } } },
+    { class: { abstract: { number: 4 } } },
+    { class: { abstract: { number: 1 } } },
+    { class: { abstract: { number: 1 } } },
+  ];
+
+  it('returns all objects matching the given path and value', () => {
```
gescobardev, line 1092: I believe this should say - `returns all objects that do not match the given path and value`

gescobardev (approved):

PR #442, opened 2021-08-10
Use full size images on news story pages

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/27280

Tested in [Tugboat](https://web-ldij0hdlillzgrlzsgjk8keqbcknof2i.demo.cms.va.gov/eastern-colorado-health-care/stories/).

### DIFF
--- a/src/site/components/image.drupal.liquid
+++ b/src/site/components/image.drupal.liquid
@@ -8,11 +8,18 @@
       {% endif %}
       src="{{ image.derivative.url }}"
       alt="{{ image.alt }}"
-      {% if image.derivative.width %}
+      {% if width %}
+        width={{ width }}
+      {% elseif image.derivative.width %}
         width="{{ image.derivative.width }}"
       {% endif %}
-      {% if image.derivative.height %}
+      {% if height %}
+        height={{ height }}
+      {% elseif image.derivative.height %}
         height="{{ image.derivative.height }}"
+      {% endif %}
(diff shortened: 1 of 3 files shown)


Rleahy22 (approved):

gescobardev:
Resolution-wise it looks a lot better!
Just one quick question, is it OK that some photos now don't fit the container? If that makes sense.

<details>
<summary>For example, some story page images now look like this:</summary>

Screen Shot 2021-08-11 at 10 49 19 AM

Screen Shot 2021-08-11 at 10 50 16 AM

</details>


<details>
<summary>as opposed to this (grainy-ness aside) : </summary>

Screen Shot 2021-08-11 at 10 53 04 AM

</details>

nicko-amida:
@gescobardev To stay in line with the design system, this kind of image needs to [maintain a 3:2 aspect ratio](https://design.va.gov/design/image-aspect-ratios). A possible solution is a to set a minimum image size at the upload point.

mpelzsherman:
We can crop the images with CSS to fit into a predefined container, but arbitrary cropping could do more harm than good.

Therefore I think we should enforce this on the CMS side and just display the images on the FE.

ndsprinkle (approved):

mpelzsherman:
Using the `_21LARGE` style is the ideal solution. Thanks to @swirtSJW for providing this info.

PR #464, opened 2021-08-19
[28401] Prevent No Commit Pushes to Origin

ndsprinkle:
## Description
Currently, developers are able to push their branches to origin without commits. This has previously caused some issues that are detailed in the original ticket. This PR adds a pre-push hook to husky that checks if there have been any commits made before allowing the push.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#28401

## Acceptance criteria
- [ ] Developers can't push branches to origin without commits


.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+./script/pre-push.sh
```
theodur, line 4: Small nit: Can we add a newline at the end of this file?

script/pre-push.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+if [[ ! `git log origin/master.. --no-merges` ]]; then
+  echo "No commits to push! Add at least one commit to push this branch to origin."
+  exit 1
+fi
```
theodur, line 6: Same here for the newline

theodur (approved):
I was able to test this out, it works great!

PR #469, opened 2021-08-24
Migrate accessibility tests to Cypress and GitHub Actions

ddzz:
## Description
This PR migrates the daily accessibility scan to Cypress. Tests will run on a daily schedule with Slack notifications. The existing Nightwatch tests will be removed in a separate PR.

## Acceptance criteria
- [ ] Cypress a11y tests run and pass on GitHub Actions
- [ ] #-daily-accessibility-scan Slack channel is notified of results of test runs


timwright12:
converted to draft since it seems like you're still working on this

ddzz:
> converted to draft since it seems like you're still working on this

Thanks, it's almost ready 🤞

1Copenut (approved):
I'm 💯 for these changes. We've worked closely on making sure things are running the way we expect. I'll notify the VA 508 office the static content checks don't include the async map load and CC our PMs.

timwright12 (approved):

.github/workflows/a11y.yml
```diff
@@ -0,0 +1,249 @@
...
+
+      - name: Get Slack bot token
+        uses: marvinpinto/action-inject-ssm-secrets@v1.1.1
+        with:
+          ssm_parameter: /devops/github_actions_slack_bot_user_token
+          env_variable_name: SLACK_BOT_TOKEN
+
+      - name: Accessibility Test Failure
+        uses: ./.github/actions/vsp-github-actions/slack-socket
+        if: ${{ failure() }}
```
CBonade, line 231: in `vets-website` in order to reference success/failure in a prior job, we had to reference it as `${{needs.a11y.result == 'failure'}}`.  Are we sure this works in this syntax?

CBonade (approved):
Just one question about the failure/success states on the slack step. If all is confirmed, LGTM!

.github/workflows/a11y.yml
ddzz: Good catch, I'll fix this.

PR #498, opened 2021-08-31
Github Actions (daily-deploy & content-release) -- Enable daily deploy in github action for prod

oseasmoran73:
## Description

This PR addresses the following tickets:

[Copy and Enable daily production deploy to GitHub Actions workflow](https://github.com/department-of-veterans-affairs/va.gov-team/issues/24191)
[Copy and Enable scheduled content release to GitHub Actions workflow](https://github.com/department-of-veterans-affairs/va.gov-team/issues/24193)

This PR enables environmental variables for prod.

## Testing done

See [Slack](https://dsva.slack.com/archives/C024DUC9S1K) for more details


script/github-actions/check-broken-links.js
```diff
@@ -18,9 +18,9 @@ if (fs.existsSync(reportPath)) {
   const shouldFail =
     brokenLinks.isHomepageBroken ||
     brokenLinks.brokenLinksCount > maxBrokenLinks;
-  const color = shouldFail ? '#D33834' : '#FFCC00'; // danger or warning, needs to be in hex
+  const color = shouldFail ? 'danger' : 'warning';
```
oseasmoran73, line 21: Looks like color has been updated at some point since it is being used in our `CI.yml`

.github/workflows/content-release.yml
```diff
@@ -334,13 +331,9 @@ jobs:
...
 
-      - name: Echo next version
-        run: echo "${{ steps.bump-tag-version.outputs.v_patch }}"
-
-      # TODO: Uncomment when ready for prod (See https://github.com/department-of-veterans-affairs/devops/blob/67bb3bcc3ade75ffb311c47774f492b7f9ff7c65/ansible/Jenkinsfiles/releases/content-build#L28-L60 for current behavior)
-      # - name: Tag release
-      #   if: ${{ needs.set-env.outputs.BUILDTYPE == "vagovprod" }}
-      #   run: git tag ${{ steps.bump-tag-version.outputs.v_patch }} ${{ needs.set-env.outputs.REF }} && git push origin ${{ steps.bump-tag-version.outputs.v_patch }} 
+      - name: Tag release
+        if: ${{ needs.set-env.outputs.BUILDTYPE == "vagovprod" }}
```
oseasmoran73, line 335: Only push release created if `prod`

.github/workflows/content-release.yml
```diff
@@ -99,10 +98,9 @@ jobs:
         if: ${{ env.DSVA_SCHEDULE_ENABLED != 'true' }}
         uses: andymckay/cancel-action@0.2
 
-      # TODO: add prod when ready
       - name: Validate workflow_dispatch deploy_environment
         if: ${{ github.event_name == 'workflow_dispatch' && !contains(fromJson('["dev", "staging"]'), github.event.inputs.deploy_environment) }}
```
indytechcook, line 102:
```suggestion
        if: ${{ github.event_name == 'workflow_dispatch' && !contains(fromJson('["dev", "staging", "prod"]'), github.event.inputs.deploy_environment) }}
```

.github/workflows/content-release.yml
```diff
@@ -99,10 +98,9 @@ jobs:
...
-        run: echo "dev and staging are the only supported environments" && exit 1
+        run: echo "dev, staging and prod are the only supported environments" && exit 1
```
indytechcook, line 103:
```suggestion
        run: echo "dev, staging, and prod are the only supported environments" && exit 1
```

.github/workflows/content-release.yml
indytechcook, line 102:
```suggestion
        if: ${{ github.event_name == 'workflow_dispatch' && !contains(fromJson('["dev", "staging", "prod"]'), github.event.inputs.deploy_environment) }}
```

indytechcook (changes requested):
Mostly Looks great.  I just added `prod` to the allowed environments for manual deploy

indytechcook:
@oseasmoran73 can we also add an if at the `notify-start:` to only run on prod https://github.com/department-of-veterans-affairs/content-build/pull/498/files#diff-90571dbe8d8f4805db215f91f9d177538a3043d2d07cf1fcd29199bee34cfa79R172.  

`

indytechcook:
Cross linking the CMS issue which will turn on GHA for the CMS triggered release: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/6906

indytechcook:
Cross linking testing issue: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6711

indytechcook:
Documenting a Slack conversation regarding this: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6711#issuecomment-961940533

Neil: 

> do you know if it's possible to limit which steps the cancel is allowed to run in?
> With the builds taking nearly 35 minutes (looking at the history) this feels like a pretty common scenario

Oseas:

> Potentially (leaning more on the side of no). I know we can add concurrency at a jobs.job level, so I am curious if we can override the cancel-in-progress flag it it gets the same group name. Where would we want to add, if it is possible?

Neil: 

> good question.  I'd say after the build step sometime would make the most sense.  Do you think?

Oseas:

> Ooooh! An idea came to mind as a work around! We can add the always() flag. Can try to create small POC if needed/want
> Makes sense. I would probably say in the archive step if we had to pin point to one. That is officially when it is interacting with s3 for the "first time"
> What it sounds like, must always be executed, thats how we do to notify-failures
> Normal flow is if fails, all subsquent one will not trigger, but with always() with combination of if statement, we can enforce

oseasmoran73:
> @oseasmoran73 can we also add an if at the `notify-start:` to only run on prod https://github.com/department-of-veterans-affairs/content-build/pull/498/files#diff-90571dbe8d8f4805db215f91f9d177538a3043d2d07cf1fcd29199bee34cfa79R172.
> 
> `

@indytechcook Do we also want to have this for `notify-failure`?

indytechcook:
> @indytechcook Do we also want to have this for `notify-failure`?

Yeah.  That would be great.  Thanks @oseasmoran73

indytechcook (approved):
Given testing in https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6711 along with extra flow slack testing with @oseasmoran73 , this looks good!

timwright12 (approved):

PR #508, opened 2021-09-02
DT-26905 Added underline for links per a11y request

zacharymorel:
## Issue
https://github.com/department-of-veterans-affairs/va.gov-team/issues/26905

## Description
Adding underline to solve the issue with the usages of color as the only method to provide information.

## Testing done
Built and spun up local env.

## Screenshots
Screen Shot 2021-09-02 at 12 12 46 PM
Screen Shot 2021-09-02 at 11 05 06 AM

## Acceptance criteria
- [x] Added underline for links

## Definition of done

### DIFF
--- a/src/site/paragraphs/linkTeaser.drupal.liquid
+++ b/src/site/paragraphs/linkTeaser.drupal.liquid
@@ -37,7 +37,7 @@
   {% if header === "h3" %}
     {% if link.title != empty %}
         <h3 class="{{ headerClass }}">
-            <a href="{{link.url.path}}"
+            <a href="{{link.url.path}}" class="vads-u-text-decoration--underline"
             {% if linkTeaser.options["target"] %}target="{{ linkTeaser.options["target"] }}"{% endif %}
             >
             {{ link.title }}
@@ -45,7 +45,7 @@
         </h3>
     {% endif %}
   {% else %}
-    <a href="{{link.url.path}}"
+    <a href="{{link.url.path}}" class="vads-u-text-decoration--underline"
     {% if linkTeaser.options["target"] %}target="{{ linkTeaser.options["target"] }}"{% endif %}
     >
         {% if link.title != empty %}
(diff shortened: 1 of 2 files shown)


src/site/paragraphs/linkTeaser.drupal.liquid
```diff
@@ -37,15 +37,15 @@
   {% if header === "h3" %}
     {% if link.title != empty %}
         <h3 class="{{ headerClass }}">
-            <a href="{{link.url.path}}"
+            <a href="{{link.url.path}}" class="vads-u-text-decoration--underline"
```
mpelzsherman, line 40: If we remove `text-decoration: none` from `.va-nav-linkslist-list a` in vets-website, we won't need to explicitly add the underline back here. I'd rather not add styles we don't need.

timwright12 (approved):

src/site/paragraphs/linkTeaser.drupal.liquid
zacharymorel: That's a good point. The reason why I did not change the property on the class is because of how many other nodes are using the class with which some have anchor tags within them.
Screen Shot 2021-09-02 at 2 44 28 PM

zacharymorel: @mpelzsherman Any thoughts on above or we're okay to merge?

mpelzsherman: You're right, @zacharymorel - changing `va-nav-linkslist-list` is beyond the scope of this ticket.

I'm actually having a hard time finding exactly where the `text-decoration: none` is coming from here.

Maybe the public websites team should investigate this at some point though (@kelsonic @brianalloyd)? If we have links without underlines elsewhere on the site, that's a a11y violation, and we shouldn't have to override system-wide CSS to fix this one page at a time.

mpelzsherman (approved):

PR #521, opened 2021-09-08
Add sattelite-vet-center ID's to vet-center-nearby widget stub

mpelzsherman:
These will be used by the nearbyVetCenter widget for deduping.
Related to https://github.com/department-of-veterans-affairs/va.gov-team/issues/27870

Locations___Escanaba_Vet_Center___Veterans_Affairs

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -563,7 +563,13 @@ module.exports = function registerFilters() {
 
   // get a value from a path of an object
   // works for arrays as well
-  liquid.filters.getValueFromObjPath = (obj, path) => _.get(obj, path);
+  liquid.filters.getValueFromObjPath = (obj, path) => {
+    if (Array.isArray(obj)) {
+      return obj.map(e => _.get(e, path));
+    } else {
+      return _.get(obj, path);
+    }
+  };
 
   // get a value from a path of an object in an array
   liquid.filters.getValueFromArrayObjPath = (entities, index, path) =>
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -561,15 +561,7 @@ module.exports = function registerFilters() {
   // sort a list of objects by a certain property in the object
   liquid.filters.sortObjectsBy = (entities, path) => _.sortBy(entities, path);
 
-  // get a value from a path of an object
-  // works for arrays as well
-  liquid.filters.getValueFromObjPath = (obj, path) => {
-    if (Array.isArray(obj)) {
-      return obj.map(e => _.get(e, path));
-    } else {
-      return _.get(obj, path);
-    }
-  };
+  liquid.filters.getValueFromObjPath = (obj, path) => _.get(obj, path);
 
   // get a value from a path of an object in an array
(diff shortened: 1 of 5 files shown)


src/site/filters/liquid.js
```diff
@@ -563,7 +563,13 @@ module.exports = function registerFilters() {
 
   // get a value from a path of an object
   // works for arrays as well
-  liquid.filters.getValueFromObjPath = (obj, path) => _.get(obj, path);
```
Rleahy22, line 566: Did this ever work for arrays?

src/site/filters/liquid.js
```diff
@@ -563,7 +563,13 @@ module.exports = function registerFilters() {
...
+  liquid.filters.getValueFromObjPath = (obj, path) => {
+    if (Array.isArray(obj)) {
+      return obj.map(e => _.get(e, path));
```
Rleahy22, line 568: I guess I'm not sure that this is the behavior I would expect for a function called `getValueFromObjPath`. I think it might be clearer if there was a separate filter for this behavior called `getValuesFromArrayByKey`, and we remove the comment about `getValueFromObjPath` working for arrays if that's inaccurate.

src/site/filters/liquid.js
mpelzsherman: No.

src/site/filters/liquid.js
mpelzsherman: Great suggestion, thanks!

src/site/filters/liquid.js
kelsonic: Can we remove the comment on line 565 then? 🙂

src/site/filters/liquid.js
```diff
@@ -563,7 +563,13 @@ module.exports = function registerFilters() {
 
   // get a value from a path of an object
   // works for arrays as well
```
kelsonic, line 565:
```suggestion
  
```

src/site/filters/liquid.js
```diff
@@ -563,7 +563,13 @@ module.exports = function registerFilters() {
...
-  liquid.filters.getValueFromObjPath = (obj, path) => _.get(obj, path);
+  liquid.filters.getValueFromObjPath = (obj, path) => {
+    if (Array.isArray(obj)) {
+      return obj.map(e => _.get(e, path));
+    } else {
```
kelsonic, line 569: You don't need `else` since you already return in the if statement above 🙂

src/site/filters/liquid.js
mpelzsherman: So after closer inspection of the liquid filters, I discovered we already had `getValuesForKey` which did what I needed.

I renamed it to `getValuesForPath`.

There are already too many filter functions. I think we need to do some housecleaning and resist the temptation to just keep adding new functions to this file willy-nilly.

src/site/filters/liquid.js
mpelzsherman: done!

Rleahy22 (approved):

gescobardev (approved):

PR #527, opened 2021-09-09
Github Actions (CI, content-release & daily-deploy) -- Fix caching

oseasmoran73:
## Description

This PR addresses the [following](https://github.com/department-of-veterans-affairs/va.gov-team/issues/28585#issuecomment-907288548)

Since we have changed to the on-demand runners, caching was not working as intended as the pathing has changed. This PR fixes by adding news keys specific to on-demand runners

## Testing done

[Yesterdays Run](https://github.com/department-of-veterans-affairs/content-build/actions/runs/1219436708)
Current run

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -238,8 +238,8 @@ jobs:
         with:
           path: |
             .cache/yarn
-            node_modules
-          key: ${{ steps.get-node-version.outputs.NODE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
+            **/node_modules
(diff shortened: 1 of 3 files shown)


alexandec:
Looks good! I can see that the cache keys have hashes, and that the cache is being used for the build and the E2E tests.

alexandec (approved):

ddzz (approved):
I'm guessing the same will be done in `vets-website`?

oseasmoran73:
@ddzz It is slightly different in `vets-website` as we are not using the on-demand runners on there. It does follow similar concept `{NODE_VERSION}-{HASH}`.

We had to do this because on-demand runner root is `/action-runner`, where as in vets-website at `/home/runners`

PR #533, opened 2021-09-13
Vet center call center informational alert

gescobardev:
## Description
Resolves issue linked below

To test you can use Colorado Springs VC - `/colorado-springs-vet-center`
## Screenshots

<details>
<summary>Screenshot of new call center info alert</summary>

Screen Shot 2021-09-13 at 9 05 36 AM

</details>

### DIFF
--- a/src/site/components/informational_alert.drupal.liquid
+++ b/src/site/components/informational_alert.drupal.liquid
@@ -1,0 +1,14 @@
+<div class="vads-u-margin-bottom--2" id="field-cc-vet-call-center">
+  <va-alert status="info" id="va-info-alert">
+    <h3 slot="headline">
+      Need help after hours?
+    </h3>
+    <div>
+      {% if entity.fieldWysiwyg.processed %}
+      {{ entity.fieldWysiwyg.processed | drupalToVaPath | phoneLinks }}
+    {% elsif entity.fieldWysiwyg.0.processed %}
+      {{ entity.fieldWysiwyg.0.processed | drupalToVaPath | phoneLinks }}
+    {% endif %}
+    </div>
+  </va-alert>
+</div>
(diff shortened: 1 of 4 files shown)


src/site/components/informational_alert.drupal.liquid
```diff
@@ -0,0 +1,14 @@
+<div class="vads-u-margin-bottom--2" id="field-cc-vet-call-center">
+  <va-alert status="info" id="va-info-alert">
+    <h3 slot="headline">
+      Need help after hours?
```
gescobardev, line 4: Alert headline is hardcoded per original issue AC.

src/site/components/informational_alert.drupal.liquid
```diff
@@ -0,0 +1,14 @@
...
+    </h3>
+    <div>
+      {% if entity.fieldWysiwyg.processed %}
+      {{ entity.fieldWysiwyg.processed | drupalToVaPath | phoneLinks }}
+    {% elsif entity.fieldWysiwyg.0.processed %}
+      {{ entity.fieldWysiwyg.0.processed | drupalToVaPath | phoneLinks }}
+    {% endif %}
```
mpelzsherman, line 11: Could we use the [processDynamicContent](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/filters/liquid.js#L731) filter here instead? I'd rather keep this kind of logic there instead of in the templates.

gescobardev: I looked it over and refactored it. Let me know what you think.
I think using `processDynamicContent` would add extra streps.
We're also checking for different datatypes when we should just be checking that `fieldWysiwyg` is an array

mpelzsherman: processDynamicContent was intended to be used for dynamic content blocks, which may contain data from different sources. It's not needed here since the data is always coming from the same source.

mpelzsherman (approved):

PR #535, opened 2021-09-13
Fix Cypress a11y test `yarn install`

ddzz:


pjhill (approved):

.github/workflows/a11y.yml
```diff
@@ -94,16 +90,13 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            ~/.cache/yarn
-            node_modules
-          key: ${{ steps.get-node-version.outputs.NODE_VERSION }}-${{ hashFiles('yarn.lock') }}
+            .cache/yarn
```
alexandec, line 93: You may want to add `/github/home/.cache/Cypress` here also since Cypress stores its binary in a separate location. For the other Cypress tests we had to add that: https://github.com/department-of-veterans-affairs/content-build/blob/master/.github/workflows/continuous-integration.yml#L459

.github/workflows/a11y.yml
```diff
@@ -94,16 +90,13 @@ jobs:
...
+            **/node_modules
+          key: ${{ steps.get-node-version.outputs.NODE_VERSION }}-on-demand-runner-${{ hashFiles('**/yarn.lock') }}
```
alexandec, line 95: We also used a different key, `on-demand-runner-cypress` (see previous link)

alexandec (approved):

PR #565, opened 2021-09-20
Update vet center styles to use VADS

gescobardev:
## Description
Resolves issue linked below


mpelzsherman (approved):

mpelzsherman (changes requested):
On closer inspection of the [review environment,](http://5121e24a2924408dfef45af2dbda8d55.review.vetsgov-internal/escanaba-vet-center/) I am seeing a few differences from the original design that should probably be addressed.

Please see screen shots and work with @lildziner or @nicko-amida  to identify which changes are OK, and which need to be corrected.

mpelzsherman:
Escanaba_Vet_Center___Veterans_Affairs

lildziner:
the preference would be for the “single spacing” between the bullets and equal margin above and below the divider line. There are a few variations of the in the spotlight box out there and all a little different. This might be a good reference for the bullet spacing and this one for margin height above/below divider

mpelzsherman:
Looks good to me locally - just resolve the conflict and we should be good to go! We should confirm in the review environment though.

mpelzsherman (approved):

PR #576, opened 2021-09-21
Fix "Getting Started" link + add --use-cached-assets

alexandec:
## Description
Update README to fix  "Getting Started" link. Also add a mention of the `--use-cached-assets` flag. See [slack discussion](https://dsva.slack.com/archives/C01512KN35G/p1632240331078300) for more info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 This repository contains scripts and templates which generate static HTML pages for the VA.gov site.
 
-There are several repositories that contain the code and content used to build VA.gov. If you're looking to get started running VA.gov locally, you should read the [Getting Started](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started) documentation.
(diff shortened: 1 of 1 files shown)


alexandec:
@timwright12 @mchelen-gov a couple small content-build README updates here

README.md
```diff
@@ -25,7 +25,7 @@ on VA.gov, you'll need to build these static pages using the following commands:
 
 `yarn build` (fetches the latest content cache from S3 by default when cache is empty)
 
-- use `--pull-drupal` to fetch fresh content from Drupal if needed (requires SOCKS proxy access)
+- use `--pull-drupal` to fetch fresh content from Drupal if needed (requires SOCKS proxy access). Add `--use-cached-assets` to skip asset download
```
timcosgrove, line 28: I think this helps. Just so I'm clear, this won't change the time downloading the initial assets; it just prevents the build from deleting the cached download once it's extracted. Is that right?

alexandec: Yeah, if you don't have any assets in `.cache/ENV/drupal/downloads`, the assets will be downloaded whether you use `--use-cached-assets` or not. But if you use the flag and you have cached assets, the cached assets will be used instead of downloading the assets.

k80bowman (approved):

PR #587, opened 2021-09-22
Github Actions (CI) -- Optimize archive compression

oseasmoran73:
## Description

Related [ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/28585#issuecomment-902232486)

With the current compression the time takes roughly `15 mins`. This new method should drastically shave time and also affect the `deploy` step as well.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -670,7 +670,7 @@ jobs:
           du -h build/${{ matrix.buildtype }}
 
       - name: Compress build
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -210,6 +210,7 @@ jobs:
         run: |
           cat > build/${{ matrix.buildtype }}/BUILD.txt << EOF
           BUILDTYPE=${{ matrix.buildtype }}
+          NODE_ENV=production
```
oseasmoran73, line 213: Fix dashboard not displaying correctly

theodur (approved):

PR #590, opened 2021-09-22
Fix Skip to Content link

mpelzsherman:
Resolves department-of-veterans-affairs/va.gov-team#27445

See also: https://github.com/department-of-veterans-affairs/vets-website/pull/18823

TL:DR - the Skip to Content link currently skips to the breadcrumbs, which is confusing for non-sighted users.
This change makes it skip to the first H1 on the page, if one exists. If not, the current behavior is preserved.
To access the Skip to Content link you must use a screen reader navigator or hit shift-TAB several times.


tlei123 (approved):
LGTM!  Facility Locator's **Skip to content** link now skips to the **h1**.

mdewey (approved):
LGTM

Rleahy22 (approved):

PR #612, opened 2021-09-30
Standardize Vet Center 

gescobardev:
## Description
Resolves issue linked below

### DIFF
--- a/src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
+++ b/src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
@@ -1,21 +1,25 @@
 <div data-template="paragraphs/q_a.collapsible_panel">
-  <va-accordion>
+  <va-accordion bordered>
     {% assign questions = entity.fieldQuestions %}
-    {% assign sectionHeader = entity.fieldSectionHeader %}
+    {% assign sectionHeader = entity.fieldSectionHeader.0.value %}
 
     {% for accordionItem in questions %}
       {% assign item = accordionItem.entity %}
-      {% assign id = item.entityId %}
+      {% assign id = item.targetId %}
 
-      <va-accordion-item header="{{ item.fieldQuestion }}">
+      <va-accordion-item
+        id="cc-question-accordion-item-{{ id }}"
+        header="{{ item.fieldQuestion.0.value }}" 
+        level="3"
+      >
         <div
-          id={{id}}
+          id={{ id }}
           data-template="paragraphs/q_a.collapsible_panel__qa"
-          data-entity-id="{{ item.entityId }}"
+          data-entity-id="{{ id }}"
           data-analytics-faq-section="{{ sectionHeader | escape }}"
-          data-analytics-faq-text="{{ item.fieldQuestion | escape }}"
+          data-analytics-faq-text="{{ item.fieldQuestion.0.value | escape }}"
         >
-          <div id="{{ item.entityBundle }}-{{ item.entityId }}">
+          <div id="{{ item.entityBundle }}-{{ item.targetId }}">
             <div itemprop="acceptedAnswer" itemscope="" itemtype="http://schema.org/Answer">
               {% for answer in item.fieldAnswer %}
                 {% assign bundleComponent = "src/site/paragraphs/" | append: answer.entity.entityBundle | append: ".drupal.liquid" %}
--- a/src/site/layouts/vet_center.drupal.liquid
+++ b/src/site/layouts/vet_center.drupal.liquid
@@ -158,9 +158,8 @@
               <p> {{ fieldCcVetCenterFaqs.fetched.fieldSectionIntro.0.value }} </p>
             {% endif %}
             <div class="vads-u-margin-bottom--3" id="field-vet-center-faqs">
-              {% include "src/site/paragraphs/facilities/q_a.accordion_panel.drupal.liquid" with
-                  fetched = fieldCcVetCenterFaqs.fetched
-                  fetchedBundle = fieldCcVetCenterFaqs.fetchedBundle
+              {% include "src/site/paragraphs/q_a.collapsible_panel.drupal.liquid" with
+                  entity = fieldCcVetCenterFaqs.fetched
               %}
             </div>
           {% endif %}
@@ -171,7 +170,6 @@
   </main>
   <div id="vet-center-up-to-top" onclick="recordEvent({ event: 'back-to-top'});">
     {% include "src/site/components/up_to_top_button.html" %}
-  </div>
 </div>
 
 {% include "src/site/includes/footer.html" %}


src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
```diff
@@ -1,21 +1,25 @@
...
+      >
         <div
-          id={{id}}
+          id={{ id }}
           data-template="paragraphs/q_a.collapsible_panel__qa"
-          data-entity-id="{{ item.entityId }}"
+          data-entity-id="{{ id }}"
           data-analytics-faq-section="{{ sectionHeader | escape }}"
-          data-analytics-faq-text="{{ item.fieldQuestion | escape }}"
+          data-analytics-faq-text="{{ item.fieldQuestion.0.value | escape }}"
```
mpelzsherman, line 20: I'm concerned that this will break analytics on other pages. We should normalize the data using the `processDynamicContent` filter function rather than modifying this template, which is used all over the site.

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
```diff
@@ -1,21 +1,25 @@
...
 
     {% for accordionItem in questions %}
       {% assign item = accordionItem.entity %}
-      {% assign id = item.entityId %}
+      {% assign id = item.targetId %}
 
-      <va-accordion-item header="{{ item.fieldQuestion }}">
+      <va-accordion-item
+        id="cc-question-accordion-item-{{ id }}"
+        header="{{ item.fieldQuestion.0.value }}"
```
mpelzsherman, line 12: I'm pretty sure this change will break other pages. We should normalize the data using the `processDynamicContent` filter function rather than modifying this template, which is used all over the site.

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
```diff
@@ -1,21 +1,37 @@
...
     {% assign sectionHeader = entity.fieldSectionHeader %}
 
     {% for accordionItem in questions %}
       {% assign item = accordionItem.entity %}
-      {% assign id = item.entityId %}
+      {% if item.entityId %}
+        {% assign id = item.entityId %}
+      {% elsif item.targetId %}
+        {% assign id = item.targetId %}
+      {% endif %}
```
gescobardev, line 12: Assigning `id` to either `entityId` or `targetId`
`targetId `is what was utilized with the prior accordion component (`q_a.accordion_panel`)

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
gescobardev: Agreed.

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
gescobardev: Agreed.

src/site/filters/liquid.js
```diff
@@ -745,6 +745,21 @@ module.exports = function registerFilters() {
           };
         }
       }
+      case 'q_a_section': {
+        if (
+          entity.fieldAccordionDisplay &&
+          entity.fieldSectionHeader &&
+          entity.fieldSectionIntro &&
+          entity.fieldQuestions
```
mpelzsherman, line 753: You might want to check whether these fields contain arrays vs scalar values. That may not always be the case.
Also I'm not sure all of these fields are always going to be present in the data.

The above probably doesn't matter now because you're only using `dynamic-content.drupal` for centralized `q_a_section` content, but that may not always be the case. We should rename it to `centralized-content.drupal` if we only intend to ever use it for centralized content, and perhaps rename `processDynamicContent` to `processCentralizedContent` as well.

I'm ashamed to see that I didn't write a unit test for `processDynamicContent`. Would you mind adding one?

Please check to see if the eslint-disable comment on line 733 can be removed now that we have a 2nd case, thanks!

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
mpelzsherman: Are you sure we need this logic here? Can we standardize the data in `processDynamicContent` instead?

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
```diff
@@ -1,21 +1,37 @@
...
           data-template="paragraphs/q_a.collapsible_panel__qa"
-          data-entity-id="{{ item.entityId }}"
+          data-entity-id="{{ id }}"
           data-analytics-faq-section="{{ sectionHeader | escape }}"
-          data-analytics-faq-text="{{ item.fieldQuestion | escape }}"
+          {% if item.fieldQuestion.0.value %}
+            data-analytics-faq-text="{{ item.fieldQuestion.0.value | escape }}"
+          {% elsif item.fieldQuestion %}
+            data-analytics-faq-text="{{ item.fieldQuestion | escape }}"
+          {% endif %}
```
mpelzsherman, line 32: This check shouldn't be needed now that we're standardizing the data in `processDynamicContent`.

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
```diff
@@ -1,21 +1,37 @@
...
 
-      <va-accordion-item header="{{ item.fieldQuestion }}">
+      <va-accordion-item
+        id="cc-question-accordion-item-{{ id }}"
+        {% if item.fieldQuestion.0.value %}
+          header="{{ item.fieldQuestion.0.value }}"
+        {% elsif item.fieldQuestion %}
+          header="{{ item.fieldQuestion }}"
+        {% endif %}
+        level="3"
```
mpelzsherman, line 21: Are you sure it's OK to hard code the level here?

gescobardev: Hmm on staging, I'm seeing that they are both 3 in the tugboat instance and staging

gescobardev: Adding a level param to the component. Thank you for catching that!

src/site/filters/liquid.js
gescobardev: I have renamed `dynamic-content.drupal` and `processDynamicContent` as suggested.
Unit tests have been added for `processCentralizedContent` filter.
Removed line 733 !

src/site/filters/liquid.unit.spec.js
```diff
@@ -1488,3 +1488,97 @@ describe('getValueFromObjPath', () => {
...
+
+    expect(
+      liquid.filters.processCentralizedContent(
+        testData,
+        vetCenterData.fieldCcVetCenterCallCenter.fetchedBundle,
+      ),
+    ).to.deep.eq(testData);
+  });
+
+  it('returns expected if contentType = wysiwyg', () => {
```
mpelzsherman, line 1523:
```suggestion
  it('converts fieldWysiwyg.processed from snake case/array to camel case/string`, () => {
```

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
gescobardev: update component to use this filter

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
gescobardev: Yes, thank you. I have updated this.

src/site/filters/liquid.js
```diff
@@ -745,6 +745,36 @@ module.exports = function registerFilters() {
...
+          };
+        else return entity;
+      }
+      case 'q_a': {
+        if (
+          entity.fieldQuestion &&
+          entity.fieldAnswer &&
+          entity.entityBundle &&
+          entity.targetId
+        ) {
```
mpelzsherman, line 769: It still looks to me like `all` of these fields must be populated for the code to use the array format. I'm not sure that's the right approach here.

src/site/includes/centralized-content.drupal.liquid
```diff
@@ -1,4 +1,4 @@
 {% assign bundleComponent = "src/site/paragraphs/" | append: contentType %}
```
mpelzsherman, line 1: Let's add a comment here to explain what this component does and why it is needed, e.g.

> Centralized content is content that is re-used across all of VA.gov. It has more restrictive editing permissionsin Drupal, and it can have any bundle type. Centralized content uses a different data format than regular content, so we use the `processCentralizedContent` filter function to normalize the data, allowing our standard templates to display it.

src/site/filters/liquid.unit.spec.js
```diff
@@ -1488,3 +1488,97 @@ describe('getValueFromObjPath', () => {
...
+      fieldQuestions: [
+        {
+          label: 'National Vet Center content > Content > Questions',
+          fieldAnswer: [],
+          fieldQuestion: [],
+        },
+      ],
+      fieldSectionHeader: [{ value: "How we're different than a clinic" }],
+      fieldSectionIntro: [{ value: 'Click on a topic for more details.' }],
+    };
```
mpelzsherman, line 1550: I'd like to see at least one test where `fieldAnswer` and `fieldQuestion` are populated. You can copy data from https://staging.va.gov/escanaba-vet-center/ (`fieldCcVetCenterFaqs`).

src/site/filters/liquid.js
gescobardev: changes made to resolve this

src/site/includes/centralized-content.drupal.liquid
gescobardev: comment added

src/site/filters/liquid.unit.spec.js
gescobardev: data added.

src/site/filters/liquid.js
```diff
@@ -728,9 +728,24 @@ module.exports = function registerFilters() {
...
+        if (Array.isArray(obj[key])) {
+          if (obj[key][0]?.value) {
+            newObj[key] = obj[key][0]?.value;
+          } else newObj[key] = obj[key];
+        } else {
+          newObj[key] = obj[key];
+        }
+      }
+      return newObj;
+    }
```
mpelzsherman, line 746: Nice! 🚀  😄

src/site/filters/liquid.js
```diff
@@ -745,6 +760,15 @@ module.exports = function registerFilters() {
           };
```
mpelzsherman, line 760: Can we use normalizeData for `wysiwyg` as well?

src/site/filters/liquid.unit.spec.js
```diff
@@ -1488,3 +1488,150 @@ describe('getValueFromObjPath', () => {
...
+
+    expect(
+      liquid.filters.processCentralizedContent(testData, 'q_a'),
+    ).to.deep.eq(expected);
+  });
+
+  it('returns test data if contentType does not match any of the cases - default', () => {
+    expect(
+      liquid.filters.processCentralizedContent('test', 'test_bundle'),
+    ).to.eq('test');
```
mpelzsherman, line 1635: Maybe use an object here instead of a string?

src/site/filters/liquid.js
```diff
@@ -745,6 +760,15 @@ module.exports = function registerFilters() {
...
         }
       }
+      case 'q_a_section': {
+        return normalizeData(entity);
+      }
+      case 'q_a': {
+        if (entity.targetId && !entity.entityId) {
+          delete Object.assign(entity, { entityId: entity.targetId }).targetId;
```
mpelzsherman, line 768: I'm not exactly sure what's going on here...

src/site/filters/liquid.js
```diff
@@ -728,9 +728,24 @@ module.exports = function registerFilters() {
     });
   };
 
-  liquid.filters.processDynamicContent = (entity, contentType) => {
+  liquid.filters.processCentralizedContent = (entity, contentType) => {
+    if (!entity) return null;
+
```
mpelzsherman, line 733: Maybe add a comment? "normalizeData" is very self-documenting...

src/site/filters/liquid.js
```diff
@@ -728,9 +728,24 @@ module.exports = function registerFilters() {
...
+    function normalizeData(obj) {
+      const newObj = {};
+      for (const [key] of Object.entries(obj)) {
+        if (Array.isArray(obj[key])) {
+          if (obj[key][0]?.value) {
```
mpelzsherman, line 738: Can we pass in `value` as an arg?

src/site/filters/liquid.js
```diff
@@ -728,9 +728,24 @@ module.exports = function registerFilters() {
     });
   };
 
-  liquid.filters.processDynamicContent = (entity, contentType) => {
+  liquid.filters.processCentralizedContent = (entity, contentType) => {
+    if (!entity) return null;
+
+    function normalizeData(obj) {
```
mpelzsherman, line 734: Use es6 syntax?

src/site/filters/liquid.unit.spec.js
```diff
@@ -1488,3 +1488,150 @@ describe('getValueFromObjPath', () => {
...
+      entityType: 'paragraph',
+      entityBundle: 'q_a',
+      pid: '29909',
+      label: 'National Vet Center content > Content > Questions',
+      status: true,
+      langcode: 'en',
+      fieldAnswer: [{ entity: [{}] }],
+      fieldQuestion: 'What about my privacy?',
+    };
+
```
mpelzsherman, line 1626: Consider removing the unchanged fields from these tests, or put the fields that aren't changing in a variable to show that they remain unchanged.

src/site/paragraphs/q_a_section.drupal.liquid
```diff
@@ -19,7 +19,7 @@
     <p>{{ entity.fieldSectionIntro }}</p>
   {% endif %}
   {% if entity.fieldAccordionDisplay %}
-      {% include "src/site/paragraphs/q_a.collapsible_panel.drupal.liquid" %}
+      {% include "src/site/paragraphs/q_a.collapsible_panel.drupal.liquid" entity = entity %}
```
mpelzsherman, line 22: The fact that this was necessary is scaring me. Liquid variables do get passed to their children, but entity is very common/generic. Maybe consider renaming `entity` inside q_a.collapsible_panel.drupal to avoid scoping confusion?

gescobardev: Thanks for noting this! I removed it and it is actually working now. I think initially, it was not working which is why I added it but looks like that was a separate issue.

src/site/filters/liquid.js
gescobardev: Currently, no. But I can change it up so that it can.

src/site/filters/liquid.js
```diff
@@ -7,6 +7,7 @@ const moment = require('moment-timezone');
 const set = require('lodash/fp/set');
 // Relative imports.
 const phoneNumberArrayToObject = require('./phoneNumberArrayToObject');
+const renameKey = require('../../platform/utilities/data/renameKey').renameKey;
```
mpelzsherman, line 10: I _think_ if you do this in renameKey.js:

```js
module.exports = renameKey
```

... you can omit the redundant `.renameKey` from the `require` here.

mpelzsherman (approved):

gescobardev:
@kelsonic @Rleahy22 Do you happen to know what other pages utilize the `q_a.collapsible_panel.drupal.liquid` accordion?
So far I've been testing my changes with
https://staging.va.gov/nashville-vet-center/
https://staging.va.gov/health-care/about-va-health-benefits/

When you get the chance, please lmk :)

kelsonic:
Hey @gescobardev! 👋 So unfortunately there are a TON of places that page is used, probably hundreds if not thousands 😂 

So q_a is used by every q_a_section, and running that through a graphql query with:

```graphql
{
  paragraphQuery(limit: 1000, offset: 0, filter: {conditions: [{field: "type", value: ["q_a_section"]}]}) {
    entities {
      entityId
      entityBundle
      ... on ParagraphQASection {
        entityId
        entityOwner {
          entityBundle
          entityId
        }
      }
    }
  }
}
```

You get this:

[q_a_sections.txt](https://github.com/department-of-veterans-affairs/content-build/files/7289994/q_a_sections.txt)

I didn't have time tonight but then with those q_a_sections, you would collect each owner ID and then make a query to get all of those (hopefully they are not child nodes and are actually pages, otherwise you may have to continue following the rabbit hole lol)

This probably isn't the straight-forward answer you were hoping for 😂 😂 😂

src/site/paragraphs/q_a.collapsible_panel.drupal.liquid
```diff
@@ -1,21 +1,27 @@
 <div data-template="paragraphs/q_a.collapsible_panel">
-  <va-accordion>
+  <va-accordion bordered>
     {% assign questions = entity.fieldQuestions %}
     {% assign sectionHeader = entity.fieldSectionHeader %}
 
     {% for accordionItem in questions %}
-      {% assign item = accordionItem.entity %}
+      {% assign item = accordionItem.entity | processCentralizedContent: accordionItem.entity.entityBundle %}
```
mpelzsherman, line 7: Since we are already calling `processCentralizedContent` in centralized-content.drupal.liquid, I think we can omit it here.

mpelzsherman:
@kelsonic We're aware that `q_a.collapsible_panel.drupal.liquid` is used all over the place. The point here is to bring Vet Centers in sync with the rest of VA.gov and use it as well. 

I wasn't expecting that we would need to make any changes to `q_a.collapsible_panel.drupal.liquid`. I'm fairly confident that the changes @gescobardev did make won't cause any issues, but we should try to keep those changes to an absolute minimum.

It's impossible to manually test every single page that uses `q_a.collapsible_panel.drupal.liquid`. All we need are a few examples to spot check.

@gescobardev - here are a few I found:

/health-care/about-va-health-benefits - nodeId 29
/health-care/how-to-apply - nodeId 35

mpelzsherman (changes requested):
Please remove all unnecessary changes to q_a.collapsible_panel, thanks.

src/platform/utilities/data/renameKey.js
```diff
@@ -0,0 +1,15 @@
...
+ * Given an object, you can rename a particular key with a new key.
+ *
+ *
+ * @param {Object} object
+ * @param {String} oldKey
+ * @param {String} newKey
+ */
+
+const renameKey = (obj, oldKey, newKey) => {
+  if (!obj) return null;
```
mpelzsherman, line 11: Should `oldKey` and `newKey` be required as well?

src/platform/testing/unit/renameKey.unit.spec.js
```diff
@@ -0,0 +1,26 @@
+import { expect } from 'chai';
+
+const renameKey = require('../../utilities/data/renameKey');
+
+describe('renameKey', () => {
+  it('returns null if null is passed', () => {
+    expect(renameKey(null, 'testKey1', 'newKey')).to.be.null;
+  });
```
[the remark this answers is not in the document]
gescobardev: updated function to handle this

src/platform/utilities/data/renameKey.js
gescobardev: I only call `renameKey` `if (targetId && !entityId)` to avoid that specific situation from happening. But I understand what you mean about it being destructive. I can update it so that it doesn't get rid of the targetId. 
This will mean that we will have instances where we have 2 keys (entityId and targetId) equal to the same value.

src/site/filters/liquid.js
```diff
@@ -728,9 +729,26 @@ module.exports = function registerFilters() {
...
+          if (obj[key][0][field]) {
+            newObj[key] = obj[key][0][field];
+          } else newObj[key] = obj[key];
+        } else {
+          newObj[key] = obj[key];
+        }
+      }
+      return newObj;
+    };
+
```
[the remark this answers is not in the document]
gescobardev: Sure, I can refactor.

mpelzsherman (approved):

src/platform/utilities/data/renameKey.js
```diff
@@ -0,0 +1,20 @@
...
+
+  if (typeof oldKey !== 'string' || typeof newKey !== 'string') {
```
Rleahy22, line 13: Are empty strings valid then? 
```javascript
renameKey({foo: 'bar}, foo, '')
```

src/site/filters/liquid.unit.spec.js
```diff
@@ -1488,3 +1488,164 @@ describe('getValueFromObjPath', () => {
...
+describe('processCentralizedContent', () => {
+  it('returns null if null is passed', () => {
+    expect(liquid.filters.processCentralizedContent(null, 'wysiwyg')).to.be
+      .null;
+  });
+
+  it('returns null if null is passed', () => {
+    expect(liquid.filters.processCentralizedContent(null, 'test_bundle')).to.be
+      .null;
+  });
```
Rleahy22, line 1501: is this basically the same test as the one above it?

gescobardev: Not quite. There is a specific case that handles `wysiwyg`
`test_bundle` handles a default scenario.

src/platform/testing/unit/renameKey.unit.spec.js
```diff
@@ -0,0 +1,63 @@
+import { expect, assert } from 'chai';
+
+const renameKey = require('../../utilities/data/renameKey');
+
+describe('renameKey', () => {
+  const testObj = { testKey1: 'test value 1' };
+
+  it('returns null if null is passed', () => {
```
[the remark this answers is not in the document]
gescobardev: thanks! I could definitely be more descriptive.

PR #631, opened 2021-10-06
In runCommand(), only exit when child exits abnormally

alexandec:
## Description
In its current implementation, `runCommand()` always causes the parent process to exit when the child process completes, regardless of exit code. This led to [an issue](https://dsva.slack.com/archives/CBU0KDSB1/p1633549113067800) where the parent process was forced to exit even though its own async task was still running.

This PR modifies `runCommand()` so it only causes the parent process to exit if the child process exits with an abnormal (non-zero) return code.

## Testing done
- Local testing of build and unit tests (successful and failing)
- Successful CI run

## Acceptance criteria
- [x] Parent process only exits if child exits with nonzero return code

## Definition of done

### DIFF
--- a/script/utils.js
+++ b/script/utils.js
@@ -2,10 +2,14 @@
 const { spawn, spawnSync } = require('child_process');
 const fs = require('fs-extra');
 
-const runCommand = (cmd, forcedExitCode = null) => {
+const runCommand = cmd => {
   const child = spawn(cmd, [], { shell: true, stdio: 'inherit' });
+
+  // If the child process exits abnormally, exit parent with the same code
   child.on('exit', code => {
-    process.exit(forcedExitCode === null ? code : forcedExitCode);
+    if (code) {
+      process.exit(code);
+    }
   });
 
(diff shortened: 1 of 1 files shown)


script/utils.js
```diff
@@ -2,10 +2,14 @@
 const { spawn, spawnSync } = require('child_process');
 const fs = require('fs-extra');
 
-const runCommand = (cmd, forcedExitCode = null) => {
```
theodur, line 5: It doesn't look like the `forcedExitCode` parameter is passed into in any of the `runCommand` function calls, so I think this is okay to remove it. I'm wondering why it was added in the first place if it isn't being used. Could be the one function call that was using it was removed 🤷

theodur (approved):
🚀 Should we update the `runCommand` function in `vets-website` also?

script/utils.js
alexandec: It is strange. I just checked in vets-website as well, and the `forcedExitCode` param isn't used there either. So yeah, either it was used in the past, or it was premature optimization that never ended up being necessary.

PR #636, opened 2021-10-08
Add workflow dispatch with commit SHA to CI workflow

ndsprinkle:
## Description
This adds the ability to manually kick off the continuous integration workflow in content-build. A commit SHA can optionally be provided to deploy a specific commit to dev/staging. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31084

## Acceptance criteria
- [ ] Can manually run the continuous integration workflow
- [ ] Can deploy a specific commit to dev/staging


oseasmoran73:
@ndsprinkle Would we want to re-run the entire pipeline? I was under the impression that if a commit is passed (due to `workflow_dispatch`), that `get-latest-run-number` would essentially kick off (under like 8 mins or something along those lines)? 

CC @timwright12 for clarification

But in regards to your question, I don't see an issue where it would cause issue. Looks good on paper.

PR #637, opened 2021-10-08
Github Actions (CI) -- composite actions

oseasmoran73:
## Description

As Github Action is maturing, we can now have reusable actions. This PR creates actions locally for the following:

- Installing dependencies
  - Able to pass cache_folder
  - Able to pass keys
- Slack notification
  - Release was created and no longer requires checkout of repo
  - Auth bots creds no longer in main pipeline

## Original issue(s)
department-of-veterans-affairs/va.gov-team#30247

## Testing done

Latest Run

### DIFF
--- a/.github/workflows/install/action.yml
+++ b/.github/workflows/install/action.yml
@@ -15,7 +15,7 @@ inputs:
     required: false
     default: ''
   path:
-    description: path for cache folders
+    description: path for actions/cache@v2
     required: false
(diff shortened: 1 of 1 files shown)


alexandec:
This is great, it really makes the config files easier to read! Could other steps like start-runner, stop-runner, validate-build-status, notify-start, etc be refactored in a similar way?

PR #644, opened 2021-10-12
VACMS-6176: Remove missed sort.

ethanteague:
## Description
Removes a sort that checks for previously removed fields.

## Testing done
Build


mpelzsherman (approved):
I'm not familiar enough with how alerts are working to give this a formal review, so I'll just say "LGTM!".

PR #646, opened 2021-10-12
Hardened VAMC Register for Care page (phase 1)

mpelzsherman:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/29494

To test, see https://web-nwlghnqkwe3ftxcgy830jkfejcxjswak.demo.cms.va.gov/alaska-health-care/register-for-care-0/


gescobardev:
Aside from the comments I left above, I believe we also need to add that bottom portion section that starts with "More about VA health care benefits". But I'm not sure if thats part of the issue you had with not knowing what the fields are.

gescobardev:
Everything else looks good !

PR #675, opened 2021-10-20
Parameterize the content release build with environment name

cweagans:
## Description

We're integrating the content-release workflow into the CMS (for the "Release Content" button) in https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6354. To do that, we need to be able to specify the environment that content is released to. This PR adds a new field to the manual trigger inputs to let us specify dev, stage, or prod.

Prod is disabled for the time being per your earlier commits to this file.

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -7,6 +7,10 @@ on:
         description: "Minutes to wait before creating release"
         required: false
         default: 5
+      deploy_environment:
+        description: "The environment to deploy content to"
+        required: true
+        default: "dev"
   # schedule:
   #   - cron: "0 13-16,20-21 * * 1-5"
   #   - cron: "45 17 * * 1-5"
@@ -309,13 +313,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: [set-env, create-release]
 
-    # TODO: Uncomment when ready for prod
-    env:
-      # ENVIRONMENT: vagovprod
-      # BUCKET: content.www.va.gov
-      ENVIRONMENT: vagovdev
-      BUCKET: content.dev.va.gov
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/content-release.yml
```diff
@@ -7,6 +7,10 @@ on:
         description: "Minutes to wait before creating release"
         required: false
         default: 5
+      deploy_environment:
+        description: "The environment to deploy content to"
```
oseasmoran73, line 11:
```suggestion
        description: The environment to deploy content to
```

.github/workflows/content-release.yml
```diff
@@ -7,6 +7,10 @@ on:
...
+        required: true
+        default: "dev"
```
oseasmoran73, line 13:
```suggestion
        default: dev
```

.github/workflows/content-release.yml
```diff
@@ -343,11 +342,32 @@ jobs:
           role-duration-seconds: 900
           role-session-name: vsp-frontendteam-githubaction
 
-      - name: Deploy
+      - name: Deploy DEV
```
oseasmoran73, line 345: I am wondering if it makes sense to move this logic into `set-env` and pass in information needed from `needs` via an output. It will make it a bit cleaner, we can pass in env associated and dest in a cleaner matrix format. Thoughts?

.github/workflows/content-release.yml
oseasmoran73: Yeah I saw that and something I overlooked 😅 , but if you can would be greatly appreciated! If not I can do a separate PR for it in the near future.

Unfortunately, everything is treated as a string

.github/workflows/content-release.yml
oseasmoran73: Reason I am a bit more inclined (and this may be lack of knowledge) is how it will work with the cron and manual dispatch. While yes bit of mapping involved, we wouldn't need to explicitly call similar jobs (example staging)

ex: `if: ${{ github.event.inputs.deploy_environment == "prod" }}` and `if: ${{ github.event.scheduled.cron == "prod" }}` (most likely incorrect syntax), assuming we don't want `||` involved

PR #715, opened 2021-10-29
Github Actions (CI) -- Add concurrency to content-release based off environment

oseasmoran73:
## Description

As discussed in [slack](https://dsva.slack.com/archives/C01512KN35G/p1635533884032500?thread_ts=1635529221.019000&cid=C01512KN35G) thread, CR workflow now queued based off of their environment

## Testing done

N/A, will need to merge first and test dispatch. Important thing to notice is 

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -15,6 +15,10 @@ on:
   #   - cron: "0 13-16,20-21 * * 1-5"
   #   - cron: "45 17 * * 1-5"
 
+concurrency:
+  group: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.deploy_environment || 'prod' }}
+  cancel-in-progress: false
+
 # TODO: Change to correct channel_id when ready
 env:
   CHANNEL_ID: C024DUC9S1K # -test-tim-accessibility-gha
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -17,7 +17,7 @@ on:
 
 concurrency:
   group: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.deploy_environment || 'prod' }}
-  cancel-in-progress: false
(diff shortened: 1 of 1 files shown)


.github/workflows/content-release.yml
```diff
@@ -15,6 +15,10 @@ on:
   #   - cron: "0 13-16,20-21 * * 1-5"
   #   - cron: "45 17 * * 1-5"
 
+concurrency:
+  group: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.deploy_environment || 'prod' }}
```
oseasmoran73, line 19: Translated to:

workflow_dispatch ? deploy_environment : prod

alexandec: Is there another way to write this line? Your translation is helpful, but without that it's hard to parse, or easy to misinterpret

alexandec (approved):

alexandec:
Would be good to add a link to the thread, and/or a bit more info about this change

.github/workflows/content-release.yml
oseasmoran73: Unfortunately there is no way. ideally I would like to have it in the syntax provided above, but GHA does not allow that type of format 😭

oseasmoran73:
After talking with CMS team, we want to add concurrency to content-release. Since dev and stage will be enabled with prod, they will be grouped with their respect environment.

The only flows that should be cancelled if a new one gets kicked off are for dev & staging. Prod should never cancel, but rather queue

.github/workflows/content-release.yml
```diff
@@ -15,6 +15,10 @@ on:
...
+  cancel-in-progress: ${{ github.event.inputs.deploy_environment != 'prod' && true || false }}
```
oseasmoran73, line 20: Translated to:

deploy_environment != prod ? true || false

If the environment is dev/stage, cancel all flows associated with group-name. Otherwise, if prod, queue and operate accordingly (much like Jenkins)

PR #721, opened 2021-11-01
Guard against null errors when computing operating status data

mpelzsherman:
## Description
Encountered an error while previewing content. This fixes the issue.

### DIFF
--- a/src/site/stages/build/drupal/page.js
+++ b/src/site/stages/build/drupal/page.js
@@ -388,22 +388,29 @@ function compilePage(page, contentData) {
       break;
     case 'health_care_local_facility':
     case 'vamc_operating_status_and_alerts':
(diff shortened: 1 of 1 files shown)


src/site/stages/build/drupal/page.js
```diff
@@ -388,29 +388,37 @@ function compilePage(page, contentData) {
       break;
     case 'health_care_local_facility':
     case 'vamc_operating_status_and_alerts':
-      mainOperatingStatusEntities = locationsOperatingStatus.entities.filter(
+      mainOperatingStatusEntities = locationsOperatingStatus?.entities.filter(
```
kelsonic, line 391: This will still throw on `.filter`

kelsonic (approved):

PR #723, opened 2021-11-02
Possible VAMC past events fix

mpelzsherman:
Possible fix for https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6800

### DIFF
--- a/src/site/stages/build/drupal/graphql/eventListingPage.graphql.js
+++ b/src/site/stages/build/drupal/graphql/eventListingPage.graphql.js
@@ -16,7 +16,7 @@ const eventListingPage = `
         {field: "status", value: "1", operator: EQUAL, enabled: $onlyPublishedContent}, 
         {field: "moderation_state", value: "archived", operator: NOT_EQUAL}, 
         {field: "type", value: "event"},
-        {field: "field_datetime_range_timezone", value: [$today], operator: SMALLER_THAN}]},
+        {field: "field_datetime_range_timezone.end_value", value: [$today], operator: SMALLER_THAN}]},
       sort: {field: "changed", direction: DESC}) {
           entities {
             ... on NodeEvent {


src/site/stages/build/drupal/graphql/eventListingPage.graphql.js
```diff
@@ -16,7 +16,7 @@ const eventListingPage = `
         {field: "status", value: "1", operator: EQUAL, enabled: $onlyPublishedContent}, 
         {field: "moderation_state", value: "archived", operator: NOT_EQUAL}, 
         {field: "type", value: "event"},
-        {field: "field_datetime_range_timezone", value: [$today], operator: SMALLER_THAN}]},
+        {field: "field_datetime_range_timezone.end_value", value: [$today], operator: SMALLER_THAN}]},
```
kelsonic, line 19: The code was previously expecting `field_datetime_range_timezone` which is an object, but `field_datetime_range_timezone.end_value` is a number so I think we'll get quite a few runtime errors with this, right? 🙂 I think we might be able to fix it without touching the graphql at all

kelsonic: Oh wow, jenkins passed though so maybe I'm wrong!

mpelzsherman: Yep! I confirmed this change in graphql explorer before I pushed up the PR.

kelsonic: Sweet! Did you test it out locally yet?

mpelzsherman: Nope. This is unplanned work. But I will circle back and do some more testing later if nobody else is able to.

PR #728, opened 2021-11-05
Limit number of link teasers

gescobardev:
## Description
Resolves - https://github.com/department-of-veterans-affairs/va.gov-team/issues/32451

Per Dave, there should be only 2 columns with 8 links max.
See slack thread: https://dsva.slack.com/archives/C0FQSS30V/p1636040104227200
Screen Shot 2021-11-04 at 10 03 34 PM

Limiting number of link to show the first 8. 

To test use this tugboat instance (it has a total of 9 links but only displays the first 8) - https://web-xzcrb121xmxquzas9jcd9p86dgwpkuam.demo.cms.va.gov/detroit-health-care/

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -1286,8 +1286,4 @@ module.exports = function registerFilters() {
       return `/health-care/${path}`;
     }
   };
-
-  // liquid.filters.sliceArray = (arr, startIndex, endIndex) => {
-  //   return _.slice(arr, startIndex, endIndex);
-  // };
 };
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -1285,4 +1285,8 @@ module.exports = function registerFilters() {
       return `/health-care/${path}`;
     }
   };
+
+  liquid.filters.sliceArray = (arr, startIndex, endIndex) => {
+    return _.slice(arr, startIndex, endIndex);
+  };
 };
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -249,8 +249,9 @@ module.exports = function registerFilters() {
   };
 
   //  liquid slice filter only works on strings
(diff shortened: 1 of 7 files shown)


src/site/filters/liquid.js
```diff
@@ -1285,4 +1285,8 @@ module.exports = function registerFilters() {
       return `/health-care/${path}`;
     }
   };
+
+  liquid.filters.sliceArray = (arr, startIndex, endIndex) => {
```
mpelzsherman, line 1289: We already have [sliceArrayFromStart](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/filters/liquid.js#L252), which could be modified to accept an `endIndex` argument.

gescobardev: Ahh true, I see it is already being used in 2 other files. Do you think modifying it in that way would have any negative affects where it is being used right now?

src/site/filters/liquid.js
```diff
@@ -249,8 +249,9 @@ module.exports = function registerFilters() {
   };
 
   //  liquid slice filter only works on strings
-  liquid.filters.sliceArrayFromStart = (arr, startIndex) => {
```
mpelzsherman, line 252: `sliceArrayFromStart` is still used in a couple of templates. Please change to `sliceArray`.

gescobardev: Are you sure? I thought I got them all.
Which templates? 
Aside from `src/site/paragraphs/facilities/list_of_link_teasers_facility.drupal.liquid`
I have changed it in `src/site/layouts/media_list_images.drupal.liquid` and `src/site/components/browse-by-audience.drupal.liquid`

mpelzsherman: Sorry - I must've been looking at an earlier commit. Looks like you got 'em all!

mpelzsherman (approved):

PR #736, opened 2021-11-09
added new Pre-check-in application

mdewey:
## Description
- Creating a new application for the pre-check-in experience. 

## Ticket
https://github.com/department-of-veterans-affairs/va.gov-team/issues/32533

### DIFF
--- a/src/applications/registry.json
+++ b/src/applications/registry.json
@@ -1132,6 +1132,15 @@
     }
   },
   {
+    "appName": "Pre-check-in",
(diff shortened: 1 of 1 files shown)


timwright12 (changes requested):
Just blocking this merge until the entrypoint files are on staging and dev, so the builds/deploys don't fail

mdewey:
The files exists on AWS, and the jobs have been kicked off again @timwright12

timwright12 (approved):

PR #776, opened 2021-11-23
Issue 32285

tmac9494:
## Description
Moves announcement banner root from footer to just below the header for a more accurate screen reader/tabbing experience.

Original Issue Number: 32285

## Testing done
Tested locally, no issues identified.


joshkimux:
@tmac9494 , currently reviewing this now. Pushing this as it is will likely trigger a moderate axe-core issue around content [the promo banner] not being nested within landmarks [e.g. the header].

kelsonic (changes requested):
Nice work! 🎉 Let's wait for @joshkimux's feedback in regards to what landmark we want the announcement-root to be located in for a11y purposes.

In the meantime, this PR will also need a follow-up PR in `vets-website` for the `dev-template`: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/landing-pages/dev-template.ejs#L229

joshkimux:
Hey folks, sorry for the delay, I spent some time exploring options. Long term, I think this is an issue that will need to be fixed at the design level as opposed to being remediated via code when exploration is done on redesigning the VA.gov homepage.

In the meantime, we can either:
- roll with what we have here (which will trigger an axe-core a11y warning that there is content _outside_ of a landmark since it was removed from the footer / contentinfo landmark) OR 
- append `role="status"` to the `div` container which should remove the axe-core a11y warning, but possibly cause confusion down the road when fixing this in the design (as using `role="status"` may not be the final design decision)

Hope this helps y'all. For additional context, I've attached potential design fixes I have been exploring.
image

lzapson (approved):
lgtm

swirtSJW:
@tmac9494 or @joshkimux  can either of you shed light on this old PR.   Is this being handled at the design system level or does the new public websites team need to pick this up?

laflannery:
I believe this is a project currently being worked on at the design system level by @humancompanion-usds [here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/725)

PR #823, opened 2021-12-07
Implement Official Name Display

gescobardev:
## Description
Resolves - https://github.com/department-of-veterans-affairs/va.gov-team/issues/33053

If a vet center has an official name that is different from its title, we display that official name right below the vet center title

To test use this tugboat instance - https://web-khwba2p3ozkcd3sjcsfi9wyyxbltjjze.demo.cms.va.gov/prescott-vet-center/

Screen Shot 2021-12-07 at 12 47 16 PM


kelsonic (approved):
lgtm!

PR #827, opened 2021-12-07
DNM: VACMS-000: alert Platform CMS team on content release failure

timcosgrove:
## Description
Currently content release failure alerts the entire #vfs-platform-builds channel; changing that to platform-cms-team.

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -600,7 +600,7 @@ jobs:
         uses: ./.github/workflows/slack-notify
         continue-on-error: true
         with:
-          attachments: '[{"mrkdwn_in": ["text"], "color": "danger", "text": "<!here> Content release for content-build has failed!: <https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}>"}]'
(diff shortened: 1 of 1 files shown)


theodur (approved):

.github/workflows/content-release.yml
```diff
@@ -600,7 +600,7 @@ jobs:
         uses: ./.github/workflows/slack-notify
         continue-on-error: true
         with:
-          attachments: '[{"mrkdwn_in": ["text"], "color": "danger", "text": "<!here> Content release for content-build has failed!: <https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}>"}]'
+          attachments: '[{"mrkdwn_in": ["text"], "color": "danger", "text": "<!platform-cms-team> Content release for content-build has failed!: <https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}>"}]'
```
theodur, line 603: I think to tag `@platform-cms-team` you'll need to use the format `<!subteam^S01M0LCA9FV|platform-cms-team>` instead for it to display properly.

`S01M0LCA9FV` is the usergroup ID

timcosgrove: I'm going to hold this until we have an agreed-upon alias for support for this. `platform-cms-team` would be better than `here` but it should probably be something like `cms-support-oncall` which we have to agree on and set up.

timcosgrove:
Withdrawing this; we are going to implement this as a PagerDuty alert. If there's work do be done in content-build for that, will open a separate request w/ a proper issue.

PR #845, opened 2021-12-13
Implement new CSAT for VAMC pages

gescobardev:
## Acceptance criteria
- [ ]

## Definition of done


src/site/includes/survey-tools.html
```diff
@@ -7,6 +9,11 @@
 {% endif %}
 
 {% if buildtype == 'vagovstaging' %}
+  {% if isVamcPage == true %}
+    <script type="text/javascript" nonce="**CSP_NONCE**">
```
gescobardev, line 13: Per Forsee tech support, adding this script tag triggers the new survey (vamc specific survey)

src/site/includes/survey-tools.html
```diff
@@ -1,3 +1,5 @@
+{% assign isVamcPage = entityUrl.path | pathContainsSubstring: 'health-care' %}
```
gescobardev, line 1: Per Michelle, the way we determine if a page is a VAMC page, is by checking if the base path contains "health-care"

src/site/filters/liquid.unit.spec.js
```diff
@@ -2014,3 +2014,32 @@ describe('isVisn8', () => {
...
+  });
+
+  it('returns true if path includes the search value - "vet-center"', () => {
+    const path = '/escanaba-vet-center/locations';
+    expect(liquid.filters.pathContainsSubstring(path, 'vet-center')).to.be.true;
+  });
+
+  it('returns false if no search value is passed', () => {
+    const path = '/escanaba-vet-center/locations';
+    expect(liquid.filters.pathContainsSubstring(path)).to.be.false;
```
kelsonic, line 2043: Interesting, makes sense! 🎉

src/site/includes/survey-tools.html
kelsonic: Should we do `window.vaSurvey` instead so it's more clear that it's being attached to the global window?

kelsonic (approved):

PR #832, opened 2021-12-08
Wait for the CMS to be ready before triggering govdelivery

cweagans:
## Description

In the past few days, there have been a number of content release failures related to trying to do stuff while the CMS is deploying.

Here are a couple of examples:

* https://github.com/department-of-veterans-affairs/content-build/actions/runs/1555528870
* https://github.com/department-of-veterans-affairs/content-build/actions/runs/1532069013

Instead of just sending the request, let's wait for the CMS to be ready and _then_ send the request. This needs some testing, which I will do in a separate PR so as not to pollute the Git history.

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #851, opened 2021-12-15
Wait for the CMS to be ready before running graphql queries

cweagans:
## Description

Follow-up to #832. There was a failure today resulting from the CMS being mid-deploy when the graphql queries were attempted. This PR should reduce the chances of that happening.

(There is still a potential failure mode where the CMS is ready before the graphql queries start, but the CMS goes down _during_ the graphql queries. There's not much that can be done about that at this point.)

Related ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7242

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -283,6 +283,9 @@ jobs:
         env:
           YARN_CACHE_FOLDER: .cache/yarn
 
+      - name: Wait for the CMS to be ready
+        uses: ./.github/workflows/wait-for-cms-ready
+
       - name: Export content build start time
         id: export-content-build-start-time
         run: echo ::set-output name=CONTENT_BUILD_START_TIME::$(date +"%s")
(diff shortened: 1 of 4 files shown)


timwright12:
@cweagans I'm not sure where going to get any milage out of this one, the GraphQL queries fail in the middle rather than the beginning. I think what you're looking for is the place in the build where the queries are actually made: https://github.com/department-of-veterans-affairs/content-build/tree/master/src/site/stages/build/drupal

cweagans:
>  the GraphQL queries fail in the middle rather than the beginning.

That's one failure mode, but the other is that the CMS is down before the graphql queries even start, as seen in this build from yesterday:  https://github.com/department-of-veterans-affairs/content-build/runs/4525879608?check_suite_focus=true#step:12:22

I've seen fails like this a couple of times, so I think it's worth doing an incremental improvement here to eliminate this specific scenario. If the next most frequent ping in #vfs-platform-builds is that the CMS goes down in the middle of the graphql queries, I'm happy to open a PR for that too.

> Also, seeing this CI failure:

Pushed a fix for this. Looks like the `working-directory` directive on the build steps doesn't apply to `uses`, so the full path needs to be specified.

ElijahLynn:
This looks good Cameron, +1 from me!

timwright12 (approved):

PR #897, opened 2022-01-10
add hook for good bad rating component

srsuddath:
## Description
This PR adds the hooks for us to add the new react good / bad ratings component on staging only

https://github.com/department-of-veterans-affairs/va.gov-team/issues/34626


src/site/layouts/checklist.drupal.liquid
```diff
@@ -88,7 +88,13 @@
           {% include "src/site/includes/tags.drupal.liquid" with fieldTags = fieldTags %}
 
           <!-- How do you rate -->
+        {% if buildtype == 'vagovprod' %}
```
kelsonic, line 91: One edge case with using `buildtype` is the preview server. The preview server technically runs on production, but there is a variable called `isPreview` to account for it (you can search for it globally and you should find instances of it being used for reference).

TSimmz (approved):

PR #906, opened 2022-01-13
Fix/security update markdown it

timwright12:
## Description
Security alert in vets-website for this dep. Just fixing here as well.


rileyanderson (approved):

alexandec:
Noticed lots of commits in here, not sure where those came from. Just wanted to mention it

timwright12:
@alexandec yeah I saw that too, not sure, but it seems like the changed files are correct

PR #930, opened 2022-01-24
VACMS-7275: Add logic to remove media paragraph html wrapper if there is no image

netboss:
- Added logic to hide media paragraph html wrapper. Pushed issue to debug in tugboat after content-build front-end.

## Acceptance criteria
- [ ]

## Definition of done


src/site/paragraphs/media.drupal.liquid
```diff
@@ -30,6 +30,8 @@
         <i class="fas fa-expand-arrows-alt"></i>
       </a>
     {% endif %}
-    <img src="{{ entity.fieldMedia.entity.image.url }}" alt="{{ entity.fieldMedia.entity.image.alt }}"/>
+    {% if entity.fieldMedia.entity.image.url != empty or entity.fieldMedia.entity.image.url.length %}
```
swirtSJW, line 33: Isn't the !empty check enough.  When would it have length and not be empty?

swirtSJW:
There are too many 'or's in this step.  
` If there is an existing image either remove or change it, or add a new image and publish, and save the node`  
A testing step should not really have "or" or you will end up with two testers testing things differently.
If there are multiple things that need variations, there should be separate steps.

We are trying to test 2 things in this PR .  When there is not image url, we we should not see an image tag.   When there is an image, we should see one.

src/site/paragraphs/media.drupal.liquid
netboss: @swirtSJW 
Yeah, I wasn't sure about the exact syntax in .liquid files. I'll remove one of them.

netboss:
> There are too many 'or's in this step. ` If there is an existing image either remove or change it, or add a new image and publish, and save the node` A testing step should not really have "or" or you will end up with two testers testing things differently. If there are multiple things that need variations, there should be separate steps.
> 
> We are trying to test 2 things in this PR . When there is not image url, we we should not see an image tag. When there is an image, we should see one.

@swirtSJW 
Updated the QA steps.

netboss:
@swirtSJW 
Removed redundant code and re-tested in personal tugboat instance and all worked as expected.

swirtSJW (changes requested):
@netboss I feel like this does not follow the issue title 
image

The issue is to remove the entire media parargraph  not just the image tag.

netboss:
> @netboss I feel like this does not follow the issue title image
> 
> The issue is to remove the entire media parargraph not just the image tag.

@swirtSJW 
I see what you're saying, but, to my understanding, the paragraph shouldn't be added to the node if there is no need to embed an image, so the way it's right now "should" work. The only thing I can see is an scenario where an editor adds the media image embed paragraph and then decides to remove the image but leave the paragraph in the node, which seems odd, but can/might happen.

src/site/paragraphs/media.drupal.liquid
```diff
@@ -18,18 +18,20 @@
...
+  <div data-template="paragraphs/media" data-entity-id="{{ entity.entityId }}" class="vads-u-display--block">
+    <div class="va-c-position--relative vads-u-display--inline-block vads-u-margin-y--1p5">
+      {% if entity.fieldAllowClicksOnThisImage %}
+        <a
+          aria-label="Open image in new tab"
+          class="vads-u-margin-right--1p5 vads-u-margin-top--1p5 vads-u-text-decoration--none vads-u-display--flex vads-u-align-items--center expand-image-button va-c-position--absolute va-c-position-top-right-corner vads-u-justify-content--center"
+          href="{{ entity.fieldMedia.entity.image.url }}"
+          target="_blank"
+        >
+          <i class="fas fa-expand-arrows-alt"></i>
```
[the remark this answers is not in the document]
jhonnyoddball: Make sure you are using proper a11y in the `<i>` tag

netboss:
> @netboss I feel like this does not follow the issue title image
> 
> The issue is to remove the entire media parargraph not just the image tag.

@swirtSJW 
Moved conditional to wrap the paragraph div instead of just the image tag. Ran tests with and without image and all worked fine. This is ready for review/QA.

swirtSJW (approved):
Nice work @netboss

jhonnyoddball (approved):

PR #931, opened 2022-01-24
Use latest `vets-website` production build in daily accessibility scan

ddzz:
## Description
PR to modify the daily accessibility scan to use the latest `master` production build of `vets-website` instead of pulling assets from S3.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/35837

## Acceptance criteria
- [ ] CI passes.

### DIFF
--- a/.github/workflows/a11y.yml
+++ b/.github/workflows/a11y.yml
@@ -161,7 +161,7 @@ jobs:
 
           mkdir -p build/vagovprod/generated
           ls
-          cp -a /vatest/. /build/vagovprod/generated
+          cp -a vatest/. build/vagovprod/generated
           cd build/vagovprod/generated
           ls
 
--- a/.github/workflows/a11y.yml
+++ b/.github/workflows/a11y.yml
@@ -1,9 +1,9 @@
 name: Accessibility Tests
 
-on:
-  workflow_dispatch:
-  schedule:
-    - cron: '0 14 * * 1-5'
+on: [push]
+  # workflow_dispatch:
+  # schedule:
+  #   - cron: '0 14 * * 1-5'
 
 env:
   CHROMEDRIVER_FILEPATH: /usr/local/share/chrome_driver/chromedriver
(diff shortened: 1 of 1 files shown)


ddzz:
> Just noting the title of the PR, but if we're doing this I think we should use staging since dev has a slightly different build process than production

I just updated the title to reflect the latest changes. This PR no longer uses the dev deploy and instead uses the latest production build. I meant to update the title when the PR was ready for review.

timwright12 (approved):

holdenhinkle (approved):

.github/workflows/a11y.yml
```diff
@@ -67,6 +73,32 @@ jobs:
...
+        id: changed-files
+        working-directory: vets-website
+        run: echo ::set-output name=all_changed_files::$(git diff --name-only origin/master...)
+
+      - name: Get app entries for changed files
+        id: get-changed-apps
+        working-directory: vets-website
+        run: echo ::set-output name=app_entries::$(node script/github-actions/get-changed-apps.js)
+        env:
+          CHANGED_FILE_PATHS: ${{ steps.changed-files.outputs.all_changed_files }}
```
theodur, line 93: Just saw this after it got merged. This will run a single app build if the changed files in the latest commit are all from an app on the allowlist. I think you'll need to omit these two steps and just run a regular full build in the `Build vets-website` step to always build full builds. i.e. (`yarn build --verbose --buildtype=vagovprod` without the `${ENTRY:+"--entry=$ENTRY"}` part).

.github/workflows/a11y.yml
```diff
@@ -87,11 +119,16 @@ jobs:
...
-      - name: Build
+      - name: Build content-build
         run: yarn build --buildtype=vagovprod --asset-source=local --drupal-address=https://cms-content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov --pull-drupal --drupal-max-parallel-requests=15 --no-drupal-proxy --verbose
         env:
           NODE_ENV: production
 
+      - name: Copy vets-website assets to content-build
+        run: |
+          cd ..
+          cp -arf vets-website/build/vagovprod/* content-build/build/vagovprod
```
theodur, line 130: I'm not sure that just moving the assets into the `content-build` directory here will make the content build use them. I think it [always downloads](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/stages/build/plugins/download-assets.js#L70-L73) the assets from S3 for non-localhost builds, which would overwrite the assets that were just copied.

.github/workflows/a11y.yml
ddzz: @theodur This should be behaving the same way as `vets-website` CI on the `master` branch. Doesn't that run the full build?

theodur: @ddzz  We have an [if statement](https://github.com/department-of-veterans-affairs/vets-website/blob/master/.github/workflows/continuous-integration.yml#L52) in the `Get app entries for changed files` job that prevents it from running on `master`. I'm not seeing that here, so this may run single app builds here. 

Since you always want a full `vets-website` build for this workflow anyway, you don't need the extra steps for single app builds.

ddzz: Ah, I missed that. I'll open a PR. Thanks for reviewing.

PR #948, opened 2022-02-02
VAGOV-0000: always record metrics once build has succeeded.

timcosgrove:
## Description
Most of this workflow is required to complete if build has completed, but the last job does not.


timcosgrove:
See #947

PR #976, opened 2022-02-17
36786 events v2

tmac9494:
## Description
 - Cut down the Timezone on event page to 2 letters instead of 3.
 - added forward arrow icon to see more events button on event page.

## Testing done
Tested locally with no issues

https://github.com/department-of-veterans-affairs/va.gov-team/issues/36786-


rmessina1010:
It looks like the method used to cut down the timezone abbreviation is to replace the 'S' with '';  This  will fail for Daylight Savings ('D'). I suggest using REGEX  to swap ([SD]T) for (T) or at least chaining to remove the possible 'D' as well, before proceeding to merge.

src/site/layouts/event.drupal.liquid
```diff
@@ -359,6 +359,7 @@
             onclick="recordEvent({ event: 'nav-secondary-button-click' });"
           >
             See more events
+            <i aria-hidden="true" style="font-size:12px" class='vads-u-margin-left--1 fa fa-chevron-right' role="presentation"></i>
```
timwright12, line 362: @tmac9494 can we use a design system class or add some new styles instead of inline CSS? It complicates the specificity

tmac9494:
added a quick fix, .replace was in the wrong location in liquid.js

rmessina1010:
`@github-actionsgithub-actions
/ Unit Tests Summary
/home/runner/work/content-build/content-build/src/site/filters/liquid.unit.spec.js.deriveFormattedTimestamp returns what we expect
expected 'Wed. Jan. 5, 2022, 1:00 p.m. – 2:00 p.m. ET' to equal 'Wed. Jan. 5, 2022, 1:00 p.m. – 2:00 p.m. EST'`
Un-updated test??

src/site/layouts/event.drupal.liquid
tmac9494: fixed

longmd:
As @rmessina1010 mentioned above, I would regex the replace to account for DST. This is my suggestion for your replace argument on the TZ. `.replace(/S|D/i, '')`. Outside of that, all looks good to me.

longmd (approved):

TSimmz (approved):
LGTM 👍

PR #983, opened 2022-02-22
Make runner teardown dependent on record-metrics being finished.

timcosgrove:
## Description

See https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8107

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -662,37 +662,6 @@ jobs:
           -H "DD-API-KEY: ${{ env.GHA_CONTENT_BUILD_DATADOG_API_KEY }}" \
           -d @- < event.json
 
-  stop-runner:
-    name: Stop on-demand-runner
-    needs: [archive, start-runner, create-release, deploy]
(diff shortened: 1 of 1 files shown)


.github/workflows/content-release.yml
```diff
@@ -779,3 +748,39 @@ jobs:
...
           -d @- < metrics.json
+
+  stop-runner:
+    name: Stop on-demand-runner
+    needs:
+      - archive
+      - start-runner
+      - create-release
+      - deploy
+      - record-metrics
```
timcosgrove, line 759: Add `record-metrics` as a dependency

.github/workflows/content-release.yml
```diff
@@ -779,3 +748,39 @@ jobs:
...
+    runs-on: ubuntu-latest
```
timcosgrove, line 760: use GHA runner to avoid the edge case where GHA selects the labeled self-hosted runner to tear itself down.

cweagans (approved):

PR #994, opened 2022-02-24
Move Content-Release Drupal Connection Vars to Build Step

olivereri:
## Description
Secrets from SSM might be scrubbed in between jobs of a GHA workflow. Move the secrets fetch to right before the build instead of in a dedicated "Get env vars" job.

https://github.com/marvinpinto/actions/blob/master/packages/aws-ssm-secrets/src/main.ts#L49

## Acceptance criteria
- [ ]

## Definition of done


timcosgrove (approved):
r+

cweagans:
I don't think this will fix your problem. I don't think the content build AWS user has access to that path. If you put stuff in `/dsva-vagov/content-build` instead, I think it should work?

PR #1011, opened 2022-03-01
Change recurring sort to grab first instance of date of event listing pages

ethanteague:
## Description
When event listings have events with multiple recurrences, we want the soonest one to appear.
closes #[37833](https://github.com/department-of-veterans-affairs/va.gov-team/issues/37833)

## Testing done
Visual

## Screenshots
Yes:
image

No:
image

## Acceptance criteria
- [ ] Recurring events created for Pittsburgh in https://pr7981-5kfpv6vznchkcxw4bgycflzjjodfbtbe.ci.cms.va.gov/ environment show soonest occurrence on listing page https://pr7981-5kfpv6vznchkcxw4bgycflzjjodfbtbe.ci.cms.va.gov/pittsburgh-health-care
- [ ] Event listing past event pages do not show upcoming events
- [ ] Featured events appear when appropriate on listing pages
- [ ] Regressions are not introduced on VAMC pages that have featured events
- [ ] Single instance (non-recurring) events still appear on listing pages when expected

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -1365,6 +1365,17 @@ module.exports = function registerFilters() {
     return futureDates[0];
   };
 
+  liquid.filters.deriveMostRecentDateOnEventsListingPage = fieldDatetimeRangeTimezone => {
+    const dates = [];
+    fieldDatetimeRangeTimezone.forEach(element => {
+      if (element?.endValue > moment().unix()) {
+        dates.push(element);
+      }
+    });
+    const sortedDates = _.sortBy(dates, 'endValue');
+    return sortedDates[0];
+  };
+
   // Given an array of services provided at a facility,
   // return a flattened array of service locations that
   // offer service of type `serviceType`
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -895,7 +895,7 @@ module.exports = function registerFilters() {
     if (!data) return null;
     const currentTimestamp = new Date().getTime();
     return data.filter(event => {
-      const mostRecentEvent = liquid.filters.deriveMostRecentDate(
+      const mostRecentEvent = liquid.filters.deriveMostRecentEventDate(
         event.fieldDatetimeRangeTimezone,
       );
       return mostRecentEvent.value * 1000 >= currentTimestamp;
@@ -1335,6 +1335,34 @@ module.exports = function registerFilters() {
     return basePath.includes(searchValue);
   };
 
+  liquid.filters.deriveMostRecentEventDate = fieldDatetimeRangeTimezone => {
+    const now = moment().unix();
+    if (!fieldDatetimeRangeTimezone) return fieldDatetimeRangeTimezone;
+
+    // Return back fieldDatetimeRangeTimezone if it is already a singular most recent date.
+    if (!_.isArray(fieldDatetimeRangeTimezone)) {
+      return fieldDatetimeRangeTimezone;
+    }
+
+    // Return back fieldDatetimeRangeTimezone's first item if it only has 1 item.
+    if (fieldDatetimeRangeTimezone?.length === 1) {
+      return fieldDatetimeRangeTimezone[0];
+    }
+    // Derive date times relative to now.
+    const dates = [];
+    fieldDatetimeRangeTimezone.forEach(element => {
+      if (element?.endValue > now) {
+        dates.push(element);
+      }
+    });
+    // Return the soonest upcoming date.
+    if (dates) {
+      return dates[0];
+    }
+
+    return [];
+  };
+
   liquid.filters.deriveMostRecentDate = (
     fieldDatetimeRangeTimezone,
     now = moment().unix(), // This is done so that we can mock the current time in tests.
@@ -1363,17 +1391,6 @@ module.exports = function registerFilters() {
 
     // Return the most recent future date if there are future dates.
     return futureDates[0];
-  };
-
-  liquid.filters.deriveMostRecentDateOnEventsListingPage = fieldDatetimeRangeTimezone => {
-    const dates = [];
-    fieldDatetimeRangeTimezone.forEach(element => {
-      if (element?.endValue > moment().unix()) {
-        dates.push(element);
-      }
-    });
-    const sortedDates = _.sortBy(dates, 'endValue');
-    return sortedDates[0];
   };
 
   // Given an array of services provided at a facility,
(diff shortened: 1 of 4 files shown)


GcioNapoleon (approved):

src/site/filters/liquid.js
```diff
@@ -1358,7 +1358,7 @@ module.exports = function registerFilters() {
 
     // Return the most recent past date if there are no future dates.
     if (_.isEmpty(futureDates)) {
-      return dates[dates?.length - 1];
+      return dates[0];
```
swirtSJW, line 1361: This seems like a more dramatic change to this filter and makes the comment one line up a bit of a lie.   If there are no future dates, doesn't this grab the very first date in the the series  (essentially grabbing the oldest expired) rather than the most recent expired?

swirtSJW: This does not seem to be covering case in the issue where it need to grab the next upcoming.  For future dates.  This line is for if there are no future dates.

ethanteague: Think I got it now, @swirtSJW

kevwalsh (changes requested):
Not sure if the [Tugboat environment](https://web-5kfpv6vznchkcxw4bgycflzjjodfbtbe.ci.cms.va.gov/pittsburgh-health-care/events/) has the latest code, but it seems like we have the opposite issue now. Clicking the representative event in the listing brings me to a page that shows the last event in the series. 

Events___VA_Pittsburgh_Health_Care___Veterans_Affairs_and_VA_Pittsburgh_Health_Care___Recurring_Monthly___Veterans_Affairs

src/site/filters/liquid.js
```diff
@@ -1365,6 +1365,17 @@ module.exports = function registerFilters() {
     return futureDates[0];
   };
 
+  liquid.filters.deriveMostRecentDateOnEventsListingPage = fieldDatetimeRangeTimezone => {
```
swirtSJW, line 1368: I'm not clear on why we need two different filters to do the same thing.  The data should be the same in either case.  What's different about the use cases?

ethanteague: I just couldn't make the other filter work for our use case on listing pages after messing with it for days. Gave up and tried a new filter without all of the conditionals, and was able to make it fire, @swirtSJW

ethanteague: Alright. Think I got it, @swirtSJW

src/site/filters/liquid.unit.spec.js
```diff
@@ -2096,7 +2096,7 @@ describe('deriveMostRecentDate', () => {
     // Assertions.
     expect(
       liquid.filters.deriveMostRecentDate(fieldDatetimeRangeTimezone, now),
-    ).to.deep.eq({ value: 1642025600, endValue: 1642029600 });
+    ).to.deep.eq({ value: 1642014000, endValue: 1642017600 });
```
swirtSJW, line 2099: Why did all the tests need new times?

ethanteague: The tests were predicated on the last occurrence in a recurring series being the item presented.

src/site/filters/liquid.js
```diff
@@ -1352,17 +1352,25 @@ module.exports = function registerFilters() {
       return fieldDatetimeRangeTimezone[0];
     }
 
-    // Derive date times relative to now.
-    const dates = _.sortBy(fieldDatetimeRangeTimezone, 'endValue');
-    const futureDates = _.filter(dates, date => date?.endValue - now > 0);
+    const sortedDates = _.sortBy(
+      fieldDatetimeRangeTimezone,
+      'endValue',
+    ).reverse();
```
swirtSJW, line 1358: This reverse is unexpected (counter intuitive).   Reversing the order seems odd.  Why do we want the dates array built from most distant to closest?

ethanteague: lodash _sortBy only allows for descending order (but it's a really quick / easy / one method way to sort an array) - seems quicker to do it this way then reverse it than with a typical three step array sort approach.

swirtSJW: That doesn't seem right.   
image
It is ascending by the value of a function return.  "endValue"  is neither a function return or a value. 

https://lodash.com/docs/4.17.15#sortBy

ethanteague: Sorry - misspoke when I said descending (guess it's ascending). Not sure I'm following what you are after, though. Are you saying I should just be using a compare type of function?

swirtSJW: Yes sortBy seems to need a comparing function, not just an array key.   I think you either need the comparing function or you need to use orderBy

src/site/filters/liquid.js
```diff
@@ -895,7 +895,7 @@ module.exports = function registerFilters() {
     if (!data) return null;
     const currentTimestamp = new Date().getTime();
     return data.filter(event => {
-      const mostRecentEvent = liquid.filters.deriveMostRecentDate(
+      const mostRecentEvent = liquid.filters.deriveMostRecentEventDate(
```
swirtSJW, line 898: I'm a bit concerned about creating a second filter that works and leaving an old one behind that does not work.    Is there a reason we can't  re-use the same name of the filter?

src/site/filters/liquid.js
```diff
@@ -1335,6 +1335,34 @@ module.exports = function registerFilters() {
...
+      if (element?.endValue > now) {
+        dates.push(element);
+      }
+    });
+    // Return the soonest upcoming date.
+    if (dates) {
+      return dates[0];
+    }
+
+    return [];
```
swirtSJW, line 1363: If there are no future dates, this returns an empty array.   Kelson's version of this had a fallback to grab the last past date in that event.   Don't we still need that fallback?

src/site/filters/liquid.js
ethanteague: Yeah, I feel that, @swirtSJW . I'm kinda nervous about doing that, because this filter is also used for clp, past events and situation updates. I just don't know what else might get broken.

src/site/filters/liquid.js
ethanteague: I tested all of the places events show up, and doesn't appear that we do - looks like past events have unique handling, @swirtSJW .

ethanteague:
Closing in favor of https://github.com/department-of-veterans-affairs/content-build/pull/1025

PR #1025, opened 2022-03-08
VACMS-37833: fixes incorrect date showing for recurring events.

ryguyk:
## Description
Recurring events date is showing last date (largest date value) rather than the nearest date in the future.  This PR fixes that.

Note: the problem is that a phantom object is being passed to the filter `deriveMostRecentDate`, and so the default parameter `now` is not actually setting to the default.  This ultimately leads to an empty `futureDates` array, which leaves the logic to just grab the last item in the original data set (rather than the first item of the now-empty `futureDates` data set).

It's unclear why this phantom parameter is being passed, but this PR implements a workaround by explicitly passing the current time as the second parameter.

## Testing done
See screenshot below.  Screenshot taken March 8.  Notice all dates are conceivably the _next_ date in the series.

## Screenshots
image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/site/components/situation_updates.drupal.liquid
+++ b/src/site/components/situation_updates.drupal.liquid
@@ -15,7 +15,8 @@
           <div class="vads-u-margin-bottom--0 no-p-bottom-margin">
             <h4 class="vads-u-margin-top--1 vads-u-margin-bottom--2">
               <!-- Derive most recent date -->
-              {% assign mostRecentDate = update.entity.fieldDatetimeRangeTimezone | deriveMostRecentDate %}
+              {% assign now = '' | currentTime %}
+              {% assign mostRecentDate = update.entity.fieldDatetimeRangeTimezone | deriveMostRecentDate: now %}
 
               {% if mostRecentDate.value != empty %}
                 {{ mostRecentDate.value | dateFromUnix: "dddd, MMM D, h:mm A"}}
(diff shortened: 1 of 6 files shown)


ethanteague (approved):
@ryguyk @swirtSJW @kevwalsh I think this is good enough for now.

swirtSJW:
Brining in AC QA from 1011

- [ ] Recurring events created for Pittsburgh in https://pr7981-5kfpv6vznchkcxw4bgycflzjjodfbtbe.ci.cms.va.gov/ environment show soonest occurrence on listing page /pittsburgh-health-care
- [ ]  Event listing past event pages do not show upcoming events
- [ ]  Featured events appear when appropriate on listing pages
- [ ]  Regressions are not introduced on VAMC pages that have featured events
- [ ]  Single instance (non-recurring) events still appear on listing pages when expected

swirtSJW:
We have agreement on most upcoming for facility events.
image

swirtSJW:
We have agreement on facility featured events.
image

swirtSJW:
We have a defect on time.... this might be an existing bug, still looking in to it 

image

swirtSJW:
Defect on time zone handling. 
image

swirtSJW:
Situation updates test out good.
image

tmac9494 (approved):

PR #1043, opened 2022-03-22
[DNM] Decouple daily prod deploy and content release

cweagans:
## Description

This is not a small PR and it is not particularly easy to parse, review, or test. Here be dragons. Please feel free to ping me on Slack to talk through anything you see here that is not clear. There is still a little bit more work to do to wrap this up, but I'm opening this early so everyone can start taking a look.

Here's what's in this PR:
* NEW
	* Workflow to send 1 hour warning for daily production deploy
	* A human-friendly tag name is now included in `BUILD.txt`
	* A notification is now sent to #status-content-build after every successful content release. This notification specifies the tag that was used to create the release for a visual demarcation between releases using code from yesterday and releases using code from today.
* CHANGED
	* Daily production deploy no longer runs a content release and instead just creates a tag at the scheduled time.
	* Content release had a ton of conditionals for non-prod environments even though it only supports prod. I removed all of these to make the workflow more understandable. From here, there are some small tweaks that we can make to enable us to test the content release workflow in PRs (!)
	* Inline arrays (for instance, in `runs-on` blocks) have been expanded to multiline arrays for better diffs in the future
	* The message emitted when a content release starts now specifies the content-build tag that will be used.
	* Broken link checks now run before the prearchive script so that the workflow can fail faster if there is a problem. This removes approximately 1.5 minutes from our cycle time if we're debugging broken links issues and doesn't impact what actually gets deployed.
* REMOVED
	*  The set-env job no longer exists in the content release workflow. All environment variables have been set in the workflow-wide `env` block.
	* Each job in the content-release workflow was setting `NODE_EXTRA_CA_CERTS`. This has been removed in favor of setting that variable globally for the entire workflow.
	* `wait-for-current-workflow-to-complete` no longer exists (we don't need it anymore)
	* Individual content releases no longer create tags nor releases. I believe this was a mistake from when we switched to Github Actions and the result of doing this was that we were deploying changes in `master` much more frequently than we intended to. After this PR, _not_ creating releases on every content build is important to ensure that we're rolling out changes only when we specifically intend to.
	* Both content release params that were previously used in `workflow_dispatch` have been removed (we no longer need to specify the environment and we didn't use the release wait param in practice)
	* Many jobs had `needs: [x, y, z]` and `if: always() && needs.x.result == 'success' && needs.y.result == 'success' && needs.z.result == 'success'`. This has been simplified to just `needs: [x, y, z]` per the [github docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

## Acceptance criteria
- [ ]

## Definition of done


.github/workflows/content-release.yml
```diff
@@ -2,41 +2,43 @@ name: Content Release
...
-      release_wait:
-        description: Minutes to wait before creating release
-        required: false
-        default: 0
-      deploy_environment:
-        type: choice
-        description: The environment to deploy content to
-        required: true
-        options:
-          - prod
```
ElijahLynn, line 15: I think it does make sense to remove this until dev and staging are actually supported. It can be added again when that functionality ships.

.github/workflows/content-release.yml
```diff
@@ -2,41 +2,43 @@ name: Content Release
...
 jobs:
   start-runner:
-    runs-on: [self-hosted, linux, x64]
+    runs-on:
+      - self-hosted
+      - linux
+      - x64
     outputs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
       ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
```
ElijahLynn, line 32: Not a change you made but worth noting that I don't see where this runner is declared.

cweagans: We have a pool of self hosted linux x64 runners that are always on and attached to the Github org. All of the jobs that don't specify a specific runner (using the label) can be run on any of our self-hosted runners.

.github/workflows/content-release.yml
cweagans: 👍🏻  agreed. I'm not sure that this job ever needs to support dev and staging, but we can always add it back later if we decide there's a need for it.

.github/workflows/content-release.yml
```diff
@@ -329,10 +247,8 @@ jobs:
           env_variable_name: DRUPAL_USERNAME
 
       - name: Build
-        run: yarn build --buildtype=${{ needs.set-env.outputs.BUILDTYPE }} --asset-source=local --drupal-address=${{ needs.set-env.outputs.DRUPAL_ADDRESS }} --drupal-user=${{ env.DRUPAL_USERNAME }} --drupal-password="${{ env.DRUPAL_PASSWORD }}" --pull-drupal --drupal-max-parallel-requests=15 --no-drupal-proxy --verbose | tee build-output.txt
+        run: yarn build --buildtype=${{ env.BUILDTYPE }} --asset-source=local --drupal-address=${{ env.DRUPAL_ADDRESS }} --drupal-user=${{ env.DRUPAL_USERNAME }} --drupal-password="${{ env.DRUPAL_PASSWORD }}" --pull-drupal --drupal-max-parallel-requests=15 --no-drupal-proxy --verbose | tee build-output.txt
```
timcosgrove, line 250: Confirming `DRUPAL_USERNAME` and `DRUPAL_PASSWORD` are set here.

.github/workflows/content-release.yml
```diff
@@ -641,15 +484,38 @@ jobs:
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Notify Slack
+        uses: ./.github/workflows/slack-notify
+        continue-on-error: true
+        with:
+          payload: '{"attachments": [{"color": "#2EB67D","blocks": [{"type": "section","text": {"type": "mrkdwn","text": "content release using ${{ needs.validate-build-status.outputs.TAG }} is complete."}}]}]}'
```
timcosgrove, line 505: Bonus points for threading this with original notification, and/or using emoji reactions to indicate in-progress & done (just thinking out loud here).

timcosgrove (commented):
comments/questions on content-release

.github/workflows/content-release.yml
```diff
@@ -2,41 +2,43 @@ name: Content Release
...
   start-runner:
-    runs-on: [self-hosted, linux, x64]
+    runs-on:
+      - self-hosted
+      - linux
+      - x64
     outputs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
       ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+      APPROX_WORKFLOW_START_TIME: ${{ steps.export-approx-workflow-start-time.outputs.APPROX_WORKFLOW_START_TIME }}
```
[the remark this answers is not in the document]
cweagans: It's approximate because of our earlier issues: the workflow start time is not necessarily the same time as the start time of the first job (e.g. if the job has to wait for a runner to become available).

.github/workflows/content-release.yml
```diff
@@ -2,41 +2,43 @@ name: Content Release
...
 concurrency:
-  group: content-release-${{ github.event_name == 'workflow_dispatch' && github.event.inputs.deploy_environment || 'prod' }}
-  cancel-in-progress: ${{ github.event.inputs.deploy_environment != 'prod' }}
+  group: content-release-prod
 
 env:
   CMS_NOTIFICATIONS_SLACK: CJT90C0UT # cms-notifications
   VFS_PLATFORM_BUILDS_SLACK: C0MQ281DJ # vfs-platform-builds
   BROKEN_LINKS_SLACK: C030F5WV2TF # content-broken-links
+  CONTENT_BUILD_CHANNEL_ID: C02VD909V08 #status-content-build
```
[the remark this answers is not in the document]
cweagans: More or less. `set-env` was important when we were trying to make this workflow work for multiple environments, but if we just lean into using this workflow for prod (as we have been for the past six months), we don't need it anymore.

timcosgrove (approved):
r+ Provided RTT also looks at this and approves.

PR #1056, opened 2022-03-29
Housekeeping for job decoupling

cweagans:
## Description

* Content release should run every hour 13:00-21:00 UTC
* Daily prod deploy is now the daily production release (since it's not actually deploying anything)
* Daily deploy warning is now the daily release warning
* Verbiage changes for the above
* Include release name in slack notification for daily prod release workflow

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/daily-release-warning.yml
+++ b/.github/workflows/daily-release-warning.yml
@@ -1,11 +1,11 @@
-name: Daily Deploy Warning
+name: Daily Release Warning
 
 on:
   schedule:
     - cron: 0 17 * * 1-5
 
 concurrency:
-  group: daily-deploy-warning
+  group: daily-release-warning
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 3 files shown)


alexandec (approved):

.github/workflows/daily-release-warning.yml
```diff
@@ -1,11 +1,11 @@
-name: Daily Deploy Warning
+name: Daily Release Warning
```
theodur, line 1: Would it make sense to move the functionality of this workflow into the Daily Production Release workflow? It seems like this workflow is just sending out an alert for what's going to happen in the Daily Production Release workflow

theodur: That makes sense about not holding up a runner. The only downside is we lose the functionality that we previously had of being able to manually set a waiting period when manually deploying (or in this case, creating a release). Not sure how useful that was to people though

theodur: And I guess if we ever wanted to change the waiting time, we'd need to update both workflows so that `daily-production-release`'s cron is always in sync with the release warning message (we've never changed it though so it's unlikely we'll ever need to). Moving to autonomous deployments entirely like you mentioned would remove the need for any of this, but that's probably a larger decision to be made 🙂

cweagans: > The only downside is we lose the functionality that we previously had of being able to manually set a waiting period when manually deploying (or in this case, creating a release). Not sure how useful that was to people though

Not sure anyone _really_ relied on that, but I could be wrong.

>  Moving to autonomous deployments entirely like you mentioned would remove the need for any of this, but that's probably a larger decision to be made 🙂

Probably so, but I just want to be super clear: we were doing unattended production deployments once per hour or more for the last six months. Content release _was_ creating new Github releases every time it ran. It was accidental, but it worked well enough that nobody noticed until now.

theodur: > Probably so, but I just want to be super clear: we were doing unattended production deployments once per hour or more for the last six months. Content release was creating new Github releases every time it ran. It was accidental, but it worked well enough that nobody noticed until now.

Content releases were _never_ deploying code changes to production, despite making a new release. They were only making new releases using the commit deployed in the last daily production deploy. So nothing new was going to prod other than CMS content updates. This was the functionality even before GitHub Actions, when we were using Jenkins. Either way, it makes sense that we're only making one release a day now

PR #1078, opened 2022-04-13
Update manual dev/staging deploy

theodur:
## Description
This PR makes some enhancements to the manual dev/staging deploy:
- Uses the `self-hosted` runner for the `Deploy` job (the deploy was timing out due to lack of resources)
- Defines and uses a matrix for the `Deploy` job, so both environments can be deployed simultaneously

## Testing done
Tested using changes in branch:
_Deploy script was commented out for testing purposes_

- [dev deploy](https://github.com/department-of-veterans-affairs/content-build/actions/runs/2158500589)
- [stging deploy](https://github.com/department-of-veterans-affairs/content-build/actions/runs/2158496736)
- [both deployed](https://github.com/department-of-veterans-affairs/content-build/actions/runs/2158504214)

## Acceptance criteria
- [x] Manual dev/staging deploy uses the `self-hosted` runner
- [x] Both dev and staging can be deployed simultaneously

## Definition of done

### DIFF
--- a/.github/workflows/manual-deploy-dev-staging.yml
+++ b/.github/workflows/manual-deploy-dev-staging.yml
@@ -25,6 +25,7 @@ jobs:
     runs-on: self-hosted
     env:
       NODE_EXTRA_CA_CERTS: /etc/ssl/certs/VA-Internal-S2-RCA1-v1.cer.pem
+      RUN_DEPLOY: ${{ github.event.inputs.deploy_environment == matrix.environment || github.event.inputs.deploy_environment == 'both' }}
 
     strategy:
       matrix:
@@ -36,9 +37,11 @@ jobs:
 
     steps:
       - name: Checkout
+        if: env.RUN_DEPLOY
         uses: actions/checkout@v2
 
       - name: Configure AWS credentials (1)
+        if: env.RUN_DEPLOY
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-deploy-dev-staging.yml
```diff
@@ -20,9 +20,37 @@ env:
...
   deploy:
     name: Deploy
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
+    needs: set-environment
+    strategy:
+      matrix: ${{ fromJson(needs.set-environment.outputs.environment) }}
+
+    env:
+      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/VA-Internal-S2-RCA1-v1.cer.pem
```
alexandec, line 53: Glad to see this, it will prevent the missing cert issue when switching to self-hosted

.github/workflows/manual-deploy-dev-staging.yml
```diff
@@ -20,9 +20,37 @@ env:
...
 
 jobs:
+  set-environment:
+    name: Set environment to deploy
+    runs-on: ubuntu-latest
+    outputs:
+      environment: ${{ steps.set-output.outputs.environment }}
+
+    env:
+      dev: "{\\\"environment\\\": \\\"vagovdev\\\", \\\"bucket\\\": \\\"content.dev.va.gov\\\"}"
```
alexandec, line 30: This approach is definitely cleaner overall 👍 Just wish there was a way to avoid all the escaping

alexandec:
One other thing that occurred to me, I remember seeing something about GHA's own runners eventually becoming configurable to provide more resources. I don't know the status on that though

alexandec (approved):

.github/workflows/manual-deploy-dev-staging.yml
theodur: I know right haha those escapes are hideous 😅 I wish GHA had a built-in way of handling them in strings. The third backslash is needed because of the string being used in the `::set-output` call. GHA strips the escape character when referencing the environment variable, which causes issues when using `fromJson`

theodur:
> One other thing that occurred to me, I remember seeing something about GHA's own runners eventually becoming configurable to provide more resources. I don't know the status on that though

That would be nice to have the ability to configure the runners to use the resources you need, kind of like in Codespaces. I hope that gets added soon, I couldn't find anything about it in the [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) 🙁 

The `ubuntu-latest` runner could  probably finish deploys if it had more time. I think it was timing out because of the [time limit](https://github.com/department-of-veterans-affairs/content-build/blob/master/.github/workflows/manual-deploy-dev-staging.yml#L51) we set on the role. It's still much quicker on the self-hosted runner though, which is probably why it's being used for the deploys in the CI workflow

PR #1089, opened 2022-04-21
VACMS-8361 Homepage iteration - Search and Common Tasks

allisonlu:
## Description

Build out Search Common Tasks for Homepage iteration: [#8361](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8361)

Merging into a feature branch for usability testing.

## Acceptance criteria
- [ ] is responsive

## Definition of done


src/site/includes/common-tasks.html
```diff
@@ -0,0 +1,57 @@
+<!--
```
ethanteague, line 1: Is there a reason to not use the .liquid extension on this file? I think our setup allows for liquid curlies and logic inside of files that don't have said extension, but wondering if the omission of .liquid is intentional, @allisonlu ?

ryguyk (approved):
This looks good. There was a question that came up around why we were implementing this without the search functionality. I can't recall for sure what our assumptions were, but it might have been that we were not sure there was a reusable component that matched the design. Is that correct?

allisonlu:
@ryguyk Good question. I believe we still need to gather analytics to figure out what our Top Tasks are. See [kickoff notes](https://docs.google.com/document/d/1e76MuHujBRy_tKrZkTl5zrVVv7v7ZfQ8_SIuaCVgjSc/edit#heading=h.ojne53j77zm9). Maybe @mmiddaugh will have better assumptions on the Search functionality?

cc @jilladams @wesrowe

mmiddaugh:
@ryguyk @allisonlu 
We're gearing up for some Veteran research on Top Tasks to inform that short list. 
I understood the plan was to have search behave like current site search. I'm not sure I understand the question though. I'm happy to gather analytics if needed.

wesrowe:
@allisonlu @ryguyk – Sorry I missed this discussion before the weekend, I missed the notification. Here's my view of the history of adding search functionality to the home page prototype: we definitely want to do it, as it's important to the usability testing. The only question was whether development of it would be "free" with reusing the search component, or would it be a little more challenging and need its own user story. 
The separate user story is written, it's #8362.

PR #1094, opened 2022-04-26
Preview server deploy should trigger after Daily Production Release.

timcosgrove:
## Description
The contents of the templating system can only change after a Daily Production Release, and so a re-deploy of the preview server should happen then. Content release cannot itself have new template or app structure, and so deploying upon changes is unneeded.


cweagans (approved):

ElijahLynn:
Thanks for clarifying how that works Tim and making this change!

PR #1097, opened 2022-04-28
Fix s3 bucket URLs referenced by CMS preview server

timcosgrove:
## Description
Preview.js was pointing to old S3 buckets for its static files. This updates to the current S3 buckets.

### DIFF
--- a/script/preview.js
+++ b/script/preview.js
@@ -94,16 +94,6 @@ const cacheDir = path.join(
 const app = express();
 const drupalClient = getDrupalClient(options);
 
-const urls = {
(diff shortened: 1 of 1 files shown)


olivereri (approved):

timwright12:
@timcosgrove if you'd like to use the variables already in the system: https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/constants/buckets-content.js

timwright12 (approved):

timcosgrove:
@timwright12 the bucket constants were already included in preview.js even. Thanks for pointing that out!

PR #1098, opened 2022-04-28
Vacms 8602 report broken links as blocks

timcosgrove:
## Description
Re-enables the 'new' Slack broken link reporting, and accounts for cases where we would send too much information via Slack message. 

Note that current Slack broken link reporting is left in content-release.yml until we can see the new version working, at which point we will remove the current version from content-release.yml


script/github-actions/check-broken-links-blocks.js
```diff
@@ -17,7 +16,8 @@ if (fs.existsSync(reportPath)) {
   const brokenLinks = JSON.parse(brokenLinksReport);
   const shouldFail =
     brokenLinks.isHomepageBroken ||
-    brokenLinks.brokenLinksCount > maxBrokenLinks;
+    brokenLinks.brokenLinksCount >
+      (brokenLinks.maxBrokenLinks ?? maxBrokenLinks);
```
ndouglas, line 20: What's the desired behavior if `brokenLinks.maxBrokenLinks` evaluates to 0 for whatever reason?

timwright12 (approved):
Feel free to reset ownership of these files

script/github-actions/check-broken-links-blocks.js
timcosgrove: It will evaluate to `brokenLinks.maxBrokenLinks`. The nullish coalescing operator only returns the right-hand value when the left-hand value is `null` or `undefined`; it's stricter than Javascript's normal falsy evaluation. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

That said, I had to look that up to be sure, so I appreciate you asking the question.

PR #1119, opened 2022-05-06
Remove conditional logic for Care we provide header so it always disp…

JayDarnell:
…lays

## Description

Removes the conditional logic around the Care we provide H3 tag so it always displays.

## Acceptance criteria

- [ ] The Care we provide at XXX H3 tag should still display as before


ndouglas (approved):

ryguyk (approved):

swirtSJW (approved):
Clean removal of feature flag check that is no longer needed.  Nice work.

PR #1123, opened 2022-05-09
VAVSA-9023: Add meta info to top task graphql queries.

ethanteague:
## Description
closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9023

## Testing done
build

## Acceptance criteria
- [ ] Visit one of each: Register for Care, Billing and Insurance, and Medical records
- [ ] Visually verify Meta title pattern: `<title>Make An Appointment | VA Eastern Oklahoma Health Care | Veterans Affairs</title>`


swirtSJW (approved):
Good catch Ethan.

maxx1128 (approved):
This looks good to me, but I think the description should be changed so it doesn't close 9023. This PR adds the needed info for the layouts but doesn't include them, so as is, it'll need a follow-up.

ethanteague:
The includes are in the templates, @maxx1128 . They just never added the graphql header info to the query.

PR #1141, opened 2022-05-20
Add manual trigger and schedule to CI job

cweagans:
## Description
Right now, we don't have a way to redeploy staging other than by pushing a commit to main in this repo. It would be helpful to periodically rebuild and redeploy dev and staging + have a way to manually do that.

This is probably not technically the _right_ way to do this, but it works for now. We'll probably look at upgrading the content release workflow to support deploying dev and staging as well, at which point we can back this change out.

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -6,6 +6,9 @@ on:
       - '**'
     tags-ignore:
       - '**'
+  workflow_dispatch:
+  schedule:
+    - cron: '17 10 * * *'
 
 env:
   CHROMEDRIVER_FILEPATH: /usr/local/share/chrome_driver/chromedriver


theodur (commented):
We have the [Manual dev/staging Deploy ](https://github.com/department-of-veterans-affairs/content-build/actions/workflows/manual-deploy-dev-staging.yml) workflow that does this

theodur:
The only thing is it doesn't run a new build, it just syncs the S3 build bucket to the deploy bucket, so the build step would need to be added to it

cweagans:
> The only thing is it doesn't run a new build,

Yep. We looked at this and it didn't do what we needed it to do. It seems like CI should probably run against new data every night anyway though -- that way, you get an early warning if there's something wrong before the daily release.

PR #1142, opened 2022-05-22
VACMS-8199: Adds static-data build step to content build

ryguyk:
## Description
See [#8199](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8199).

This standardizes a method to create static data files from Drupal data that can be fetched from the browser. As an example, this PR also creates one file that gets a start on [#8201](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8201).

## Testing done
- `yarn build` locally.
- Note logging message in screenshot below
- Confirm rest of build runs successfully

## Screenshots
image

## Notes
This PR creates a simple method for creating static data files from Drupal data. It is stable and useful, but needs further iteration   to address a couple issues:
- Add support for JSON:API queries
- Add support for caching queries and/or generated files


ethanteague (approved):
LEGIT :+1:

src/site/stages/build/drupal/static-data-files/generate.js
```diff
@@ -0,0 +1,55 @@
...
+const getApiClient = require('../api');
+const { logDrupal } = require('../utilities-drupal');
+const { DATA_FILE_PATH, DATA_FILES } = require('./config');
+
+function generateStaticDataFilesFromDrupal(
+  files,
+  buildOptions,
+  onlyPublishedContent = true,
+) {
+  if (!ENABLED_ENVIRONMENTS.has(buildOptions.buildtype)) {
```
swirtSJW, line 12: What environments would we not want these built on?

src/site/stages/build/drupal/static-data-files/vamcEhrSystem/index.js
```diff
@@ -0,0 +1,43 @@
+const query = `
+  query($onlyPublishedContent: Boolean!) {
+    nodeQuery(
```
swirtSJW, line 3: This is using graphql?

swirtSJW (commented):
THis is looking really great Ryan.   Thanks for taking this on.

I want to go through it with you and figure out what implementation details have been hit, and which ones haven't.   I'll connect up with you later.

src/site/stages/build/drupal/static-data-files/generate.js
ryguyk: @swirtSJW This is simply matching a pattern used elsewhere in content build where the build talks to Drupal. It's not immediately clear to me when there would ever be a `buildtype` passed in that would not be one of the `ENABLED_ENVIRONMENTS` (`localhost`, `vagovdev`, `vagovstaging`, `vagovprod`), but I wanted to keep the logic consistent so if somebody thought they were disabling Drupal, they would disable for this static-data-file integration as well as for the rest of the Drupal queries. My hunch is that this check always resolves to false, though.

src/site/stages/build/drupal/static-data-files/vamcEhrSystem/index.js
ryguyk: @swirtSJW Yes, it's using GraphQL. That was not my initial approach, so I initially got things wired up with JSON:API. To make it bulletproof, though, would require some more consideration, so I determined it was outside the initial scope. I created a [new issue](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9178).

theodur (changes requested):

src/site/stages/build/drupal/static-data-files/generate.js
swirtSJW: That's a great point @theodur Thanks for calling it out.

swirtSJW:
@ryguyk as we discussed earlier in the week, facilities team needs this to support grabbing the json from a view and moving it unaltered to the FE.   I'm going to be out next week so ping @JayDarnell  and @ethanteague  with any updates on Tuesday.

src/site/stages/build/drupal/static-data-files/generate.js
ryguyk: @theodur I added a simple check for the `--pull-drupal` flag. If it's there, the static files are generated. If not, they aren't generated. This is a simple first step. This needs to be expanded to actually cache the files and pull from the cache, but that is a bigger lift.

theodur (approved):

PR #1165, opened 2022-06-09
Vagov 0000 verbose download logging

timcosgrove:
## Description
This moves reporting of downloading an asset from after completion to before attempt. If there is no error, we can assume the asset was downloaded.

Right now, if an asset is problematic for any reason, we get a failure without any reporting of which asset is the problematic one. Reporting a successful download is not useful information in this case.


alexandec (approved):

timcosgrove:
This is not needed to diagnose image issues.

PR #1175, opened 2022-06-15
VACMS-9449 Adjust asset url rewriting

timcosgrove:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9449

This ensures that external asset URLs in HTML are not modified when internal URLs are modified.

### DIFF
--- a/src/site/stages/build/drupal/assets.js
+++ b/src/site/stages/build/drupal/assets.js
@@ -39,9 +39,19 @@ function replacePathInData(data, replacer, ancestors = new Set()) {
 function convertAssetPath(url) {
   // After this path are other folders in the image paths,
   // but it's hard to tell if we can strip them, so I'm leaving them alone
-  const withoutHost = url.replace(/^.*\/sites\/.*\/files\//, '');
-  const path = withoutHost.split('?', 2)[0];
+  // Check that this item is on a VA domain.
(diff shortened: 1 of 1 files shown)


ndouglas (approved):

src/site/stages/build/drupal/assets.js
```diff
@@ -39,9 +39,19 @@ function replacePathInData(data, replacer, ancestors = new Set()) {
 function convertAssetPath(url) {
   // After this path are other folders in the image paths,
   // but it's hard to tell if we can strip them, so I'm leaving them alone
-  const withoutHost = url.replace(/^.*\/sites\/.*\/files\//, '');
-  const path = withoutHost.split('?', 2)[0];
+  // Check that this item is on a VA domain.
+  // @todo Allow specified domains not of form *.cms.va.gov.
+  // @todo alternately, check against DRUPAL_ADDRESS, accounting for protocol.
+  const assetPath = url.replace(
```
swirtSJW, line 45: It would be nice is this could leverage 
`vaDomainRegex` that you added previously so that these two processes can not drift.

swirtSJW (commented):
One suggestion about making the pattern re-useable so that the file mover and the url rewriting stay in sync, otherwise they can drift like they did.

It would also be nice to know how this was and should be tested.

PR #1176, opened 2022-06-15
added closed to empty hour fields

jtmst:
## Description
Issue: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9390
Added logic to add 'Closed' to field office hours where there are no hours or comment

## Testing done
Visual testing on local env

## Screenshots
image

Screen Shot 2022-06-15 at 10 57 12 AM

## Acceptance criteria
- [ ] Visually confirm that facility hours are shown properly, and that 'Closed' is appearing instead of any blank hour fields example: (butler-health-care/billing-and-insurance)

## Definition of done

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -72,7 +72,7 @@ jobs:
         with:
           mode: start
           github-token: ${{ env.VA_VSP_BOT_GITHUB_TOKEN }}
-          ec2-image-id: ami-036668e0b11e4bf99
+          ec2-image-id: ${{ env.RUNNER_AMI_ID }}
           ec2-instance-type: ${{ env.INSTANCE_TYPE }}
           subnet-id: ${{ env.SUBNET_ID }}
           security-group-id: sg-0e23b56be3798e3a1
--- a/src/site/facilities/health_care_region_detail_content.drupal.liquid
+++ b/src/site/facilities/health_care_region_detail_content.drupal.liquid
@@ -19,10 +19,7 @@
 
 
   {% if fieldTableOfContentsBoolean != empty and fieldTableOfContentsBoolean %}
-    <nav id="table-of-contents" aria-labelledby="on-this-page">
-      <h2 id="on-this-page" class="vads-u-margin-bottom--2 vads-u-font-size--lg">On this page</h2>
-      <ul class="usa-unstyled-list" role="list"></ul>
-    </nav>
+    <va-on-this-page class="vads-u-margin-left--1 vads-u-margin-bottom--0 vads-u-padding-bottom--0"></va-on-this-page>
   {% endif %}
 
   {% if fieldFeaturedContent != empty and fieldFeaturedContent.length > 0 %}
--- a/src/site/includes/blog-promo.html
+++ b/src/site/includes/blog-promo.html
@@ -1,6 +1,11 @@
 <!-- 
-  Blog promo banner
+  Blog promo section on home page
+
+  Homepage redesign iteration (/homepage-test)
+  Date: April 2022
+  Author: allison@cityfriends.tech
 -->
+
   <!-- Hero Content -->
   <div class="vads-u-background-color--primary-darker">
     <div class="vads-l-grid-container vads-u-padding-x--0 homepage-blog">
--- a/src/site/includes/common-tasks.html
+++ b/src/site/includes/common-tasks.html
@@ -1,7 +1,7 @@
 <!-- 
   Search and Common tasks hub on home page
 
-  Homepage iteration 
+  Homepage redesign iteration (/homepage-test)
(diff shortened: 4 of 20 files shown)


src/site/includes/hours-item.liquid
```diff
@@ -1,15 +1,20 @@
...
+    {% endif %}
+    {% if hours.starthours %}
+        {{ hours.starthours | officeHoursTimeFormatter}}
+        to
+    {% endif %}
+    {% if hours.endhours %}
+        {{ hours.endhours | officeHoursTimeFormatter}}
+    {% endif %}
+
+    {{ hours.comment }}
```
kind-claudio, line 15: could lines 7 - 15 be moved into an `{% else %}` block down below, at line 19? This way we don't run the other conditional blocks if we don't need to.

src/site/includes/hours.liquid
```diff
@@ -28,24 +28,28 @@
...
-								<span class="abbrv-day">Sun.</span>
-							{% endif %}
-						</li>
+                        <li>
+                            {% if headerType == 'clinical' %}
+                                <b class="abbrv-day">Sat:</b>
+                                Closed
+                            {% else %}
+                                <span class="abbrv-day">Sat.</span>
+                                Closed
```
kind-claudio, line 37: You could remove "Closed" from these two conditional blocks, and instead place it only once at the end of the `<li />` block, after line 38, since it be displayed/rendered regardless of the conditional results.

src/site/includes/hours.liquid
```diff
@@ -28,24 +28,28 @@
...
+                    {% if sunday %}
+                        {% include "src/site/includes/hours-item.liquid" with hours = sunday %}
+                    {% else %}
+                        <li>
+                            {% if headerType == 'clinical' %}
+                                <b class="abbrv-day">Sun:</b>
+                                Closed
+                            {% else %}
+                                <span class="abbrv-day">Sun.</span>
+                                Closed
```
kind-claudio, line 50: Same comment as the one above:
 You could remove "Closed" from these two conditional blocks, and instead place it only once at the end of the <li /> block, but after line 51, since it be displayed/rendered regardless of the conditional results.

kind-claudio (changes requested):

src/site/includes/hours-item.liquid
maxx1128: ☝️ I second this

src/site/includes/hours.liquid
jtmst: Good call, fixed those

src/site/includes/hours-item.liquid
jtmst: Good idea,  just commited that change

kind-claudio (approved):

src/site/includes/hours.liquid
```diff
@@ -28,24 +28,26 @@
...
-								<span class="abbrv-day">Sun.</span>
-							{% endif %}
-						</li>
+                        <li>
+                            {% if headerType == 'clinical' %}
+                                <b class="abbrv-day">Sat:</b>
+                            {% else %}
+                                <span class="abbrv-day">Sat.</span>
+                            {% endif %}
+							Closed
```
maxx1128, line 37: Should this "Closed" text be in another conditional like the one in the other file, that checks if there are no hours or comments?

maxx1128 (changes requested):
Looks good, one small question then I should be ready to approve.

src/site/includes/hours.liquid
jtmst: That check happens in the above if block, this Closed text will only render if Saturday / Sunday hours arent listed

wullaski (changes requested):

maxx1128 (approved):

src/site/includes/hours-item.liquid
```diff
@@ -1,15 +1,20 @@
 <li>
-	{% if headerType == 'clinical' %}
-		<b class="abbrv-day">{{ hours.day | officeHoursDayFormatter }}:</b>
+    {% if headerType == 'clinical' %}
+        <b class="abbrv-day">{{ hours.day | officeHoursDayFormatter }}:</b>
+    {% else %}
+        <span class="abbrv-day">{{ hours.day | officeHoursDayFormatter }}.</span>
+    {% endif %}
+
+	{% if !hours.starthours && !hours.endhours && !hours.comment %}
```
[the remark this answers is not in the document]
jtmst: The most common use of the comment block is editors writing in "closed" and i believe their is drupal logic that formats that.  This would result in "closed" being rendered twice.  The desired logic in the original issue is to only make this assumption if there is no comment.

wullaski (approved):

PR #1213, opened 2022-07-12
9718 online location added

jtmst:
## Description
Logic added to include location for online events

## Testing done
Local visual confirmation

## Screenshots
Screen Shot 2022-07-12 at 11 22 13 AM

## Acceptance criteria
- [ ] "This is an online event" displays as the location for online events

## Definition of done

### DIFF
--- a/src/site/layouts/event.drupal.liquid
+++ b/src/site/layouts/event.drupal.liquid
@@ -74,11 +74,7 @@
                     <p class="vads-u-margin--0">{{ fieldLocationHumanreadable }}</p>
                   </div>
                 {% else %}
-
                   <div class="vads-u-display--flex vads-u-flex-direction--column">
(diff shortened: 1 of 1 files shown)


wullaski (changes requested):
Can you add your fieldLocationType check to the set of 'or' statements above and add an elseif for it that way we don't have to repeat the 'where:' block.

ryguyk (changes requested):
I think I agree with @wullaski. There are three conditions that should result in "Where:" being displayed. Currently, two are handled together, then the third one on its own. I think they should all be handled together (i.e. move new condition into block above), or all be moved individually (i.e. split up two conditions above to look more like the new one). This might be a matter of personal taste.

wullaski (approved):

ryguyk (approved):

PR #1222, opened 2022-07-19
VACMS-9840: Works around a Drupal bug with updating alt text in image media.

ndouglas:
## Description
See department-of-veterans-affairs/va.gov-cms#9840.

If you update the alt text of a Media Image in Drupal, the alt text is not actually changed; see this core issue: https://www.drupal.org/project/drupal/issues/3232414 (which I've yet to chime in on)

I've changed the GraphQL queries referring to thumbnail.alt and the liquid templates so that the content build will reflect the latest alt text, rather than the original alt text.

I made the simplest and probably dumbest changes possible because I don't really speak GraphQL and I'm an idiot even when I'm fluent in the thing in question, so apologies if I'm missing something obvious.

## Acceptance criteria
- [ ]

## Definition of done


olivereri (approved):

ndouglas:
Thank you so much for the review, @maxx1128 -- I'll look into that.  I saw that when I was searching for things to replace, but I assumed it was an image field and not a media field.  That might've been an incorrect assumption -- I'll trace through and find out what that type actually is.  Thank you!

PR #1226, opened 2022-07-20
Conditionally use the expandable alert component

bkjohnson:
## Description

This updates a template to follow [Design System guidance on the Additional Info component](https://design.va.gov/components/additional-info#when-to-consider-something-else):

> Do not use this component inside an [Alert](https://design.va.gov/components/alert). Try the [Alert - Expandable](https://design.va.gov/components/alert-expandable) component instead, especially when the Alert is within the page content and not at the top of the page.

## Testing done

Local build :eyes: 

## Screenshots

The `/cincinnati-health-care/health-services/suicide-prevention` page:

### Before

Screenshot of the page on staging with the "How do I talk to someone right now" alert in the viewport and the "Find out how to get support anytime, day or night" additional info component inside the alert expanded.

### After

Screenshot of a browser page showing `<va-expandable-alert>` with `data-template="blocks/alert"`

A regular alert on the `careers-employment/vocational-rehabilitation/how-to-apply` page:

Screenshot of a browser page showing a regular alert with the `data-template="blocks/alert"`

However, the changes here still don't address content structures like this which would still give us an additional-info lookalike inside of an alert component:
 (on page `/resources/choosing-between-urgent-and-emergency-care/`)
 
Screenshot of a page showing an alert `date-template="paragraphs/alert"` with a `data-template="paragraphs/expandable_text"` inside which looks like an additional info component.

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/site/blocks/alert.drupal.liquid
+++ b/src/site/blocks/alert.drupal.liquid
@@ -5,29 +5,16 @@
 {% endif %}
 
 {% if isPreview or alert.entityPublished %}
-  <va-alert data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" status="{{ alertType }}" class="vads-u-margin-top--3" role="alert">
-    <h2 slot="headline" class="vads-u-font-size--h3">
-      {{ alert.fieldAlertTitle }}
-    </h2>
-
-    {% if alert.fieldAlertContent.entity.fieldTextExpander == empty %}
-      {{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}
-    {% endif %}
-
     {% if alert.fieldAlertContent.entity.fieldTextExpander %}
-      {% comment %}
-        NOTE: .additional-info-container is a class utilized by
-        createAdditionalInfoWidget.js to add toggle functionality to info alerts
-      {% endcomment %}
-      <div data-alert-box-title="{{ alert.fieldAlertTitle }}" data-label="{{ alert.fieldAlertContent.entity.fieldTextExpander }}" class="form-expanding-group borderless-alert additional-info-container">
-        <div class="additional-info-title">
-          {{ alert.fieldAlertContent.entity.fieldTextExpander }}
-        </div>
-
-        {% if alert.fieldAlertContent.entity.fieldWysiwyg %}
-          <div class="additional-info-content usa-alert-text" hidden>{{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}</div>
-        {% endif %}
-      </div>
+      <va-alert-expandable status="{{ alertType }}" trigger="{{alert.fieldAlertTitle }}" data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" class="vads-u-margin-top--3" role="alert">
+        {{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}
+      </va-alert-expandable>
+    {% elsif alert.fieldAlertContent.entity.fieldTextExpander == empty %}
+      <va-alert data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" status="{{ alertType }}" class="vads-u-margin-top--3" role="alert">
+        <h2 slot="headline" class="vads-u-font-size--h3">
+          {{ alert.fieldAlertTitle }}
+        </h2>
+        {{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}
+      </va-alert>
     {% endif %}
-  </va-alert>
 {% endif %}
--- a/src/site/blocks/alert.drupal.liquid
+++ b/src/site/blocks/alert.drupal.liquid
@@ -10,7 +10,7 @@
         {{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}
       </va-alert-expandable>
     {% elsif alert.fieldAlertContent.entity.fieldTextExpander == empty %}
-      <va-alert data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" status="{{ alertType }}" class="vads-u-margin-top--3" role="alert">
+      <va-alert status="{{ alertType }}" data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" class="vads-u-margin-top--3" role="alert">
         <h2 slot="headline" class="vads-u-font-size--h3">
           {{ alert.fieldAlertTitle }}
         </h2>


src/site/blocks/alert.drupal.liquid
```diff
@@ -5,29 +5,16 @@
...
-      <div data-alert-box-title="{{ alert.fieldAlertTitle }}" data-label="{{ alert.fieldAlertContent.entity.fieldTextExpander }}" class="form-expanding-group borderless-alert additional-info-container">
-        <div class="additional-info-title">
-          {{ alert.fieldAlertContent.entity.fieldTextExpander }}
-        </div>
-
-        {% if alert.fieldAlertContent.entity.fieldWysiwyg %}
-          <div class="additional-info-content usa-alert-text" hidden>{{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}</div>
-        {% endif %}
-      </div>
+      <va-alert-expandable status="{{ alertType }}" trigger="{{alert.fieldAlertTitle }}" data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" class="vads-u-margin-top--3" role="alert">
```
bkjohnson, line 9: One thing to be aware of is that this uses the `fieldAlertTitle` as the text that triggers expanding/collapsing instead of `fieldTextExpander` like we had previously. But with this new structure of not nesting an additional info inside an alert, we have to either lose the alert heading or the additional info trigger text.

theodur (approved):

src/site/blocks/alert.drupal.liquid
swirtSJW: @bkjohnson Sorry I am only half-following this.  In the CMS the field fieldTextExpander is a string and is designed to contain text that will be shown.  I don't think we can just relegate it to a bool.   Let me find some examples where it is being used  so we can see what the impact of this change would be.

timwright12: @bkjohnson I think I remembered @joshkimux mentioning that we're moving away from using `role="alert"` outside of certain situations. Might want to check with him

joshkimux: @bkjohnson hi hi! 👋 @timwright12 thanks for roping me in here, I can confirm that we're trying to use the role="alert" to limited contexts and situations (it's kinda like the equivalent of a mobile tornado warning).

If I have a little more context here I can help out more 🙏

bkjohnson: @swirtSJW here are all of the liquid templates where `fieldTextExpander` is being used:

- [`paragraphs/expandable_text.drupal.liquid`](https://github.com/department-of-veterans-affairs/content-build/blob/d73c90ae075f7d1e9f18a47e45b29c59482bbf28/src/site/paragraphs/expandable_text.drupal.liquid#L16)
- [`paragraphs/spanish_translation_summary.drupal.liquid`](https://github.com/department-of-veterans-affairs/content-build/blob/d73c90ae075f7d1e9f18a47e45b29c59482bbf28/src/site/paragraphs/spanish_translation_summary.drupal.liquid#L21)
- [`blocks/alert.drupal.liquid`](https://github.com/department-of-veterans-affairs/content-build/blob/d73c90ae075f7d1e9f18a47e45b29c59482bbf28/src/site/blocks/alert.drupal.liquid#L24)

All three of them are using additional info components judging by the classes, but only one of them would hardcode an additional info component inside of an alert. Even with this change, that value could still be rendered for `expandable_text` entities, as my third screenshot shows.

---

@joshkimux, I opened this PR after [this slack discussion](https://dsva.slack.com/archives/C01DBGX4P45/p1658338829178029) asking a question about this nesting pattern. I originally wrote it as a draft to act as a demo of how we could use an expandable alert instead of nesting an additionl info inside of an alert, and then after the DST updated guidance I made it a full PR.

joshkimux: Got it! Thanks for the context @bkjohnson . If this alert:

- isn't dynamically added into the page
- is not critical for the user to know immediately

We might want to avoid using `role="alert"` here. 

If this alert is dynamically added into the page in response to a user's action, we could use `aria-live="polite"` instead for a more gentle way of letting them know something new has appeared.

bkjohnson: These aren't dynamically added to the page, but I don't know whether or not the alert will contain critical information. Would an [alert like this](https://staging.va.gov/shreveport-health-care/make-an-appointment/#how-do-i-talk-to-someone-right) meet that criteria?

A screenshot of the linked page above, starting at the "Mental health care appointments" header. There is a warning alert underneath with the header "How do I talk to someone right now" and the additional info component inside is expanded, showing more content and phone numbers.

Those types of alerts are also seen on [suicide prevention pages](https://staging.va.gov/san-francisco-health-care/health-services/suicide-prevention/)

bkjohnson:
Closing per @swirtSJW's [comment](https://dsva.slack.com/archives/CT4GZBM8F/p1659733819698119?thread_ts=1659717970.256749&cid=CT4GZBM8F):

>  I don't think we can just make the change you are proposing without getting the editors in charge of that content to take a look and make changes.

swirtSJW:
Adding this link to slack discussion.  https://dsva.slack.com/archives/CT4GZBM8F/p1659717970256749

Summary: 
This is essentially a breaking change as the content being displayed  is not what the editor had intended.
We can not just on a template basis elect to throw away one field of content and use another.  They are not equivalent 
image

bkjohnson:
For future reference, here's an example of keeping all of the current content. This diff shows a header being added above `<va-alert-expandable>` and the `trigger` prop switching to the `fieldTextExpander` value.

```diff
diff --git a/src/site/blocks/alert.drupal.liquid b/src/site/blocks/alert.drupal.liquid
index 28699b450..093a082a2 100644
--- a/src/site/blocks/alert.drupal.liquid
+++ b/src/site/blocks/alert.drupal.liquid
@@ -6,7 +6,8 @@
 
 {% if isPreview or alert.entityPublished %}
     {% if alert.fieldAlertContent.entity.fieldTextExpander %}
-      <va-alert-expandable status="{{ alertType }}" trigger="{{alert.fieldAlertTitle }}" data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" class="vads-u-margin-top--3" role="alert">
+      <h2>{{ alert.fieldAlertTitle }}</h2>
+      <va-alert-expandable status="{{ alertType }}" trigger="{{ alert.fieldAlertContent.entity.fieldTextExpander }}" data-template="blocks/alert" data-entity-id="{{ alert.entityId }}" class="vads-u-margin-top--3" role="alert">
         {{ alert.fieldAlertContent.entity.fieldWysiwyg.processed }}
       </va-alert-expandable>
     {% elsif alert.fieldAlertContent.entity.fieldTextExpander == empty %}
```

Screenshot of a page with an h2 above an expandable alert component and no other content underneath the h2

src/site/blocks/alert.drupal.liquid
joshkimux: This is tricky, because the content _feels_ like it's very important, but the visual design of it is far from it (it's not the first, major piece of content you see when you open the page). The use of `role="alert"` would be more similar to a red banner at the top of the page.

We should loop in our content team, @AngelaFowler82 , and @laflannery to discuss this in more detail for the future. Thanks so much @bkjohnson for the added context here, you always ask the right questions and do amazing work 🙏

DanielleThierryUSDSVA:
Hi, all. Per @joshkimux's note, I'd be happy to talk more about these types of alerts. Feel free to grab some time on my calendar if that's helpful. Thanks!

PR #1228, opened 2022-07-22
9794: Add place schema

wullaski:
## Description

Adds schema for place to liquid template

## Testing done

- Visual with inspector
- Rich results test
  - https://search.google.com/test/rich-results/result/r%2Flocal-business?id=d2JRyeUkLIGz1bLGcBStgw

## Screenshots

Screen Shot 2022-07-28 at 4 41 52 PM

## Acceptance criteria
- [x] A property of [Place](https://schema.org/Place) exists with the following sub-properties
  - [x] [name](https://schema.org/name) - Uses node title 
  - [x] [address](https://schema.org/address) - Uses field_address
  - [x] [telephone](https://schema.org/telephone) - Uses field_phone_number
  - [x] [openingHours](https://schema.org/OpeningHoursSpecification) - Uses field_office_hours
  - [x] image - Uses field_media
  - [x] [geo](https://schema.org/geo) - Uses field_geolocation (if stored in CMS, if not use FE generated data)
  - [x] *OPTIONAL* [branchCode](https://schema.org/branchCode) - Can store Facility ID (field_facility_locator_api_id)  

- [x] tests added for filter

## Definition of done


maxx1128 (approved):
The code looks good, and I confirmed the code renders as expected locally.

Screen Shot 2022-07-27 at 11 35 47 AM

ndouglas (approved):

src/site/layouts/health_care_local_facility.drupal.liquid
```diff
@@ -74,12 +74,17 @@
...
+                          "@type": "GeoCoordinates",
+                          "latitude": "{{ fieldGeolocation.lat }}",
+                          "longitude": "{{ fieldGeolocation.lon }}"
+                        }
                       }
                     </script>
                     <h3
                         class="vads-u-font-size--lg vads-u-margin-top--0 vads-u-line-height--1 vads-u-margin-bottom--1">
-                      Address
+                      Address {{ fieldGeolocation.lat }}
```
maxx1128, line 87: **Comment:** Is the lattitude supposed to be included in the header? Unsure how that would look on the front-end.

wullaski: Good catch! No it's not that's a test I forgot to clean up.

ryguyk (commented):
These GraphQL changes look good. That said, I wonder if it's necessary to update all of these queries before they are actually used. This PR seems to only be concerned with adding schema attributes for the `health_care_local_facility` template. Why update the GraphQL for all the other places if they don't need it and it's not pulled out into a reusable fragment. Are there plans to use the lat/long coordinates in other templates as well?

wullaski:
> These GraphQL changes look good. That said, I wonder if it's necessary to update all of these queries before they are actually used. This PR seems to only be concerned with adding schema attributes for the `health_care_local_facility` template. Why update the GraphQL for all the other places if they don't need it and it's not pulled out into a reusable fragment? Are there plans to use the lat/long coordinates in other templates as well?

I think I agree that we should only update the ones for this template. I was having trouble determining which queries mapped to which templates. I can go ahead and remove the geo block from the files that don't need it.

PR #1242, opened 2022-08-03
9997 fe queries for vamc content types need field administration

wullaski:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9997
All content types that can potentially appear in VAMC Lovell need to have field_administration added to the graphQL query.

## Testing done

Visual developer console

## Acceptance criteria
- The graphQL queries for the following content types include field_administration 
   - Facility specific pages
      - [x] VAMC Detail Page / health_care_region_detail_page
      - [x] VAMC Facility / health_care_local_facility
      - [x] VAMC Facility Health Service / health_care_local_health_service
      - [x] VAMC System Health Service / regional_health_care_service_des
      - [x] VAMC Facility Non-clinical Service / vha_facility_nonclinical_service
      - [x] VAMC System / health_care_region_page
      - [x] VAMC System Banner Alert with Situation Updates / full_width_banner_alert
      - [x] VAMC System Billing and Insurance / vamc_system_billing_insurance
      - [x] VAMC System Locations List / locations_listing
      - [x] VAMC System Medical Records Office / vamc_system_medical_records_offi
      - [x] VAMC System Operating Status / vamc_operating_status_and_alerts
      - [x] VAMC System Policies Page / vamc_system_policies_page
      - [x] VAMC System Register for Care / vamc_system_register_for_care
      - [x] Health Services List / health_services_listing
   - Spanning all products including facilities.
      - [x] Events List / event_listing
      - [x] Event / event
      - [x] Leadership List / leadership_listing
      - [x] Staff Profile / person_profile
      - [x] News Releases List / press_releases_listing
      - [x] News Release / press_release
      - [x] Stories List / story_listing
      - [x] Story / news_story

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/graphql/bioPage.graphql.js
+++ b/src/site/stages/build/drupal/graphql/bioPage.graphql.js
@@ -6,61 +6,68 @@ const entityElementsFromPages = require('./entityElementsForPages.graphql');
 const { generatePaginatedQueries } = require('../individual-queries-helpers');
 
 const personProfileFragment = `
- fragment bioPage on NodePersonProfile {
-  ${entityElementsFromPages}
-  fieldNameFirst
-  fieldLastName
-  fieldSuffix
-  fieldDescription
-  fieldEmailAddress
-  fieldPhoneNumber
-  fieldCompleteBiographyCreate
-  fieldCompleteBiography { entity { url } }
-  fieldOffice {
-      entity {
-        entityLabel
-        entityType
-        ...on NodeHealthCareRegionPage {
-          ${entityElementsFromPages}
-          title
+  fragment bioPage on NodePersonProfile {
+    ${entityElementsFromPages}
+    fieldNameFirst
+    fieldLastName
+    fieldSuffix
+    fieldDescription
+    fieldEmailAddress
+    fieldPhoneNumber
+    fieldCompleteBiographyCreate
+    fieldCompleteBiography { entity { url } }
+    fieldOffice {
+        entity {
+          entityLabel
+          entityType
+          ...on NodeHealthCareRegionPage {
+            ${entityElementsFromPages}
+            title
+          }
         }
       }
-    }
-  fieldIntroText
-  fieldPhotoAllowHiresDownload
-  fieldMedia {
-    thumbnail: entity {
-      ... on MediaImage {
-        image {
-          alt
-          title
-          derivative(style: _23MEDIUMTHUMBNAIL) {
-            url
-            width
-            height
+    fieldIntroText
+    fieldPhotoAllowHiresDownload
+    fieldMedia {
+      thumbnail: entity {
+        ... on MediaImage {
+          image {
+            alt
+            title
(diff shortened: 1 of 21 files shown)


src/site/stages/build/drupal/graphql/bioPage.graphql.js
```diff
@@ -6,61 +6,68 @@ const entityElementsFromPages = require('./entityElementsForPages.graphql');
...
-    thumbnail: entity {
-      ... on MediaImage {
-        image {
-          alt
-          title
-          derivative(style: _23MEDIUMTHUMBNAIL) {
-            url
-            width
-            height
+    fieldIntroText
```
swirtSJW, line 29: Thanks for cleaning up the indention.

src/site/stages/build/drupal/graphql/healthCareLocalFacilityPage.graphql.js
```diff
@@ -139,6 +139,20 @@ const healthCareLocalFacilityPageFragment = `
               }
             }
           }
+          fieldAdministration {
+            entity{
+              ... on TaxonomyTermAdministration {
```
swirtSJW, line 144: I don't think we need this one, but I might be wrong and it causes no harm being there, so I think it's fine.

swirtSJW (approved):
@wullaski Code looks great and checks out good against the list of 21 content types needing this addition.

 Nice work.

swirtSJW:
Note for other reviewers.  There is nothing in content-build that is using this property, but it lays the groundwork for a bunch of work that is coming over the next several sprints.  The main thing is the queries should not break... which tests are showing they are not.

ryguyk:
> Note for other reviewers. There is nothing in content-build that is using this property, but it lays the groundwork for a bunch of work that is coming over the next several sprints. The main thing is the queries should not break... which tests are showing they are not.

@swirtSJW I find myself a bit hesitant with this sort of proactive addition to GraphQL queries. I tend to fall on the side of adding to GraphQL when a template change actually necessitates it.

Two drawbacks that I see:
1. Adds unnecessary build time.
2. Introduces confusion over trying to ensure we've found "all the places" to make the change. This feels like a lift that would best be undertaken on an as-needed basis, which would significantly lower the surface area.

I'd welcome thoughts from anybody else. I'm very open to being convinced otherwise, as I imagine there is compelling argument from the other side of the table.

wullaski:
@ryguyk, I agree with you that this could add the field where it might not be necessary and increase build time. That said this is blocking the [issue](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9945) I'm working on now and I'd have to add changes I've made here to that other issue and in a lot of the same places. If I did it that way I'm not confident that I would add the field to all of the content types that require it now or might require it in the future due to changes with the content. If that makes sense. I'm happy to talk this out with the group.

swirtSJW:
Moving this to integration branch.

PR #1254, opened 2022-08-11
Increase build timeout from 30m to 45m

ElijahLynn:
Jobs are timing out and failing. Sometimes that step takes 22m but a couple runs lately have timeout at 30m:

image

https://github.com/department-of-veterans-affairs/content-build/runs/7795882449?check_suite_focus=true

image

## Description
e.g. second failure from a timeout here > https://github.com/department-of-veterans-affairs/content-build/runs/7795882449?check_suite_focus=true

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -259,7 +259,7 @@ jobs:
 
       - name: Build
         run: yarn build --buildtype=${{ env.BUILDTYPE }} --asset-source=local --drupal-address=${{ env.DRUPAL_ADDRESS }} --drupal-user=${{ env.DRUPAL_USERNAME }} --drupal-password="${{ env.DRUPAL_PASSWORD }}" --pull-drupal --drupal-max-parallel-requests=15 --no-drupal-proxy --verbose | tee build-output.txt
-        timeout-minutes: 30
+        timeout-minutes: 45
         env:
           NODE_ENV: production
 


ElijahLynn:
I'm still looking into if there is one thing that is taking abnormally long.

ElijahLynn:
On the timeout build, GQL seems normal. So maybe it is the file downloads?

> https://prod.cms.va.gov/ response time: : 2:55.254 (m:ss.mmm)
> Total time to load content from GraphQL: 2:57.619 (m:ss.mmm)
> Drupal successfully loaded!

alexandec (approved):

cweagans:
Looks like the file downloads are the culprit to me.

https://github.com/department-of-veterans-affairs/content-build/runs/7795404712?check_suite_focus=true says it took 99465ms (succeeded).

https://github.com/department-of-veterans-affairs/content-build/runs/7794734477?check_suite_focus=true was 729281ms (failed).

cweagans (changes requested):
The CMS considers the state stale if it hasn't gotten a status update in more than 40 minutes. 45 minutes is going to cause issues here. It would be better to let it fail + then the CMS will retry.

ElijahLynn:
> Looks like the file downloads are the culprit to me.

Thanks, I couldn't see the end time in the one I was looking at but that one you linked to did finish that step and shows the end time. So, I just verified on the PROD EFS volume that the files are stored on that we aren't reaching any throughput limits:

image
https://console.amazonaws-us-gov.com/efs/home?region=us-gov-west-1#/file-systems/fs-1c9ba71d?tabId=monitoring

ElijahLynn:
> The CMS considers the state stale if it hasn't gotten a status update in more than 40 minutes. 45 minutes is going to cause issues here. It would be better to let it fail + then the CMS will retry as a result of the error (as of today).

Good to know! I think we should close this for now then. I had just posted this in Slack too, so I'm on the same page:

> Thanks, maybe we keep the PR open for a bit to see if it flickers again. I am thinking maybe it was just an issue with runner resource utilization or a network path thing.

timcosgrove:
FWIW I don't think we should do this, except as a possible stopgap measure. If the build is consistently over 30 minutes, that is a problem that should be addressed directly.

PR #1265, opened 2022-08-17
VACMS-10312: Fixes defect with missing Outreach Materials

ryguyk:
## Description

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10312

Outreach materials (`outreach_assets`) query was broken in a non-obvious way. It was still returning results, but not limited to the `outreach_assets` type. So, it was returning all nodes, and the 10000 limit was insufficient. This corrects the filter.

## Testing done
Testing done in GraphQL explorer.

### DIFF
--- a/src/site/stages/build/drupal/graphql/file-fragments/outreachAssets.graphql.js
+++ b/src/site/stages/build/drupal/graphql/file-fragments/outreachAssets.graphql.js
@@ -1,5 +1,8 @@
 const outreachAssets = `
-  outreachAssets: nodeQuery(filter: {conditions: [{field: "type", value: "outreach_asset", field: "status", value: ["1"], enabled: $onlyPublishedContent}]}, limit: 10000) {
+  outreachAssets: nodeQuery(filter: {conditions: [
+    { field: "type", value: "outreach_asset" },
+    { field: "status", value: ["1"], enabled: $onlyPublishedContent }
+  ]}, limit: 10000) {
     entities {
       ... on NodeOutreachAsset {
         entityId


src/site/stages/build/drupal/graphql/file-fragments/outreachAssets.graphql.js
```diff
@@ -1,5 +1,8 @@
 const outreachAssets = `
-  outreachAssets: nodeQuery(filter: {conditions: [{field: "type", value: "outreach_asset", field: "status", value: ["1"], enabled: $onlyPublishedContent}]}, limit: 10000) {
+  outreachAssets: nodeQuery(filter: {conditions: [
+    { field: "type", value: "outreach_asset" },
+    { field: "status", value: ["1"], enabled: $onlyPublishedContent }
+  ]}, limit: 10000) {
```
ndouglas, line 5: Is 10K still needed? 😬

ndouglas (approved):

src/site/stages/build/drupal/graphql/file-fragments/outreachAssets.graphql.js
ryguyk: Probably not, but I'm not sure it really matters. We should likely just remove the limit.

ryguyk: Actually, I think removing results in some default limit.

swirtSJW: I have a feeling the limit was set that high because content was already being bumped and nobody caught why... just a a hunch.

ryguyk: @ndouglas thoughts on a downside to a very high limit to indicate "we want all of them"?

swirtSJW (approved):
Approved, but please dont merge before I get an issue created that we can link this number to.

jtmst (approved):

swirtSJW:
Title updated to reflect ticket number https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10312

src/site/stages/build/drupal/graphql/file-fragments/outreachAssets.graphql.js
ndouglas: I feel like the limit being set to 10000 should have set off alarm bells in the first place, but it didn't.  It'd be nice to lower it shortly, but we should have some kind of structural solution to detecting this kind of wildly inefficient query... might be worth looking at other queries that have high limits and see where they're necessary and where they might indicate a programming issue.  IDK if it matters, though.  This might just be a weird one-off.

ryguyk: We have a lot of queries, I imagine, where we just want all of them. No limit. Drupal's GraphQL defaults the limit to 10 if no limit is specified. This feels weird, but there's probably good reasons I can think of. I'm not sure that `limit: 10000` is going to have any effect versus `limit: 1000` in this case, aside from the latter option failing at an earlier point in the future when the number of nodes in question first exceeds 1000. 

That said, I see the point about 10000 being a red flag. Why did somebody use this in the first place? Perhaps it's a pattern employed to capture "all of them". I can't say for sure, but I'll defer to others if there's a strong feeling that this should be changed. I can see both sides.

allisonlu (approved):

PR #1262, opened 2022-08-16
Reapply "Build prototype for new focus styles"

bkjohnson:
## Description

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1058

This reverts the revert in department-of-veterans-affairs/content-build#1261 :upside_down_face:. 

The goal here is to create a prototype page with the double box-shadow focus styling that can be used in user testing sessions. I have modified the original boolean expression so that our web component script gets loaded on all pages that aren't the prototype. The prototype will be the [COVID-19 page](https://dev.va.gov/health-care/covid-19-vaccine/) in the dev environment.

Just to be safe I'm planning on merging this after the daily deploy, and not on Friday.

## Additional details

The reason why I'm removing the "regular" web component loader script on the prototype page is because the browser will use whichever custom element constructor gets defined in its registry first. The [spec says](https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition):

>  When invoked, the `define(name, constructor, options)` method must run these steps:
> ...
> 3. If this [CustomElementRegistry](https://html.spec.whatwg.org/multipage/custom-elements.html#customelementregistry) contains an entry with [name](https://html.spec.whatwg.org/multipage/custom-elements.html#concept-custom-element-definition-name) `name`, then throw a ["NotSupportedError"](https://webidl.spec.whatwg.org/#notsupportederror) [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException).

I didn't see a "NotSupportedError" in the console when I was defining different versions of the web components under the same name so it may be getting caught already, but we get undesirable behavior where the web component internals don't have the prototype focus styles but other global elements do:

[Video showing focus styles not matching between global links and links internal to a web component's shadow DOM](https://user-images.githubusercontent.com/2008881/185467559-73da3f3d-7a37-42e6-8f11-f2a5d1c53f6c.webm)

### Dev build

These screenshots were taken on a site that was built using:

```
yarn build --buildtype vagovdev
```

And served using:

```
yarn serve --buildtype vagovdev
```

<details>

<summary>COVID-19 page on a dev build</summary>

This prototype has the double box-shadow focus styling so that it can be tested with Veterans.

A screenshot of the header with the "Contact us" link highlighted. It has the proposed double box-shadow focus styling

A screenshot of the page content. A link in the "On this page" component is highlighted using the box-shadow focus styling

</details>

<details>

<summary>Other page on a dev build</summary>

There are no changes here, and web components still work.

[...]

PR #1266, opened 2022-08-18
Assert all web component tags are defined

bkjohnson:
## Description

Related to #1262 

After accidentally deploying a prod build which didn't include the script that defines the web component tags in the browser, it seemed like a good idea to try to prevent that in the future with a test.

cc @department-of-veterans-affairs/platform-design-system-fe 

## Testing done

E2E tests. In [this commit](https://github.com/department-of-veterans-affairs/content-build/pull/1266/commits/4fdd01e23c364d664a54bdce437b35a41ae404f4) we can see that [this new test will fail](https://github.com/department-of-veterans-affairs/content-build/runs/7902604966?check_suite_focus=true#step:9:117) when the script responsible for defining web components isn't on the page.

A screenshot of the GitHub Action link above showing the Cypress failure when `va-accordion` isn't defined.

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "@babel/preset-react": "^7.14.5",
     "@babel/register": "^7.15.3",
     "@department-of-veterans-affairs/eslint-plugin": "^1.1.0",
-    "@department-of-veterans-affairs/web-components": "^4.11.0",
+    "@department-of-veterans-affairs/web-components": "~4.9.6",
     "@octokit/rest": "^18.10.0",
     "@sentry/browser": "^6.13.1",
     "@testing-library/cypress": "^8.0.2",
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "@babel/preset-react": "^7.14.5",
     "@babel/register": "^7.15.3",
     "@department-of-veterans-affairs/eslint-plugin": "^1.1.0",
+    "@department-of-veterans-affairs/web-components": "^4.11.0",
     "@octokit/rest": "^18.10.0",
     "@sentry/browser": "^6.13.1",
     "@testing-library/cypress": "^8.0.2",
--- a/src/site/includes/header.html
+++ b/src/site/includes/header.html
@@ -42,7 +42,6 @@
 
   <!-- Add web components. -->
   <link rel="stylesheet" data-entry-name="web-components.css">
-  <script nonce="**CSP_NONCE**" defer data-entry-name="web-components.js"></script>
 
   <!-- Render GA template. -->
   {% include 'src/site/includes/google-analytics.liquid' %}
--- a/src/site/includes/header.html
+++ b/src/site/includes/header.html
@@ -42,6 +42,7 @@
 
   <!-- Add web components. -->
   <link rel="stylesheet" data-entry-name="web-components.css">
+  <script nonce="**CSP_NONCE**" defer data-entry-name="web-components.js"></script>
 
   <!-- Render GA template. -->
   {% include 'src/site/includes/google-analytics.liquid' %}
--- a/src/site/tests/home/custom-elements.cypress.spec.js
+++ b/src/site/tests/home/custom-elements.cypress.spec.js
@@ -1,0 +1,15 @@
+const webComponentDocs = require('@department-of-veterans-affairs/web-components/component-docs.json');
+
+const webComponentTags = webComponentDocs.components.map(comp => comp.tag);
+
+describe('Web Components', () => {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -90,6 +90,7 @@
     "@babel/preset-react": "^7.14.5",
     "@babel/register": "^7.15.3",
     "@department-of-veterans-affairs/eslint-plugin": "^1.1.0",
+    "@department-of-veterans-affairs/web-components": "~4.9.6",
```
bkjohnson, line 93: If this were to get ahead of the version we have in `vets-website` to the point where it has a new web component, then the test would fail. We can confirm this with a quick check.

At the moment [`vets-website` is using version `4.9.6` for the web-components package](https://github.com/department-of-veterans-affairs/vets-website/blob/e700daa10ab06a98c24cfa816fab7da82d88aa7b/yarn.lock#L2514). If I make these local changes here in `content-build`:

```diff
diff --git a/package.json b/package.json
index 53ae8731d..dd589be38 100644
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "@babel/preset-react": "^7.14.5",
     "@babel/register": "^7.15.3",
     "@department-of-veterans-affairs/eslint-plugin": "^1.1.0",
-    "@department-of-veterans-affairs/web-components": "~4.9.6",
+    "@department-of-veterans-affairs/web-components": "^4.9.6",
     "@octokit/rest": "^18.10.0",
     "@sentry/browser": "^6.13.1",
     "@testing-library/cypress": "^8.0.2",
diff --git a/yarn.lock b/yarn.lock
index 0fd28b547..03c1617ad 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -1088,10 +1088,10 @@
   dependencies:
     minimist "^1.2.6"
 
-"@department-of-veterans-affairs/web-components@~4.9.6":
-  version "4.9.6"
-  resolved "https://registry.yarnpkg.com/@department-of-veterans-affairs/web-components/-/web-components-4.9.6.tgz#16e9f057c0dbef33aa4342fc4d6ea94cdfd07c41"
-  integrity sha512-VJQq4VY8OQ0ZD32BacCjnnBeHyrJ9OoRmlBKZ7bR4LTLQwUWBDNYgSnLh8LQMO1h2gLnXdeeuMOz8xKEqT/Jig==
+"@department-of-veterans-affairs/web-components@^4.9.6":
+  version "4.12.0"
+  resolved "https://registry.yarnpkg.com/@department-of-veterans-affairs/web-components/-/web-components-4.12.0.tgz#8ff834053609739de8ecadf36df9788a3bd93fc1"
+  integrity sha512-e3QQaSwSimA5E0jzUZrTOTKi6ckFNxZOhR+DU9ztlprMZYcfOhvQo/ml6VbNA1gDrlj2d0UtYmVfo9kGwyIerw==
   dependencies:
     "@stencil/core" "^2.10.0"
     aria-hidden "^1.1.3"
```

And run the Cypress test, we get an error that `va-file-input` isn't defined.

Screenshot of the Cypress test runner. We see passing assertions for components like va-accordion being defined, but the test fails when it expects va-file-input to be defined.

We see the failure because this web component was [added in version `4.12.0` of the `web-comopnents` package](https://github.com/department-of-veterans-affairs/component-library/commit/2160809484ccad2aedc063a283e87bd3167915a0#diff-7298512f791c54d2827facfc8d26ae8a5db2ac960e464e30953ccaab05dd412eR3) so it will be in the `component-docs.json` that we are using. But `vets-website` isn't using this version yet, so when we pull in that build the component won't be defined.

---

Additionally, if this were to fall _behind_ a major version release (in the event that a web component was removed) relative to `vets-website`, I expect that this test would fail for the same reason. Using this dependency so that we can easily get a list of web component tags rather than hard-coding a tag name could mean a bit of chore work.

bkjohnson (commented):
If folks would rather we not add the dev dependency and instead have a test like this:

```js
describe('Web Components', () => {
  it('browser has VADS web components defined in its registry', () => {
    cy.visit('/');
    cy.window().then(window => {
      // This will be the constructor the browser uses when it sees a
      // Design System alert web component in the document.
      const alertElement = window.customElements.get('va-alert');

      assert.isFunction(customElement, 'va-alert is defined');
    });
  });
});
```

feel free to let me know. I figured that the dev dependency would make things a bit more obvious for someone who might be new and not know about design system web components, even though it might add a bit of chore work.

k80bowman (approved):

timwright12 (approved):

PR #1274, opened 2022-08-22
VACMS-10166 Add lovell menu bifurcation

wullaski:
## Description
Bifurcates menus for Lovell pages.

## Testing done

Visual

## Screenshots
Screen Shot 2022-08-22 at 5 09 08 PM

Screen Shot 2022-08-22 at 5 08 48 PM

## Acceptance criteria
- [ ] Remove Lovell federal health care from the array of sidebar menus to build.
   - [ ] Modify menu query to get Lovell federal health care menu with the additional field of field_menu_section 
- [ ] When on pages that are identified Lovell VA (both Lovell Federal Health Care and Lovell VA) they include a menu that only includes menu items labelled both or Lovell VA.
- [ ] When on pages that are identified Lovell TRICARE (both Lovell Federal Health Care and Lovell TRICARE) they include a menu that only includes menu items labelled both or Lovell TRICARE.

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/metalsmith-drupal.js
+++ b/src/site/stages/build/drupal/metalsmith-drupal.js
@@ -356,24 +356,15 @@ function lovellPageModify(page, variant) {
   return page;
 }
 
-function appendDrupalDataWithLovellPages(
-  drupalData,
-  lovellClonePages,
-  variant,
-) {
+function cloneLovellPages(lovellClonePages, variant) {
   // Deep clone with lodash
   const clonedPages = cloneDeep(lovellClonePages);
 
   // Modify the clones
-  const modifiedLovellPages = clonedPages.map(page => {
+  return clonedPages.map(page => {
     page = lovellPageModify(page, variant);
     return page;
   });
-
-  // Add the cloned pages to the drupal data
-  drupalData.data.nodeQuery.entities.push(...modifiedLovellPages);
-
-  return drupalData;
 }
 
 function lovellMenusModifyLinks(links, variation) {
@@ -493,19 +484,18 @@ function getDrupalContent(buildOptions) {
         },
       );
 
-      // Deep clone the lovell pages to create tricare
-      drupalData = appendDrupalDataWithLovellPages(
-        drupalData,
-        lovellVaClonePages,
-        'va',
-      );
-
       // Deep clone the lovell pages to create va pages
-      drupalData = appendDrupalDataWithLovellPages(
-        drupalData,
+      const vaClones = cloneLovellPages(lovellVaClonePages, 'va');
+      // Add the cloned pages to the drupal data
+      drupalData.data.nodeQuery.entities.push(...vaClones);
+
+      // Deep clone the lovell pages to create tricare pages
+      const tricareClones = cloneLovellPages(
         lovellTricareClonePages,
         'tricare',
       );
+      // Add the cloned pages to the drupal data
+      drupalData.data.nodeQuery.entities.push(...tricareClones);
 
       // Delete originals
       lovellClonePageIndexes.forEach(index => {
--- a/src/site/stages/build/drupal/metalsmith-drupal.js
+++ b/src/site/stages/build/drupal/metalsmith-drupal.js
@@ -307,22 +307,18 @@ async function loadCachedDrupalFiles(buildOptions, files) {
   }
 }
 
-function isLovellClonePage(page) {
-  // Pages that should be cloned have the value of 347
-  if (page.fieldAdministration) {
-    return page.fieldAdministration.entity.entityId === '347';
-  }
-  return false;
-}
+// Pages that should be cloned have the value of 347
+const isLovellClonePage = page =>
+  page.fieldAdministration &&
+  page.fieldAdministration.entity.entityId === '347';
 
 function lovellPageModify(page, variant) {
-  const linkMod = variant;
   const fieldOfficeMod = variant === 'tricare' ? 'Tricare' : 'VA';
 
   // Modify the path
   page.entityUrl.path = page.entityUrl.path.replace(
     '/lovell-federal-health-care',
-    `/lovell-federal-${linkMod}-health-care`,
+    `/lovell-federal-${variant}-health-care`,
   );
 
   // Modify the title used for querying the menus
@@ -350,7 +346,12 @@ function lovellPageModify(page, variant) {
   return page;
 }
 
-function appendDrupalDataWithLovellTricarePages(drupalData, lovellClonePages) {
+function appendDrupalDataWithLovellTricarePages(drupalData) {
+  // Get lovell 'clone' pages
+  const lovellClonePages = drupalData.data.nodeQuery.entities.filter(
+    isLovellClonePage,
+  );
+
   // Deep clone with lodash
   const clonedPages = cloneDeep(lovellClonePages);
 
@@ -463,16 +464,8 @@ function getDrupalContent(buildOptions) {
 
       await loadCachedDrupalFiles(buildOptions, files);
 
-      // Get lovell 'clone' pages
-      const lovellClonePages = drupalData.data.nodeQuery.entities.filter(
-        isLovellClonePage,
-      );
-
       // clone and modify pages
-      drupalData = appendDrupalDataWithLovellTricarePages(
-        drupalData,
-        lovellClonePages,
-      );
+      drupalData = appendDrupalDataWithLovellTricarePages(drupalData);
 
       // clone and modify menu
       drupalData = appendDrupalDataWithLovellTricareMenus(drupalData);
--- a/src/site/stages/build/drupal/metalsmith-drupal.js
+++ b/src/site/stages/build/drupal/metalsmith-drupal.js
@@ -324,6 +324,19 @@ function isLovellVaPage(page) {
   );
 }
 
+function getLovellClonePageIndexes(drupalData) {
+  return drupalData.data.nodeQuery.entities.map((page, index) => {
+    if (
+      page?.fieldAdministration?.entity?.entityId === '347' ||
+      page?.fieldAdministration?.entity?.entityId === '1039' ||
+      page?.fieldAdministration?.entity?.entityId === '1040'
+    ) {
+      return index;
+    }
+    return false;
+  });
+}
+
 function lovellPageModify(page, variant) {
   const fieldOfficeMod = variant === 'tricare' ? 'Tricare' : 'VA';
 
@@ -367,6 +380,24 @@ function cloneLovellPages(lovellClonePages, variant) {
   });
 }
 
+function getLovellClonePages(drupalData) {
+  // Get lovell 'clone' pages
+  const lovellTricareClonePages = drupalData.data.nodeQuery.entities.filter(
+    isLovellTricarePage,
+  );
+  const lovellVaClonePages = drupalData.data.nodeQuery.entities.filter(
+    isLovellVaPage,
+  );
+
+  // Deep clone the lovell pages to create va pages
+  const vaClones = cloneLovellPages(lovellVaClonePages, 'va');
+
+  // Deep clone the lovell pages to create tricare pages
+  const tricareClones = cloneLovellPages(lovellTricareClonePages, 'tricare');
+
+  return [...vaClones, ...tricareClones];
+}
+
 function lovellMenusModifyLinks(links, variation) {
   const titleVar = variation === 'va' ? 'VA' : 'Tricare';
   const linkVar = variation === 'va' ? 'va' : 'tricare';
@@ -406,10 +437,7 @@ function lovellMenusModifyLinks(links, variation) {
   });
 }
 
-function appendDrupalDataWithLovellMenus(drupalData) {
-  // Get the lovell menu key name
-  const lovellMenuKey = 'lovellFederalHealthCareFacilitySidebarQuery';
-
+function getLovellCloneMenus(drupalData, lovellMenuKey) {
   // Clone the original menu
   const lovellTricareMenu = cloneDeep(drupalData.data[lovellMenuKey]);
   const lovellVaMenu = cloneDeep(drupalData.data[lovellMenuKey]);
@@ -437,16 +465,10 @@ function appendDrupalDataWithLovellMenus(drupalData) {
   );
 
   // Add the cloned menus to the drupal data
-  drupalData.data = {
-    ...drupalData.data,
+  return {
     [lovellTricareMenuKey]: lovellTricareMenu,
     [lovellVaMenuKey]: lovellVaMenu,
   };
-
-  // Remove the original menu
-  delete drupalData.data[lovellMenuKey];
-
-  return drupalData;
 }
 
 function getDrupalContent(buildOptions) {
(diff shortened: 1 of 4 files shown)


src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -351,8 +463,19 @@ function getDrupalContent(buildOptions) {
 
       await loadCachedDrupalFiles(buildOptions, files);
 
-      // clone and modify for lovell tricare pages
-      drupalData = appendDrupalDataWithLovellTricarePages(drupalData);
+      // Get lovell 'clone' pages
+      const lovellClonePages = drupalData.data.nodeQuery.entities.filter(
```
maxx1128, line 467: **Refactor:** this variable is only used for `appendDrupalDataWithLovellTricarePages`, so it can be moved to that function and it'll only need to take `drupalData` as an argument.

maxx1128 (changes requested):
I have a couple suggestions for refactoring, but otherwise this looks good to me 👍

maxx1128 (approved):
🎉

allisonlu (approved):

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
-  const isLovellPage = page => {
-    if (page.fieldAdministration) {
-      return page.fieldAdministration.entity.entityId === '347';
-    }
-    return false;
-  };
+// Pages that should be cloned have the value of 347
+const isLovellClonePage = page =>
+  page.fieldAdministration &&
+  page.fieldAdministration.entity.entityId === '347';
```
ryguyk, line 313: Replace with optional chaining? This would remove the need for the check on line 312.

`page?.fieldAdministration?.entity?.entityId === '347`

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
 
-function appendDrupalDataWithLovellTricarePages(drupalData) {
-  const isLovellPage = page => {
-    if (page.fieldAdministration) {
-      return page.fieldAdministration.entity.entityId === '347';
-    }
-    return false;
-  };
+// Pages that should be cloned have the value of 347
+const isLovellClonePage = page =>
```
ryguyk, line 311: It seems a bit out of place to have a function expression here when all other functions are function declarations. I prefer function expressions, personally, but it feels out of place in this file.

ryguyk (commented):
I left a couple comments about some small clean up.

src/site/stages/build/drupal/metalsmith-drupal.js
ryguyk: Is it sufficient to check only this field to determine if the page should be cloned? I don't know the answer, to be clear. I'm just wondering if there might be (now or in the future) pages with this administration-field value that should not be cloned.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
+  page.fieldAdministration &&
+  page.fieldAdministration.entity.entityId === '347';
 
-  // Adjust this if needed to get the fieldAdministration field
-  const lovellPages = drupalData.data.nodeQuery.entities.filter(isLovellPage);
+function lovellPageModify(page, variant) {
```
ryguyk, line 315: I don't love that this function modifies one its parameters. In the name of functional programming principles, I'd like to see us all strive for pure functions wherever possible.

In light of how this is being called in the `forEach` below, though, changing this to be more pure would likely require overarching changes to how the page-data array is being modified.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
+  );
+
+  // Deep clone with lodash
+  const clonedPages = cloneDeep(lovellClonePages);
+
+  // Modify the clones
+  const modifiedLovellPages = clonedPages.map(page => {
+    page = lovellPageModify(page, 'tricare');
+    return page;
   });
```
ryguyk, line 362: This can likely be simplified to `page => lovellPageModify(page, 'tricare')`.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
 
+  // Modify the original pages
+  lovellClonePages.forEach(page => {
+    lovellPageModify(page, 'va');
+  });
```
ryguyk, line 367: This commentary relates to the discussion of pure functions and, more generally, functional programming at large. What's happening here can be quite tricky and confusing, and I'd like to strive for changing this.  Here's my take on the confusion:

- `drupalData.data.nodeQuery.entities` is an array of all page data
- `lovellClonePages` is a subset _separate array_ that is a subset of the array above. It's _pointing_ at objects in the array above, though. 
- So, in this `foreach` call, we are actually indirectly modifying `drupalData.data.nodeQuery.entities` and that is not immediately obvious (and likely very ripe for bugs).

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
+  }
 
-    // Modify the title
-    lovellPage.title = lovellPage.title.replace('Federal', 'Federal TRICARE');
+  // Modify the title this will become more complex to handle specific cases
+  if (variant === 'tricare') {
+    page.title = page.title.replace('Federal', 'Federal Tricare');
+  }
+  if (variant === 'va') {
+    page.title = page.title.replace('Federal', 'Federal VA');
```
swirtSJW, line 343: If it works, we might want to consider a more long tail search "Lovell Federal " (note the trailing space) to avoid the risk of "federal" being in the title for some other reason.  I see a bit of risk of something getting caught in the crossfire.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
 }
 
+function lovellMenusModifyLinks(links, variation) {
+  const titleVar = variation === 'va' ? 'VA' : 'Tricare';
+  const linkVar = variation === 'va' ? 'va' : 'tricare';
+
+  links.forEach(link => {
+    // If items are in both menus modify labels and links
+    if (link.entity.fieldMenuSection === 'both') {
+      link.label = link.label.replace('Federal', `Federal ${titleVar}`);
```
swirtSJW, line 382: Similar consideration for if a more specific search string would reduce the risk of crossfire.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
+      );
+    }
+
+    // Use recursion to modify nested links
+    if (link && link.links.length > 0) {
+      // Remove the links that don't belong in either the va or tricare menus
+      link.links = link.links.filter(menuItem => {
+        if (
+          menuItem.entity.fieldMenuSection === 'va' &&
+          variation === 'tricare'
```
swirtSJW, line 396: I feel like we are repeating a lot of the same strings.  Should some of these be set up as constants to make it more maintainable?

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,38 +307,149 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
+  // Get the lovell menu key name
+  const lovellMenuKey = 'lovellFederalHealthCareFacilitySidebarQuery';
+
+  // Clone the original menu
+  const lovellTricareMenu = cloneDeep(drupalData.data[lovellMenuKey]);
+  const lovellVaMenu = cloneDeep(drupalData.data[lovellMenuKey]);
+
+  // Rename the name so our new cloned pages can find the cloned menus
+  lovellTricareMenu.name = lovellTricareMenu.name.replace(
+    'Federal',
```
swirtSJW, line 423: crossfire risk?

swirtSJW (commented):
@wullaski This is looking really good.  
I have a couple concerns  related to our search strings being too small to avoid crossfire and wonder if we should stabilize some of the string re-use with some constants.

src/site/stages/build/drupal/metalsmith-drupal.js
swirtSJW: I think I am following what you are saying @ryguyk but I want to make sure.  Your objection is to altering the variable "page" by reference as opposed to a pure function that would return a new page and need to be stuck back into the original pagedata?

swirtSJW: nevermind... I just read your other comment down farther. ;)

src/site/stages/build/drupal/metalsmith-drupal.js
swirtSJW: > but I think that could be made more clear without implicit mutations.

By implicit are you meaning consistent naming, like any thing that is a modifier begin with the verb "modify" or are something more than that?

swirtSJW: Sorry I just realize I completely misread that line.    Disregard my last comment.

src/site/stages/build/drupal/metalsmith-drupal.js
swirtSJW: No if they have that value specifically, they are must be cloned.

src/site/stages/build/drupal/metalsmith-drupal.js
ryguyk: These two fiddles might help clarify my general point:

1. Not so good: https://jsfiddle.net/rh7nepy8/2/
2. Better solution: https://jsfiddle.net/nkem4vgw/2/

ryguyk: And here's a fiddle that roughly captures what I see as a more clear/immutable approach: https://jsfiddle.net/wszpa3qo/2/

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -351,8 +485,25 @@ function getDrupalContent(buildOptions) {
...
-      drupalData = appendDrupalDataWithLovellTricarePages(drupalData);
+      // clone and modify pages for Lovell
+      const lovellClonePageIndexes = getLovellClonePageIndexes(drupalData);
+      drupalData.data.nodeQuery.entities = [
+        ...drupalData.data.nodeQuery.entities,
+        ...getLovellClonePages(drupalData),
+      ];
+      // Delete original pages for lovell
+      lovellClonePageIndexes.forEach(index => {
+        delete drupalData.data.nodeQuery.entities[index];
```
ryguyk, line 496: I'm not sure `delete` is the best option, as this will leave a hole in the existing data. This _might_ be fine, but it feels odd.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -305,36 +309,166 @@ async function loadCachedDrupalFiles(buildOptions, files) {
...
+      page?.fieldAdministration?.entity?.entityId === '347' ||
+      page?.fieldAdministration?.entity?.entityId === '1039' ||
+      page?.fieldAdministration?.entity?.entityId === '1040'
+    ) {
+      return index;
     }
     return false;
-  };
+  });
+}
```
ryguyk, line 338: I think `filter` _combined_ with `map` would be better here. The current solution will return an array with a _lot_ of `false` values that are unnecessary. All you really want, in the end, is a list of index values that represent the Lovell pages. If it's not a Lovell page, you just don't want to worry about it. We can filter those out.

Beyond this, the manner in which this is being used (`delete data[index]`) doesn't seem to make a lot of sense with a `false` value passed in. It might work, but it's confusing and unnecessary. Please see note around use of the `delete` as well. I'm not sure this is the best solution.

src/site/stages/build/drupal/metalsmith-drupal.js
```diff
@@ -351,8 +485,25 @@ function getDrupalContent(buildOptions) {
...
+      });
+
+      // clone and modify menus for Lovell
+      const lovellMenuKey = 'lovellFederalHealthCareFacilitySidebarQuery';
+      drupalData.data = {
+        ...drupalData.data,
+        ...getLovellCloneMenus(drupalData, lovellMenuKey),
+      };
+      // Remove the original menu
+      delete drupalData.data[lovellMenuKey];
```
ryguyk, line 506: Could all this Lovell-specific work be pulled out into a wrapper function? It feels like this function is pretty cleanly laid out with high-level steps that constitute the work of getting Drupal content. This Lovell code feels like it makes this harder to read.

Ex:
```
      await loadCachedDrupalFiles(buildOptions, files);
      processLovellPages(drupalData);
      pipeDrupalPagesIntoMetalsmith(drupalData, files);
```

src/site/stages/build/drupal/metalsmith-drupal.js
ryguyk: My preference would still be for constructing a new array of pages by building up from scratch and including the processed pages as you process them. This would remove the need for anything like `delete`.

Ex:
```
const {
    lovellFederalPages,
    lovellVaPages,
    lovellTricarePages,
    otherPages,
  } = drupalData.data.nodeQuery.entities.reduce(
    (acc, page) => {
        acc.lovellFederalPages.push(page);
      } else if (isLovellTricarePage(page)) {
        acc.lovellTricarePages.push(page);
      } else if (isLovellVaPage(page)) {
        acc.lovellVaPages.push(page);
      } else {
        acc.otherPages.push(page);
      }

      return acc;
    },
    {
      lovellFederalPages: [],
      lovellTricarePages: [],
      lovellVaPages: [],
      otherPages: [],
    },
  );

  // modify tricare pages
  const modifiedLovellTricarePages = lovellTricarePages.map(page =>
    getModifiedLovellPage(page, 'tricare'),
  );
  // modify va pages
  const modifiedLovellVaPages = lovellVaPages.map(page =>
    getModifiedLovellPage(page, 'va'),
  );
  // Each federal page needs to be duplicated and modified once each for tricare/va
  const lovellFederalPagesClonedTricare = lovellFederalPages.map(page =>
    getModifiedLovellPage(cloneDeep(page), 'tricare'),
  );
  const lovellFederalPagesClonedVa = lovellFederalPages.map(page =>
    getModifiedLovellPage(cloneDeep(page), 'va'),
  );

  drupalData.data.nodeQuery.entities = [
    ...modifiedLovellTricarePages,
    ...modifiedLovellVaPages,
    ...lovellFederalPagesClonedTricare,
    ...lovellFederalPagesClonedVa,
    ...otherPages,
  ];
```

This would require a few additional changes:
1. `getModifiedLovellPage` would be a function just like the existing `lovellPageModify`, but would return a value (the modified page).
2. Page checks for tricare/va would only check for their individual entity id, then a new function `isLovellFederalPage(page)` would check for `entityId = LOVELL_FEDERAL_ENTITY_ID`
```
function isLovellFederalPage(page) {
  return (
    page?.fieldAdministration?.entity?.entityId === LOVELL_FEDERAL_ENTITY_ID
  );
}
function isLovellTricarePage(page) {
  return (
    page?.fieldAdministration?.entity?.entityId === LOVELL_TRICARE_ENTITY_ID
  );
}
function isLovellVaPage(page) {
  return page?.fieldAdministration?.entity?.entityId === LOVELL_VA_ENTITY_ID;
}
```

ryguyk (commented):
Added some commentary. Let's chat if it's helpful.

src/site/stages/build/drupal/metalsmith-drupal.js
wullaski: Yes for sure! Max and I were both thinking that this should happen.

maxx1128: 100% agree here

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -0,0 +1,186 @@
...
+    LOVELL_TITLE_STRING,
+    `${LOVELL_TITLE_STRING} ${fieldOfficeMod}`,
+  );
+
+  return page;
+}
+
+function lovellMenusModifyLinks(link) {
+  const { variant } = this;
+  const titleVar = variant === 'va' ? 'VA' : 'Tricare';
```
swirtSJW, line 62: This has been finalized to be all caps.
```suggestion
  const titleVar = variant === 'va' ? 'VA' : 'TRICARE';
```

swirtSJW (changes requested):
This is all a lot more readable and consistent.  Thank you for taking on all the refactoring.
Just made one little minor suggestion.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -0,0 +1,186 @@
...
+    ...drupalData.data,
+    ...getLovellCloneMenu(drupalData, LOVELL_MENU_KEY, 'va'),
+    ...getLovellCloneMenu(drupalData, LOVELL_MENU_KEY, 'tricare'),
+  };
+  // Remove the original menu
+  delete drupalData.data[LOVELL_MENU_KEY];
+}
+
+module.exports = {
+  processLovellPages,
```
maxx1128, line 185: 🔥 for this entire page

src/site/stages/build/drupal/process-lovell-pages.js
wullaski: @swirtSJW I made this change and pushed up another commit.

PR #1308, opened 2022-09-21
VACMS-10168 Add Lovell categorized event listings

maxx1128:
## Description

The event listing pages on Lovell Tricare and VA only have federal events, but not events specific to Tricare or the VA. This takes the Tricare or VA specific events and adds them to their respective listings.

UPDATE on 10/12: Thanks to @ryguyk for helping me get this ticket on the right track. This PR now updates the listings for events, press releases, and stories. Screenshots have been updated. The broken menus are a separate issue and will be addressed later.

## Testing done

Visual, see below screenshots

## Screenshots

Screen Shot 2022-10-12 at 12 05 56 PM

Screen Shot 2022-10-12 at 12 06 05 PM

Screen Shot 2022-10-12 at 12 06 16 PM

Screen Shot 2022-10-12 at 12 06 23 PM

Screen Shot 2022-10-12 at 12 06 41 PM

Screen Shot 2022-10-12 at 12 06 32 PM

## Acceptance criteria
- [ ] The Lovell Tricare event listing page has federal and Tricare events, but not VA events
- [ ] The Lovell VA event listing page has federal and VA events, but not Tricare events
- [ ] The event listings still sort and filter events as expected

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -29,7 +29,7 @@ function isLovellVaPage(page) {
   return page?.fieldAdministration?.entity?.entityId === LOVELL_VA_ENTITY_ID;
 }
 
-function isLovellListingPage(page) {
+function isListingPage(page) {
   const listingPageTypes = [
     'event_listing',
     'press_releases_listing',
@@ -351,19 +351,19 @@ function processLovellPages(drupalData) {
   } = drupalData.data.nodeQuery.entities.reduce(
     (acc, page) => {
       if (isLovellFederalPage(page)) {
-        if (isLovellListingPage(page)) {
+        if (isListingPage(page)) {
           acc.lovellFederalListingPages.push(page);
         } else {
           acc.lovellFederalNonListingPages.push(page);
         }
       } else if (isLovellTricarePage(page)) {
-        if (isLovellListingPage(page)) {
+        if (isListingPage(page)) {
           acc.lovellTricareListingPages.push(page);
         } else {
           acc.lovellTricareNonListingPages.push(page);
         }
       } else if (isLovellVaPage(page)) {
-        if (isLovellListingPage(page)) {
+        if (isListingPage(page)) {
           acc.lovellVaListingPages.push(page);
         } else {
           acc.lovellVaNonListingPages.push(page);
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 /* eslint-disable no-param-reassign, no-console */
 const cloneDeep = require('lodash/cloneDeep');
 
@@ -282,18 +283,12 @@ function updateLovellSwitchLinks(page, pages) {
  */
 function combineLovellListingPages(tricareOrVaPages, federalPages) {
   return tricareOrVaPages.map(listingPage => {
-    let pastObjectLabel;
-    switch (listingPage.entityBundle) {
-      case 'press_release_listing':
-        pastObjectLabel = 'pastPressReleases';
-        break;
-      case 'story_listing':
-        pastObjectLabel = 'pastNewsStories';
-        break;
-      default:
-        pastObjectLabel = 'pastEvents';
-        break;
-    }
+    const typePastMap = {
+      event_listing: 'pastEvents',
+      press_releases_listing: 'pastPressReleases',
+      story_listing: 'pastNewsStories',
+    };
+    const pastObjectLabel = typePastMap[listingPage.entityBundle];
     const {
       entityBundle,
       [pastObjectLabel]: pastListItems,
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -294,17 +294,23 @@ function combineLovellListingPages(tricareOrVaPages, federalPages) {
       [pastObjectLabel]: pastListItems,
       reverseFieldListingNode,
     } = listingPage;
+
     const listingPageToCombine = federalPages.find(
       page => page.entityBundle === entityBundle,
     );
+
+    const {
+      [pastObjectLabel]: pastListItemsToCombine,
+      reverseFieldListingNode: reverseFieldListingNodeToCombine,
+    } = listingPageToCombine;
 
     const pastListItemEntities = pastListItems?.entities || [];
     const allListItemEntities = reverseFieldListingNode?.entities || [];
 
     const pastListItemEntitiesToCombine =
-      listingPageToCombine[pastObjectLabel]?.entities || [];
+      pastListItemsToCombine?.entities || [];
     const allListItemEntitiesToCombine =
-      listingPageToCombine?.reverseFieldListingNode?.entities || [];
+      reverseFieldListingNodeToCombine?.entities || [];
 
     const combinedPastListItemEntities = [
       ...pastListItemEntities,
@@ -315,20 +321,17 @@ function combineLovellListingPages(tricareOrVaPages, federalPages) {
       ...allListItemEntitiesToCombine,
     ];
 
-    const finalListingPage = {
+    return {
       ...listingPage,
       reverseFieldListingNode: {
         ...reverseFieldListingNode,
         entities: combinedAllListItemEntities,
       },
+      [pastObjectLabel]: {
+        ...pastListItems,
+        entities: combinedPastListItemEntities,
+      },
     };
-
-    finalListingPage[pastObjectLabel] = {
-      ...pastListItems,
-      entities: combinedPastListItemEntities,
-    };
-
-    return finalListingPage;
   });
 }
 
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -64,14 +64,32 @@ function getModifiedLovellPage(page, variant) {
 
   // Add a switchPath field
   // These get modified later in the processLovellPages function
-  page.entityUrl.switchPath = originalPath.replace(
-    '/lovell-federal-health-care',
-    `/lovell-federal-${
+  if (
+    page.entityUrl.path.includes(
+      `/lovell-federal-${LOVELL_TRICARE_LINK_VARIATION}-health-care`,
+    ) ||
+    page.entityUrl.path.includes(
+      `/lovell-federal-${LOVELL_VA_LINK_VARIATION}-health-care`,
+    )
+  ) {
+    page.entityUrl.switchPath = originalPath.replace(
+      variant === 'va'
+        ? LOVELL_VA_LINK_VARIATION
+        : LOVELL_TRICARE_LINK_VARIATION,
       variant === 'va'
         ? LOVELL_TRICARE_LINK_VARIATION
-        : LOVELL_VA_LINK_VARIATION
-    }-health-care`,
-  );
+        : LOVELL_VA_LINK_VARIATION,
+    );
+  } else {
+    page.entityUrl.switchPath = originalPath.replace(
+      '/lovell-federal-health-care',
+      `/lovell-federal-${
+        variant === 'va'
+          ? LOVELL_TRICARE_LINK_VARIATION
+          : LOVELL_VA_LINK_VARIATION
+      }-health-care`,
+    );
+  }
 
   // Modify Breadcrumb
   if (page.entityUrl.breadcrumb) {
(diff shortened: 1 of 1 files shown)


swirtSJW:
@maxx1128 something doesn't look right in these screenshots.  There should be another event showing up on both 
image

It should be the one that is pulled from /lovell-federal-health-care/events   (node id 36319)  


Also this is only for events,  The https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10168 is for 4 types of list pages.  Wee need the logic to cover all 4

maxx1128:
I amended the commit so the federal event appears for both and still filters properly.

Screen Shot 2022-09-22 at 12 06 43 PM
Screen Shot 2022-09-22 at 12 06 50 PM
Screen Shot 2022-09-22 at 12 07 51 PM
Screen Shot 2022-09-22 at 12 07 58 PM

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -226,6 +226,65 @@ function updateLovellSwitchLinks(page, pages) {
...
+        eventPage.entityBundle === 'event' && isLovellVaPage(eventPage),
+    );
+
+    const addedEvents = page.entityUrl.path.includes(
+      'lovell-federal-tricare-health-care',
+    )
+      ? lovellTricareEvents
+      : lovellVaEvents;
+
+    const allPastEvents = [...page.pastEvents.entities, ...addedEvents]
```
swirtSJW, line 247: I am concerned that we are duplicating this process which already gets applied to event listing pages,  If we just get the data on the page, the normal processing that occurs during the page build should handle it.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -226,6 +226,65 @@ function updateLovellSwitchLinks(page, pages) {
...
+        eventPage.entityBundle === 'event' && isLovellTricarePage(eventPage),
+    );
+
+    const lovellVaEvents = pages.filter(
+      eventPage =>
+        eventPage.entityBundle === 'event' && isLovellVaPage(eventPage),
+    );
+
+    const addedEvents = page.entityUrl.path.includes(
+      'lovell-federal-tricare-health-care',
```
swirtSJW, line 242: Looking up the pages by path seems problematic / fragile.  We should be able to look them up by type and fieldAdminstration.  We might not have that filter build yet.   

This particular lookup is not needed since the processing of individual events is not needed.  However it would still be needed for grabbing the list pages.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -226,6 +226,65 @@ function updateLovellSwitchLinks(page, pages) {
   return page;
 }
 
+function updateEventPageListings(page, pages) {
```
swirtSJW, line 229: Can we make this less specific to event listing pages?  it should be something like updateListingPages
that iterates through the three types of listing pages.  They should each have the structure that I show a bit farther down in the comments.  Ideally there should be nothing specific to events in this.  It is just about copying and appending the reverseFieldListingNodes entities .

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -226,6 +226,65 @@ function updateLovellSwitchLinks(page, pages) {
...
+  if (page.entityBundle === 'event_listing') {
+    const lovellTricareEvents = pages.filter(
+      eventPage =>
+        eventPage.entityBundle === 'event' && isLovellTricarePage(eventPage),
```
swirtSJW, line 233: I don't think we need to process individual events.

Here we are kind of querying the events individually, but I don't think we have to. They exist correctly on the listing pages (same for stories and press releases)


The data for the events is already on the listing page.
image

You should just be able to grab it from the page and move it, then delete the original page.
image

This should all take place before any of the bifurcation logic makes copies of the "both" page.

swirtSJW (changes requested):
I think this pattern is more intensive than it needs to be and does not scale to the two other listing pages.  See my comments below for more details.

longmd (dismissed):
LGTM 🚀

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -243,6 +243,75 @@ function updateLovellSwitchLinks(page, pages) {
   return page;
 }
 
+function updateListingsForSection(page, federalPages) {
```
swirtSJW, line 246: Immediately after this is called for both VA and TRICARE, the "both" pages need to removed from the page data so they do not run through the bifurcation.

swirtSJW (changes requested):
Looks like you are on the right track but we have some timing / order of operations issues that need to be worked out.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -243,6 +243,75 @@ function updateLovellSwitchLinks(page, pages) {
...
+  // .filter((element, index, allEvents) => {
+  //   return allEvents.indexOf(element) !== index;
+  // });
+
+  // console.log({
+  //   futureEvents,
+  //   pastEvents,
+  // });
+
+  if (page.reverseFieldListingNode && futureEvents.length > 0) {
```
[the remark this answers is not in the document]
swirtSJW: If we could get your modification to happen in front of that  step, where those are processed for future and passed, it would ideal.  Not sure if it is possible.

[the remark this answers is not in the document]
maxx1128: This is just a check to make sure past and future events are only added to pages where they exist. But I can see the issue with this in hindsight - if there are no Tricare/VA events, but there are several federal events, then they won't be added together. I'll try removing it now.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -274,27 +343,28 @@ function processLovellPages(drupalData) {
   );
 
   // modify tricare pages
-  const modifiedLovellTricarePages = lovellTricarePages.map(page =>
-    getModifiedLovellPage(page, 'tricare'),
-  );
+  const modifiedLovellTricarePages = lovellTricarePages
+    .map(page => updateListingsForSection(page, lovellFederalPages))
```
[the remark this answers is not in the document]
maxx1128: The real bifurcation takes place with the `getModifiedLovellPage` functions. Before that, they're just being sorted into different groups for the bifurcation. So the listing updates are happening right before the bifurcation page updates unless I'm misunderstanding the full scope of the process.

src/site/stages/build/drupal/process-lovell-pages.js
maxx1128: Since the function is only run when making the VA and Tricare pages (and they're the only ones passed to the function), I don't believe the federal pages are actually affected here, just used to modify the VA/Tricare pages. The line that sets the Drupal data, `drupalData.data.nodeQuery.entities = [...processedLovellPages, ...otherPages]`, shows the Va/Tricare pages are still the only ones being added back to the main Drupal data so the "both" pages aren't used.

src/site/stages/build/drupal/process-lovell-pages.js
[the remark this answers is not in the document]
maxx1128: Tried removing it but I went back to getting this error, the page went back to throwing this error. Going to add the checks back.

```javascript
Drupal successfully loaded!
TypeError: Cannot set property 'entities' of undefined
    at mergeListItemsFromLovellBoth (/Users/maxwellantonucci/Desktop/Repos/content-build/src/site/stages/build/drupal/process-lovell-pages.js:297:33)
    at /Users/maxwellantonucci/Desktop/Repos/content-build/src/site/stages/build/drupal/process-lovell-pages.js:349:18
    at Array.map (<anonymous>)
    at processLovellPages (/Users/maxwellantonucci/Desktop/Repos/content-build/src/site/stages/build/drupal/process-lovell-pages.js:349:6)
    at Ware.<anonymous> (/Users/maxwellantonucci/Desktop/Repos/content-build/src/site/stages/build/drupal/metalsmith-drupal.js:357:7)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
Failed to pipe Drupal content into Metalsmith!

```

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -243,49 +268,150 @@ function updateLovellSwitchLinks(page, pages) {
...
 
+/**
+ * For each listing page in tricareOrVaPages, finds the first occurrence of that listing page type in
+ * federalPages and injects pastEvents and reverseFieldListingNode from that entity into
+ * corresponding properties on the listing page from tricareOrVaPages.
+ *
+ * @param {*} tricareOrVaPages Listing pages that will have pastEvents and reverseFieldListingNode properties merged from federalPages
+ * @param {*} federalPages Listing pages to merge into tricareOrVaPages
+ * @returns
+ */
```
swirtSJW, line 279: Thank you for the docblock

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -29,6 +29,16 @@ function isLovellVaPage(page) {
   return page?.fieldAdministration?.entity?.entityId === LOVELL_VA_ENTITY_ID;
 }
 
+function isLovellListingPage(page) {
```
swirtSJW, line 32: minor:  Reports doing something that it is not 'isLovellListingPage'  gives the impression it is checking for it being a lovell page, but is only checking for it being a listing page.  `isListingPage` would be more truthful.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -243,49 +268,150 @@ function updateLovellSwitchLinks(page, pages) {
...
+      case 'press_release_listing':
+        pastObjectLabel = 'pastPressReleases';
+        break;
+      case 'story_listing':
+        pastObjectLabel = 'pastNewsStories';
+        break;
+      default:
+        pastObjectLabel = 'pastEvents';
+        break;
+    }
```
swirtSJW, line 293: Did we need a catchall (default) for event?  If we don't really need the catchall it might be tighter to use the actual value as a map to avoid something unexpected
```suggestion
    const typePastMap = [
      'event_listing' => 'pastEvents',
      'press_releases_listing' => 'pastPressReleases',
      'story_listing' => 'pastNewsStories',
    ];
    pastObjectLabel = typePastMap[listingPage.entityBundle];   
```

swirtSJW (commented):
I left a couple comments/questions.  
Depending on the answer to the question about the catchall, it may need changes or not.

src/site/stages/build/drupal/process-lovell-pages.js
maxx1128: This looks like a good idea, going to try it out now and commit it if it works.

maxx1128:
@swirtSJW Updates have been made and I confirmed it still works for me locally, should be good for another (and hopefully final 🙏) round of review.

ryguyk (changes requested):
This is looking pretty good overall. I added a few comments and listed one big concern that I think should be changed.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -54,14 +65,29 @@ function getModifiedLovellPage(page, variant) {
 
   // Add a switchPath field
   // These get modified later in the processLovellPages function
-  page.entityUrl.switchPath = originalPath.replace(
-    '/lovell-federal-health-care',
-    `/lovell-federal-${
+  if (
+    !page.entityUrl.path.includes('/lovell-federal-health-care') &&
```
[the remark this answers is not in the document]
maxx1128: Good catch, this check isn't needed here.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -29,6 +30,16 @@ function isLovellVaPage(page) {
...
+  const listingPageTypes = [
+    'event_listing',
+    'press_releases_listing',
+    'story_listing',
+  ];
+
+  return listingPageTypes.includes(page.entityBundle);
+}
+
 function getModifiedLovellPage(page, variant) {
```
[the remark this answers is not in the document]
maxx1128: I agree that a lot of pieces in this function could be broken up into smaller functions, likely ones that just need a page, the variant, and maybe a few of the values defined at the top level. But this kind of refactor seems better suited to a follow-up ticket.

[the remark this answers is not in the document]
swirtSJW: Yah lets clean up this bit in a different round.  This is going to integration for now so we can do a final round of refactoring separately.

swirtSJW (approved):
Nice work @maxx1128 and @ryguyk   If Ryan approves it is go for merge to integration.
I will build our demo env from this branch so we should be good.

swirtSJW:
I am merging this based on conversation that we will followup on the integration branch with a round of refactoring and documentation.

PR #1316, opened 2022-09-30
10801 - tweaks to filter logic to function regardless of trailing slash

apisandipas:
## Description
This PR adjust the functionality of some Liquid filters to allow for passing of paths with or without trailing slashing, causing them to function the same way.

## Testing done
Manual tested completed and automated tests have been updated to support this work.

## Screenshots
n/a

## Acceptance criteria
- [x] path with trailing slash and path without trailing slash are handled the same, when including 
  - [x] `/test/` vs. `/test`
  - [x] `/test/*/` vs. `/test/*`
- [x] path with trailing slash and path without trailing slash are handled the same, when excluding
  - [x] `!/test/` vs. `!/test`
  - [x] `!/test/*/` vs. `!/test/*`
- [x] Add unit tests as needed

## Definition of done


maxx1128 (approved):

jtmst (approved):

ryguyk (changes requested):
I'm not sure this captures what we need. This is confusing because there are a lot of possible cases of trailing slashes and wildcards. But the most basic piece that we want to capture is that if an editor wants to show a banner on `/some/page`, the editor should be able to put `/some/page`  *OR* `/some/page/`. This is still not possible. 

I added these tests locally, and they both fail.
```
  it('TEST 1', () => {
    const targetPaths = ['/some/path/test'];
    const currentPath = '/some/path/test/';

    expect(liquid.filters.isBannerVisible(targetPaths, currentPath)).to.equal(
      true,
    );
  });

  it('TEST 2', () => {
    const targetPaths = ['/some/path/test'];
    const currentPath = '/some/path/test';

    expect(liquid.filters.isBannerVisible(targetPaths, currentPath)).to.equal(
      true,
    );
  });
```

I have some thoughts about how this might most easily be done. They boil down to needing to call `formatPath` on the `targetPaths`.

apisandipas:
@ryguyk I think I may be mixing up the meaning of currentPath v targetsPaths here and am seeking some clarification. I was under the impression based on the ticket that targetPaths was a list of paths that could possible end in a trailing slash but might not depending on how they were entered, by whom etc. and the currentPath be the current path of the page thats open in the browser. 

It sounds like from what you said that we want to optionally add the trailing slash to the browser url. Is that correct?

rmessina1010 (approved):
LGTM

src/site/filters/liquid.unit.spec.js
```diff
@@ -1386,6 +1386,24 @@ describe('formatPath', () => {
...
+
+  it('returns true without a trailing slash on currentPath', () => {
+    const targetPaths = ['/some/path/test'];
+    const currentPath = '/some/path/test';
+
+    expect(liquid.filters.isBannerVisible(targetPaths, currentPath)).to.equal(
+      true,
+    );
+  });
+
```
ryguyk, line 1406: I think my previous comment that included some sample tests might have been confusing.  I apologize.  Let me explain better.

This is what I wrote:
```
  it('TEST 1', () => {
    const targetPaths = ['/some/path/test'];
    const currentPath = '/some/path/test/';

    expect(liquid.filters.isBannerVisible(targetPaths, currentPath)).to.equal(
      true,
    );
  });

  it('TEST 2', () => {
    const targetPaths = ['/some/path/test'];
    const currentPath = '/some/path/test';

    expect(liquid.filters.isBannerVisible(targetPaths, currentPath)).to.equal(
      true,
    );
  });
```
The only difference between those two tests is the trailing slash on `currentPath`. However, that difference takes attention away from what should have been the real focus - a `targetPath` that lacks a trailing slash. A `targetPath` lacking a trailing slash is the very thing that brought this issue to life. An editor added a `targetPath` without a trailing slash, and the banner did not appear as the editor thought it would because the existing code did not accommodate an absent trailing slash on the target path. 

What I intended to capture with the two sample tests above is that a `targetPath` without a trailing slash would not work regardless of whether the `currentPath` included a trailing slash. Instead, it's clear that I drew attention to the idea that we need to test for the presence and absence of trailing slashes on the `currentPath`. It's not necessarily bad to do so, but the structure of the `currentPath` is more or less static. It's not going to change based on editor input. 

**We're still lacking a test for demonstrating that the same `targetPath` will work regardless of the presence or absence of a trailing slash.**

I apologize for possibly leading you astray.

ryguyk (changes requested):
See comment. The code changes look good. The tests still need a tweak. I think the confusion was likely created by me.

src/site/filters/liquid.unit.spec.js
apisandipas: You’ve confirmed my original understanding of how this worked, but I'm not positive I'm any clearer on the ask. Ive added tests that demonstrate that it works with and without a trailing slash in the target path, which make sense given the Regex that was added. Let me know if you need anything further here.

PR #1332, opened 2022-10-18
[VA-11155] Lovell Menus not showing on front-end

maxx1128:
## Description

Lovell pages aren't showing any of their menus. The cause is the Lovell menu structure was changed, and the Federal menu needs to be before the Tricare or VA menus. This fix rearranges the menu as part of the cloning process so the federal one is always placed first, which should help avoid similar bugs in the future.

## Testing done

Visual, see below screenshots.

## Screenshots

Screen Shot 2022-10-18 at 7 12 32 PM

## Acceptance criteria
- [ ] Lovell pages (federal, Tricare, and VA) all show the side menu.
- [ ] Lovell side menus all have working links

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -196,6 +196,14 @@ function getLovellCloneMenu(drupalData, lovellMenuKey, variant) {
     `${LOVELL_TITLE_STRING} ${titleVar}`,
   );
 
+  // Move federal health care links to the top of the menu
+  // Otherwise the menu renders as blank
+  const federalLinksIndex = lovellCloneMenu.links
+    .map(menu => menu.label)
+    .indexOf('Lovell Federal health care');
+  const federalLinks = lovellCloneMenu.links.splice(federalLinksIndex, 1);
+  lovellCloneMenu.links = [...federalLinks, ...lovellCloneMenu.links];
+
   // Change the root level item
   // It's coming in from the cms as a va item when it should be both
   lovellCloneMenu.links[0].label = 'Lovell Federal Health Care';


jtmst (approved):

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -196,6 +196,14 @@ function getLovellCloneMenu(drupalData, lovellMenuKey, variant) {
...
   );
 
+  // Move federal health care links to the top of the menu
+  // Otherwise the menu renders as blank
+  const federalLinksIndex = lovellCloneMenu.links
+    .map(menu => menu.label)
+    .indexOf('Lovell Federal health care');
+  const federalLinks = lovellCloneMenu.links.splice(federalLinksIndex, 1);
+  lovellCloneMenu.links = [...federalLinks, ...lovellCloneMenu.links];
+
```
ryguyk, line 206: I agree this does precisely what it says it does - moves federal links to the top. It does that very nice and neatly!

The question I'm asking myself, though, is _why_ do they need to be moved to the top?

ryguyk: It appears that the code just below this to "Change the root level item" _does_ kind of, sort of, maybe expect the Federal links to be first. It's no question to me that this wouldn't work without the change in this PR. However, it feels like we might be using duct tape to hold together a sinking house rather than just fixing the foundation.

ryguyk (commented):
I'm not sure about this. The change in this PR looks like some very solid code. If we can confirm that moving the Federal links to the top should be a legitimate step in this process - and not just a product of something else that previously introduced some problems - then I'd be very happy to see this code merged. I'm not confident that is the case though. This is very difficult to follow and I'm just not confident in what I'm seeing _outside_ this PR.

src/site/stages/build/drupal/process-lovell-pages.js
swirtSJW: It seems like there are two possible approaches
A:  Grab the index of lovell federall and move it to the top spot (what this PR is doing now)
B: Grab the index of lovell federall and use it down here to grab the right spot 
image

Approach B would save 2 lines of code and 2 lines of comments.  But either seem equally reliable.

Is that what you were thinking @ryguyk or something more foundational.

ryguyk (approved):
Discussion with the team led to an agreement that this PR is good to go. This does what it says it does, and does that very cleanly.

PR #1350, opened 2022-11-04
Outreach filteres added and sorted

jtmst:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11153

Adjusted graphql query to allow outreach filters to be added dynamically

## Testing done
Local testing with backend tugboat instance

## Screenshots
Screen Shot 2022-11-02 at 9 45 32 AM
Screen Shot 2022-11-04 at 1 14 48 PM

## Acceptance criteria
- [ ] The topic filter options are powered by field_lc_categories from the data items on the page (this is a new field and will need to be added to the graphql.)
- [ ] ~~The file type filter options are powered by field_format  (this is not a new field)~~
- [ ] Topic filter and type filter select lists are alpha sorted
- [ ] A veteran should never be able to select a topic option that has no data. (example: Disability should not appear as a filter if there are no data items of topic Disability.)
- [ ] The field field_benefits is no longer referenced on the template and no longer in the graphQL 
- [ ] Accessibility review - receives 508 approval 
- [ ] Create future issue -  The two filters are aware of each other and self restrict the other's limits.  (example: If I select format = video, I should not see types in the filter that contain no videos. This would prevent from selecting a filter combination that has no results.

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -109,6 +109,30 @@ module.exports = function registerFilters() {
 
   liquid.filters.formatDate = (dt, format) => prettyTimeFormatted(dt, format);
 
+  liquid.filters.buildTopicList = topics => {
+    if (!topics) return null;
+    const topicArray = [];
+
+    // Add unique topics to topic array
+    for (let i = 0; i < topics.length; i++) {
+      const fieldObjects = topics[i].fieldLcCategories;
+      if (fieldObjects[0]) {
+        const { entity } = fieldObjects[0];
+        const index = topicArray.findIndex(
+          object => object.name === entity.name,
+        );
+        if (index === -1) {
+          topicArray.push(entity);
+        }
+      }
+    }
+    return topicArray;
+  };
+
+  liquid.filters.alphabetizeList = items => {
+    return _.orderBy(items, [item => item?.name?.toLowerCase()], ['asc']);
+  };
+
   liquid.filters.drupalToVaPath = content => {
     let replaced = content;
     if (content) {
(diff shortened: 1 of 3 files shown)


src/site/filters/liquid.js
```diff
@@ -109,6 +109,30 @@ module.exports = function registerFilters() {
 
   liquid.filters.formatDate = (dt, format) => prettyTimeFormatted(dt, format);
 
+  liquid.filters.buildTopicList = topics => {
+    if (!topics) return null;
+    const topicArray = [];
+
```
apisandipas, line 115: For building up an array like this, I think it's a little more idiomatic and easier to read if you use `reduce` function. I'm interested to know @ryguyk's take on that though.

apisandipas (changes requested):
I had one concern/comment/question. I don't think it should be a blocker, but perhaps check with @ryguyk for his opinion.

apisandipas (approved):
LGTM, refactor looks good .

jilladams:
One flag here: 
https://prod.cms.va.gov/admin/content?title=PACT&type=outreach_asset&moderation_state=All&owner=All -- in Prod we now have a batch of Publications using the PACT act taxonomy, as of EOD yesterday 11/8

Verified that those nodes contain that data on Staging, e.g. https://staging.cms.va.gov/node/50916/edit

But on Staging (where this fix is now deployed), filtering by PACT act doesn't show any publications: https://staging.va.gov/outreach-and-events/outreach-materials/

jilladams:
(Will take a look once it's on prod, just in case this is some Staging data weirdness.)

PR #1362, opened 2022-11-15
VACMS-11373 Lovell Giving Wrong Switch Paths

maxx1128:
## Description

Two tickets show how several Lovell pages have incorrect switch links and/or switch link banners. The cause was modifying the Lovell page path was inadvertently changing the `originalPath` variable, which rippled into how the switch links were being set. This is because the `entityUrl.path` was being checked for making switch links, but the change was being done off of `originalPath`. So it lead to some data inconsistencies among pages.

Closes [#11373](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11373) and likely [#11374](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11374)

## Testing done

Debugging, see below screenshots.

## Screenshots

These are the pages that had incorrect switch links. Both pointed to `lovell-federal-health-care` instead of the Tricare/VA inversions. Debugging confirmed they now have the correct switch link URLs.

Screen Shot 2022-11-15 at 2 34 39 PM
Screen Shot 2022-11-15 at 2 35 15 PM

## Acceptance criteria
- [ ] Lovell pages have correct switch links when visible at the top banner
- [ ] Pages without switch links have no banner

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -47,7 +47,6 @@ function getModifiedLovellPage(page, variant) {
       : LOVELL_TRICARE_TITLE_VARIATION;
   const linkVar =
     variant === 'va' ? LOVELL_VA_LINK_VARIATION : LOVELL_TRICARE_LINK_VARIATION;
-  const originalPath = `${page.entityUrl.path}`;
 
   // Add a field for canonical if it has a clone and it's a tricare variant
   if (variant === 'tricare' && isLovellFederalPage(page)) {
@@ -58,10 +57,11 @@ function getModifiedLovellPage(page, variant) {
   }
 
   // Modify the path
-  page.entityUrl.path = `${originalPath}`.replace(
+  const modifiedPath = page.entityUrl.path.replace(
     '/lovell-federal-health-care',
     `/lovell-federal-${linkVar}-health-care`,
   );
+  page.entityUrl.path = modifiedPath;
 
   // Add a switchPath field
   // These get modified later in the processLovellPages function
@@ -73,7 +73,7 @@ function getModifiedLovellPage(page, variant) {
       `/lovell-federal-${LOVELL_VA_LINK_VARIATION}-health-care`,
     )
   ) {
-    page.entityUrl.switchPath = `${page.entityUrl.path}`.replace(
+    page.entityUrl.switchPath = page.entityUrl.path.replace(
       variant === 'va'
         ? LOVELL_VA_LINK_VARIATION
         : LOVELL_TRICARE_LINK_VARIATION,
@@ -82,7 +82,7 @@ function getModifiedLovellPage(page, variant) {
         : LOVELL_VA_LINK_VARIATION,
     );
   } else {
-    page.entityUrl.switchPath = `${page.entityUrl.path}`.replace(
+    page.entityUrl.switchPath = page.entityUrl.path.replace(
       '/lovell-federal-health-care',
       `/lovell-federal-${
         variant === 'va'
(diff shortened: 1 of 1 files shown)


src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -47,7 +47,7 @@ function getModifiedLovellPage(page, variant) {
       : LOVELL_TRICARE_TITLE_VARIATION;
   const linkVar =
     variant === 'va' ? LOVELL_VA_LINK_VARIATION : LOVELL_TRICARE_LINK_VARIATION;
-  const originalPath = page.entityUrl.path;
+  const originalPath = `${page.entityUrl.path}`;
```
ryguyk, line 50: This is interesting. @maxx1128 can you shed some light on why you made this change?

maxx1128: Sure thing! When I was debugging the code I saw that `originalPath` was changing after its assignment. The cause was because `originalPath` was directly assigned `page.entityUrl.path`, so when the latter changed then so did the former. Adding the small string interpolation makes sure it's a string copy disconnected from the first one that won't change it. I used it with other assignment patterns here to be extra cautious about avoiding unexpected string mutations.

ndouglas: I thought "surely there's a way to deep-clone a string in JavaScript that better conveys intent", but after reading [this SO question and responses](https://stackoverflow.com/questions/31712808/how-to-force-javascript-to-deep-copy-a-string) it seems like this is the best way to do it.  I'm pretty disappointed in JavaScript right now 😂

ndouglas: Screen Shot 2022-11-16 at 1 16 34 PM
david-tennant-crying-in-the-rain

maxx1128: We could separate this into a separate function to better convey the intent, with something like this.

```javascript
const deepCloneString = (string) => `${string}`
```

ryguyk: I'm not sure this makes sense to me. The idea of making a copy of a string makes sense on the surface, but I don't think it's necessary here. In fact, I don't think it's likely ever necessary in this context. Take a look at [this JSFiddle](https://jsfiddle.net/pLa2mr16/). 

The string assignment `const originalPath = page.entityUrl.path;` is going to assign the _value_ to `originalPath` (not a memory location/reference). If `page.entityUrl.path` ever changes in the future, the value stored in `originalPath` is _not_ going to change.

I'm not sure where you might have noticed "that originalPath was changing after its assignment". I'd likely need to see that to understand better. But if that's the case, my fear is that something else is going on.

maxx1128: The JSFiddle didn't change original path, including when `replace` was used. I don't see any other areas of the code where that could have been caused, so it's possible that was something I mistook or was a red herring.

If the switch links are working now, it may be due solely to the second change, where the switch links are now being set off of `page.entityUrl.path` instead of `originalPath`. That would explain why the current demo environment is still working, since the pages that weren't being properly set had original paths without `va` or `tricare` in them, but the modified entity URL paths did.

The solution may be to just drop the string deep copy changes, and keep the change from `originalPath` to `page.entityUrl.path` in the assignments.

maxx1128: @ryguyk I think it's worth at least merging this without the deep string copying to be on the safe side.

maxx1128:
I paired with @ryguyk on this and we saw that the string deep copies aren't needed here. The main fix is the reassignment tweaks on lines 76 and 85, and I made another small adjustment where the `page.entityUrl.path` is assigned so it's more clear. This should be good for the last bit of review now and be good for tomorrow's demo.

cc: @ndouglas

swirtSJW (approved):
I like the refined scope.  Waiting on @ryguyk

ryguyk (approved):
Nice work. Thanks for making the few tweaks.

PR #1363, opened 2022-11-15
VACMS-9179: Adds support for caching static data files

ryguyk:
## Description
Closes [va.gov-cms#9179](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9179).

The generation of static (KISS) data files from Drupal was completed in [an earlier PR](https://github.com/department-of-veterans-affairs/content-build/pull/1142) to close [va.gov-cms#8199](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8199). That work excluded any caching mechanism. This PR implements that caching mechanism.

Notably, caching is required for these static data files to work on review instances (review instances do not have SOCKS access and cannot ping Drupal directly). This PR makes it possible to test static-data-reliant features on review instances.

## Testing done
Thorough testing done locally, including edge cases around the static-data-file `config.js` file's existence and structure.

Unfortunately, confirming that this works on a review instance is a bit of a catch-22. It is not possible until after the data is cached. Once this merges, content builds will save static data files to cache, and the review instances will pull from this cache.


swirtSJW (commented):
Concept sounds good and I don't see any obvious flaws.  I'm not sure how to test this right now.

ryguyk:
@swirtSJW Thanks. I've added some in-depth QA steps that should help.

That said, there's no real good way to test this beyond local because of the catch-22 mentioned. Review instances will not be able to read from the cache until it exists, and it will not exist until this gets merged and a content build is run to generate the cache.

In fact, as mentioned above as well, this will require a two-step merge in order to accommodate that catch-22.

it-harrison (approved):

swirtSJW (approved):

swirtSJW (approved):
IN light of Catch-22 not allowing this to be fully tested,  this seems low risk to merge as any catastrophic  failure would have caused tests to fail.

apisandipas (approved):
Great job with the QA Steps, it was a very clear process. This PR looks good to me.

PR #1385, opened 2022-11-30
[VA-11538] New Homepage "Create Account" Updates

maxx1128:
## Description

The redesigned home page has some text and style changes for the "Create Account" section in the top right.

closes [#11538](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11538)
relates to [#11570](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11570)

## Testing done & Screenshots

Visual

Screen Shot 2022-11-30 at 1 50 23 PM

## QA steps

1. Check out this branch and do a local PR build
2. Go to `http://localhost:3002/new-home-page/`
3. Confirm that the Create Account area has the new styles and elements as expected

1. The other areas of the new homepage need to be checked for proper styling and functionality to make sure they weren't indirectly broken.

1. Test the page on different screen sizes
   - [ ] Validate that the Create Account area still renders properly, visibly, and is still accessible
2. Then try using the page with a screen reader
   - [ ] Validate that the content is read properly and not conveying any different meaning than what visual users get.

## Acceptance criteria

- [ ] "Welcome to VA.gov" is bigger and has 1px width
    * Using a larger font size broke the text into two lines, so changing this will likely require an additional change from `vets-website` where the headline width is set.
- [ ] The link to the new account creation page in Resources and Support has the "Learn how an account helps you" label and is styled as a link. 
- [ ] A Veteran or beneficiary can click on the "Learn how an account helps you" label to view the new Resource and Support account creation page [_link TBD_, left as `#` placeholder for now]
- [ ] "Create account" component has updated intro text. 

## Definition of done

### DIFF
--- a/src/site/includes/hero.drupal.liquid
+++ b/src/site/includes/hero.drupal.liquid
@@ -53,6 +53,9 @@
             <button class="vads-u-padding-x--4 vads-u-margin-bottom--3" onclick="openLoginModal()">
               Create account
             </button>
+            <a href="#">
+              Learn how an account helps you
+            </a>
           </div>
         </div>
       </div>
--- a/src/site/includes/hero.drupal.liquid
+++ b/src/site/includes/hero.drupal.liquid
@@ -10,7 +10,7 @@
             vads-u-padding-bottom--3
             small-desktop-screen:vads-u-padding-bottom--8">
           <h1
-            class="homepage-hero__welcome-headline vads-u-color--white vads-u-margin-top--0 vads-u-margin-bottom--2 vads-u-padding-y--1 vads-u-border-top--2px vads-u-border-bottom--2px vads-u-font-size--lg vads-u-font-family--serif">
+            class="homepage-hero__welcome-headline vads-u-color--white vads-u-margin-top--0 vads-u-margin-bottom--2 vads-u-padding-y--1 vads-u-border-top--1px vads-u-border-bottom--1px vads-u-font-size--lg vads-u-font-family--serif">
             Welcome to VA.gov</h1>
           <h2
             class="vads-u-color--white vads-u-margin-top--3 vads-u-font-size--xl small-desktop-screen:vads-u-font-size--2xl">
--- a/src/site/includes/hero.drupal.liquid
+++ b/src/site/includes/hero.drupal.liquid
@@ -45,12 +45,12 @@
       <div class="vads-l-col--12 medium-screen:vads-l-col--6 homepage-hero__container">
         <div class="vads-u-display--flex vads-u-width--full vads-u-align-items--center vars-u-justify-content--center">
           <div
-            class="vads-u-background-color--white vads-u-padding-x--3 vads-u-padding-y--2 vads-u-width--full homepage-hero__create-account ">
+            class="va-flex vads-u-flex-direction--column vads-u-align-items--flex-start vads-u-background-color--white vads-u-margin-top--6 vads-u-margin-bottom--6 vads-u-padding-x--3 vads-u-padding-y--2 vads-u-width--full homepage-hero__create-account ">
             <h2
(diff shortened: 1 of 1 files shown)


src/site/includes/hero.drupal.liquid
```diff
@@ -45,14 +45,17 @@
       <div class="vads-l-col--12 medium-screen:vads-l-col--6 homepage-hero__container">
         <div class="vads-u-display--flex vads-u-width--full vads-u-align-items--center vars-u-justify-content--center">
           <div
-            class="vads-u-background-color--white vads-u-padding-x--3 vads-u-padding-y--2 vads-u-width--full homepage-hero__create-account ">
+            class="va-flex vads-u-flex-direction--column vads-u-align-items--flex-start vads-u-background-color--white vads-u-margin-top--6 vads-u-margin-bottom--6 vads-u-padding-x--3 vads-u-padding-y--2 vads-u-width--full homepage-hero__create-account ">
             <h2
-              class="vads-u-font-size--lg vads-u-margin-top--0 vads-u-padding-right--2 vads-u-font-family--sans vads-u-font-weight--normal">
-              Get access to the tools you need to manage your VA benefits and care
+              class="vads-u-font-size--md vads-u-color--gray vads-u-margin-top--0 vads-u-padding-right--2 vads-u-font-family--sans vads-u-font-weight--normal">
+              Create an account to manage your VA benefits and care in one place - any time, from anywhere."
```
jilladams, line 51: s/-/—

jilladams (changes requested):

src/site/includes/hero.drupal.liquid
jilladams, line 51: - [ ] Remove trailing "

jilladams (changes requested):

src/site/includes/hero.drupal.liquid
jilladams, line 51: @maxx1128 can you also confirm the font treatment for the "create an account" copy here? Two questions: 
1. the line-height is more spaced out in the design vs. your screenshot of build. We should lean on the design system, so if you can point to design system guidance, that's prob fine. If not, let's try to match the design.
2. The markup looks like that text is being handled as an h2. Ticket doesn't spell out how this should be handled semantically, so you may need to consult with @laflannery if we aren't sure whether / why that's the right call.

jilladams (changes requested):

src/site/includes/hero.drupal.liquid
laflannery: @jilladams I believe you are talking about the "Create an account to manage...." text and if so an H2 is correct, it's my original guidance from the mural, I would still like this to be an H2 even though the text changed:
image

Also, looking at the screenshots above I noticed in the PR screenshot there is a " character at the end but that isn't in sketch.  I didn't notice that in your comment so just making sure we remove that extra character if that is in fact just a typo.
image

maxx1128: @jilladams Good catch with the line-height, I agree it's off from what's in the design. I'll adjust that now along with all the other updates.

maxx1128:
Updated local homepage.

Screen Shot 2022-12-01 at 9 14 38 AM

apisandipas (approved):
This looks good to me. Noting that the link for “Learn how an account helps you” is TBD.

jilladams:
@mmiddaugh FYI this is ready to go, with one placeholder link: “Learn how an account helps you" in the Create Account section. Any update on where that should point?

jilladams (approved):

mmiddaugh:
@jilladams The page is in process but the URL won't be finalized until the content is (since it uses the H1). We may not have the URL for that page until the staging review.

jilladams:
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11771 ticketed for follow up on that URL later.

PR #1393, opened 2022-12-08
[VA-11691] Keep Switch links for pages with missing breadcrumbs

maxx1128:
## Description

The Lovell VA Policies page was missing its switch link. This was due to some page logic where it adds in a breadcrumb path if it's missing, which inadvertently removed any properties that were `breadcrumb` or `url`. This change keeps any other properties already there to preserve that data, including switch links.

closes [#11691](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11691)

## Testing done & Screenshots

Visually, see below.

Screen Shot 2022-12-08 at 1 16 45 PM

## QA steps

Build Lovell locally and check the [Lovell VA Policies page.](http://localhost:3002/lovell-federal-va-health-care/policies/) The switch link should be at the top of the page.

Check the Lovell Tricare Policies page (the switch link) and other Lovell pages to ensure those links still work as well.

1. Go to the Lovell VA Policies page
   - [ ] The switch link is there
2. Then go to the other Lovell pages
   - [ ] Their switch links still work as expected

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/page.js
+++ b/src/site/stages/build/drupal/page.js
@@ -353,7 +353,10 @@ function compilePage(page, contentData) {
   const pageId = { pid: pageIdRaw };
 
   if (!('breadcrumb' in entityUrl)) {
-    page.entityUrl = generateBreadCrumbs(entityUrl.path);
+    page.entityUrl = {
+      ...entityUrl,
+      ...generateBreadCrumbs(entityUrl.path),
+    };
   }
 
   let pageCompiled;


src/site/stages/build/drupal/page.js
```diff
@@ -353,7 +353,10 @@ function compilePage(page, contentData) {
   const pageId = { pid: pageIdRaw };
 
   if (!('breadcrumb' in entityUrl)) {
-    page.entityUrl = generateBreadCrumbs(entityUrl.path);
+    page.entityUrl = {
+      ...page.entityUrl,
```
ryguyk, line 357: **CONSIDER**
It _might_ make more sense to reference `entityUrl` rather than `page.entityUrl` on line 357. These point to the same object in memory, but it feels like the "read" on line 357 should read the const that was defined on line 350. The "write" on line 356 has to be `page.entityUrl`, so that one cannot change.

(This is not a required change.)

ryguyk (approved):
Approved with one change for consideration.

src/site/stages/build/drupal/page.js
maxx1128: I agree it makes a bit more sense to do this, I overlooked the destructuring on 350 that would fit here 👍

PR #1396, opened 2022-12-09
VACMS-11310: reduce logging in content build

timcosgrove:
**note: marked do not merge while discussed with CMS teams**

## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11310.

The Content Build process takes an optional flag `--verbose` which outputs more information about the build process. This flag is hard-coded on, so we always have verbose logs.

This PR makes using `--verbose` dependent on the workflow being run with the debug flag set. When workflows are being run in debug mode, a secret `ACTIONS_RUNNER_DEBUG` is set to `true`. We use this in turn to decide whether to turn on verbose logging. 

Debug logging can also be turned on at a workflow level by a repo admin, so we have the option to turn this on for all runs as necessary.

## Testing done & Screenshots
The logic was implemented in a test workflow we keep for work on Github Actions: https://github.com/department-of-veterans-affairs/content-build/actions/workflows/content-release-testing.yml

The workflow was run with an intentional failure; once it failed, it was rerun with 'enable debug logging' checked.

* Run without debug: https://github.com/department-of-veterans-affairs/content-build/actions/runs/3661443109/attempts/1
* Run with debug on: https://github.com/department-of-veterans-affairs/content-build/actions/runs/3661443109

Specifically, the Build step in the Build job of those workflows shows differences in logging, e.g.

* Without debugging: https://github.com/department-of-veterans-affairs/content-build/actions/runs/3661443109/jobs/6189681949#step:16:796
* With debugging: https://github.com/department-of-veterans-affairs/content-build/actions/runs/3661443109/jobs/6190679758#step:16:864


.github/workflows/content-release.yml
```diff
@@ -257,8 +259,12 @@ jobs:
           url: ${{ env.DRUPAL_ADDRESS }}/api/content_release/inprogress
           method: GET
 
+      - name: Get verbose level
+        if: ${{ env.ACTIONS_RUNNER_DEBUG == 'true' }}
+        run: echo "VERBOSE_BUILD= --verbose" >> $GITHUB_ENV
+
       - name: Build
-        run: NODE_OPTIONS=--max-old-space-size=${{ env.MAXIMUM_HEAP }} yarn build:content_release --buildtype=${{ env.BUILDTYPE }} --asset-source=local --drupal-address=${{ env.DRUPAL_ADDRESS }} --drupal-user=${{ env.DRUPAL_USERNAME }} --drupal-password="${{ env.DRUPAL_PASSWORD }}" --pull-drupal --drupal-max-parallel-requests=15 --no-drupal-proxy --verbose | tee build-output.txt
+        run: NODE_OPTIONS=--max-old-space-size=${{ env.MAXIMUM_HEAP }} yarn build:content_release --buildtype=${{ env.BUILDTYPE }} --asset-source=local --drupal-address=${{ env.DRUPAL_ADDRESS }} --drupal-user=${{ env.DRUPAL_USERNAME }} --drupal-password="${{ env.DRUPAL_PASSWORD }}" --pull-drupal --drupal-max-parallel-requests=15 --no-drupal-proxy ${{env.VERBOSE_BUILD}} | tee build-output.txt
```
timcosgrove, line 267: If the variable is not set this prints nothing, which is fine.

olivereri (approved):

swirtSJW (approved):
This approach makes sense to me.  My main concerns are:

1.  that it would prevent debugging - but you addressed that by saying that it can be turned on for subsequent runs if we need to see what is happening
2. that it would hide something failing - and it does not seem that it is capabable of doing that.

So I think my concerns are addressed.  Thank you for implementing this.

PR #1399, opened 2022-12-13
Fix URLs for Lovell federal pages with 'va'

maxx1128:
## Description

Lovell federal pages should be checked if they have `va` in the CMS URL. This could potentially be creating a bug where federal pages aren't being properly cloned for the Tricare side, hence why so many of those pages have correct menu links but no actual pages.

This PR checks if a page is a Federal page, and if the URL is already the opposite of the page variation being built (for example, the Tricare variation of a federal page but it already has `va` in the URL). This resets the URL to remove the variation so it can have the proper URL set and go through the rest of the page build steps without issue.

## Testing done & Screenshots

Confirmed it locally.

## QA steps

Rebuild Lovell locally with `yarn build --pull-drupal --drupal-address=https://main-liqqgmzl7tsyfbwyulbkuborufuuttav.demo.cms.va.gov/` and check the Lovell pages. Make sure that all the Tricare versions of them exist and can be clicked on in both the VA switch links and the Tricare side menu.

## Acceptance criteria

- [ ] All federal pages have proper variations built for both `va` and `tricare`.
- [ ] All Tricare pages can be accessed correctly in the menu and switch links.

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -48,6 +48,22 @@ function getModifiedLovellPage(page, variant) {
   const linkVar =
     variant === 'va' ? LOVELL_VA_LINK_VARIATION : LOVELL_TRICARE_LINK_VARIATION;
 
+  // Fix any incorrect URLs based on the variant
+  if (isLovellFederalPage(page)) {
+    const oppositeVariant =
+      variant === 'va'
+        ? LOVELL_TRICARE_LINK_VARIATION
+        : LOVELL_VA_LINK_VARIATION;
+    const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;
+
+    if (page.entityUrl.path.includes(reverseUrl)) {
+      page.entityUrl.path = page.entityUrl.path.replace(
+        reverseUrl,
+        `/lovell-federal-health-care`,
+      );
+    }
+  }
+
   // Add a field for canonical if it has a clone and it's a tricare variant
   if (variant === 'tricare' && isLovellFederalPage(page)) {
     page.canonicalLink = page.entityUrl.path.replace(


ryguyk (approved):
**I think this is serving the intended purpose. We should merge it.**

--------

That said, I think we should think about refactoring this function sooner than later. This additional step in the process of modifying a Lovell page seems confusing, and it's adding to a process that is already confusing. I think it'd serve  us all very well to try to zoom out a bit and nail down what exactly we are attempting to do with this function. I think it can likely be cleaned up a bit to - first and foremost - be more easy to reason about.

**One example:**
The logic that is being added in this PR modifies `page.entityUrl.path`. There is already a block of code below it that also modifies `page.entityUrl.path`. It feels like it'd be a lot more clear to merge or co-locate this logic. It could possibly even be merged with the setting of `page.canonicalLink` which sits between the two blocks currently, since they both have behavior that changes according to if the page is a Lovell Federal page.

**Additional example/concern:**
I'm fairly certain that the `else` condition in the code below is practically unreachable. If there is data that would make it reachable, it would take no useful action.
```
// Modify the path
  const modifiedPath = page.entityUrl.path.replace(
    '/lovell-federal-health-care',
    `/lovell-federal-${linkVar}-health-care`,
  );
  page.entityUrl.path = modifiedPath;

  // Add a switchPath field
  // These get modified later in the processLovellPages function
  if (
    page.entityUrl.path.includes(
      `/lovell-federal-${LOVELL_TRICARE_LINK_VARIATION}-health-care`,
    ) ||
    page.entityUrl.path.includes(
      `/lovell-federal-${LOVELL_VA_LINK_VARIATION}-health-care`,
    )
  ) {
    page.entityUrl.switchPath = page.entityUrl.path.replace(
      variant === 'va'
        ? LOVELL_VA_LINK_VARIATION
        : LOVELL_TRICARE_LINK_VARIATION,
      variant === 'va'
        ? LOVELL_TRICARE_LINK_VARIATION
        : LOVELL_VA_LINK_VARIATION,
    );
  } else {
    page.entityUrl.switchPath = page.entityUrl.path.replace(
      '/lovell-federal-health-care',
      `/lovell-federal-${
        variant === 'va'
          ? LOVELL_TRICARE_LINK_VARIATION
          : LOVELL_VA_LINK_VARIATION
      }-health-care`,
    );
  }
```

This is concerning, but it isn't something to fix here and now. This PR almost certainly makes things work better than they did previously, so we should get it merged. But, again, I think our future selves will thank us if we clean this up sooner than later.

maxx1128:
I agree with @ryguyk that the Entity URL logic should be improved in the future. I can open up a ticket for that as a reminder than then merge this code in.

PR #1408, opened 2022-12-20
Revert to older output setting syntax for broken link checking.

timcosgrove:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12040

This reverts these scripts back to older, deprecated syntax to restore broken link checking. A follow-up ticket will be made to migrate the scripts to become proper custom actions.

### DIFF
--- a/script/github-actions/check-broken-links-blocks.js
+++ b/script/github-actions/check-broken-links-blocks.js
@@ -91,7 +91,7 @@ if (fs.existsSync(reportPath)) {
   console.log(
     `${brokenLinks.brokenLinksCount} broken links found. \n ${brokenLinks.summary}`,
   );
-  console.log(`SLACK_BLOCKS=${JSON.stringify(payload)} >> $GITHUB_OUTPUT`);
+  console.log(`::set-output name=SLACK_BLOCKS::${JSON.stringify(payload)}`);
 
   if (!IS_PROD_BRANCH && !contentOnlyBuild) {
     // Ignore the results of the broken link checker unless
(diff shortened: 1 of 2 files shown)


script/github-actions/check-broken-links-blocks.js
```diff
@@ -91,7 +91,7 @@ if (fs.existsSync(reportPath)) {
   console.log(
     `${brokenLinks.brokenLinksCount} broken links found. \n ${brokenLinks.summary}`,
   );
-  console.log(`SLACK_BLOCKS=${JSON.stringify(payload)} >> $GITHUB_OUTPUT`);
+  console.log(`::set-output name=SLACK_BLOCKS::${JSON.stringify(payload)}`);
```
ndouglas, line 94: 🤔

ndouglas (approved):

swirtSJW (approved):
Worked before, should be no worries reverting.

script/github-actions/check-broken-links-blocks.js
timcosgrove: the change is from first line to second:
```
echo ::set-output name=FOO::BAR
echo FOO=BAR >> $GITHUB_OUTPUT
```

I made the (apparently bad) assumption that `console.log` would act as `echo` in both cases as far as GHA was concerned, since it was working for the first case. It's possible there was a different error also. 

I'm going to approach the broken link report script as its own ticket. It might be figuring out what my simple error was; or, it might mean reimplementing the broken link checker as a proper Github Action.

ndouglas: $GITHUB_OUTPUT is a file, right?  Use fs.writefilesync imho

PR #1409, opened 2022-12-20
VACMS-12011 Fix Lovell Tricare Menu links

maxx1128:
## Description

The "opposite variant" URL checks need to be applied to menu links as well as page links. This change puts that logic into a function that can be applied to both and fix the broken Tricare links that would otherwise bring links back to the VA pages.

closes [#12011](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12011)

## Testing done & Screenshots

Visual testing. See URLs from the highlighted menu links in the bottom left of each screenshot.

Screen Shot 2022-12-20 at 3 28 47 PM
Screen Shot 2022-12-20 at 3 28 52 PM

## QA steps

Tricare menu links, and they link to other Tricare pages.

The menu links on VA pages still link to VA pages, and the switch links still work properly.

## Acceptance criteria

- [ ] The side menu on Tricare pages link to other Tricare pages, not VA pages
- [ ] VA pages still link to other VA pages

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -200,6 +200,19 @@ function lovellMenusModifyLinks(link) {
       `${LOVELL_TITLE_STRING} ${titleVar}`,
     );
 
+    const oppositeVariant =
+      variant === 'va'
+        ? LOVELL_TRICARE_LINK_VARIATION
+        : LOVELL_VA_LINK_VARIATION;
+    const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;
+
+    if (link.url.path.includes(reverseUrl)) {
+      link.url.path = link.url.path.replace(
+        reverseUrl,
+        `/lovell-federal-health-care`,
+      );
+    }
+
     link.url.path = link.url.path.replace(
       '/lovell-federal-health-care',
       `/lovell-federal-${linkVar}-health-care`,
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -40,6 +40,18 @@ function isListingPage(page) {
   return listingPageTypes.includes(page.entityBundle);
 }
 
+function resetToFederalUrlIfNeeded(path, variant) {
+  const oppositeVariant =
+    variant === 'va' ? LOVELL_TRICARE_LINK_VARIATION : LOVELL_VA_LINK_VARIATION;
+  const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;
+
+  if (path.includes(reverseUrl)) {
+    return path.replace(reverseUrl, `/lovell-federal-health-care`);
+  }
+
+  return path;
+}
+
 function getModifiedLovellPage(page, variant) {
   const fieldOfficeMod =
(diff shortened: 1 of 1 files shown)


src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -40,6 +40,18 @@ function isListingPage(page) {
...
+    variant === 'va' ? LOVELL_TRICARE_LINK_VARIATION : LOVELL_VA_LINK_VARIATION;
+  const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;
+
+  if (path.includes(reverseUrl)) {
+    return path.replace(reverseUrl, `/lovell-federal-health-care`);
+  }
+
+  return path;
+}
+
```
ryguyk, line 54: **CONSIDER**
I think my personal preference would be to name this starting with lowercase: `urlOppositeVariant`. I'm not sure if we have any coding standards specifically mandating that.

ryguyk: **SHOULD**
I don't think this function is doing what it advertises. I'm not sure it actually ever returns the opposite-variant url, does it? There are two possible return values:
1. If the condition included on lines 48-50 evaluates to true, it will return a url without any variant (a federal url).
2. If the condition evaluates to false, it'll return the original path.

src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -50,18 +62,7 @@ function getModifiedLovellPage(page, variant) {
...
-        : LOVELL_VA_LINK_VARIATION;
-    const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;
-
-    if (page.entityUrl.path.includes(reverseUrl)) {
-      page.entityUrl.path = page.entityUrl.path.replace(
-        reverseUrl,
-        `/lovell-federal-health-care`,
-      );
-    }
+    page.entityUrl.path = URLOppositeVariant(page.entityUrl.path, variant);
```
ryguyk, line 65: This relates to the comment above about the function not doing what it advertises. It seems to be confirmed here. If this were doing strictly what is advertised, I don't think it would make sense. I don't think we would want to set the page path to be the opposite variant. I think what this function is really doing is _resetting_ the page path to be the federal path.

ryguyk (changes requested):
A few changes to consider. It largely boils down to naming, though might possibly imply a few other tweaks.

src/site/stages/build/drupal/process-lovell-pages.js
maxx1128: Resetting it to the federal path is a better name for it, the opposite URL part is more part of the logic used instead of the larger purpose. Some better names would probably be `resetToFederalUrl` or `resetFederalUrlIfNeeded`.

ryguyk (approved):
I think this name change makes things much more clear!

I still wonder if this is the best overall approach to this particular logic. It seems confusing (though I ultimately understand what's going on) to reset the url based on the opposite variant. All this said, I think we can address this in our pending work to refactor this function.

swirtSJW (approved):
This is approved.  Waiting only on code freeze to be lifted to merge.

swirtSJW:
I am closing the corresponding ticket.

PR #1412, opened 2022-12-22
VACMS-11995 Add Lovell switch link Cypress tests

maxx1128:
## Description

Lovell switch links are important to test, both as a feature and a quick health check of Lovell's structure. This adds Cypress tests to some of the key Lovell pages.

closes [#11995](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11995)

## Testing done & Screenshots

Visual, Cypress, see local results below.

Screen Shot 2022-12-22 at 9 17 03 AM

## QA steps

**What needs to be checked to prove this works?** Do a local Lovell build with `yarn build --pull-drupal --drupal-address=https://main-liqqgmzl7tsyfbwyulbkuborufuuttav.demo.cms.va.gov/`, then run cypress tests locally with `yarn cy:run`.
**What needs to be checked to prove it didn't break any related things?** Click through a few other Lovell pages to ensure they're still working normally.
**What variations of circumstances (users, actions, values) need to be checked?** None.

## Acceptance criteria

- [ ] Lovell switch links are properly checked, so the tests pass when the pages work and fail when they don't.

## Definition of done

### DIFF
--- a/src/site/tests/lovell/switch-links.cypress.spec.js
+++ b/src/site/tests/lovell/switch-links.cypress.spec.js
@@ -1,0 +1,56 @@
+const lovellSwitchLinkPages = [
+  {
+    title: 'System Health Care Page',
+    path: '',
+  },
+  {
+    title: 'Locations',
+    path: '/locations',
+  },
+  {
+    title: 'Health Services',
+    path: '/health-services',
+  },
+  {
+    title: 'Listings (Events)',
+    path: '/events',
+  },
+  {
+    title: 'Listings (News Releases)',
+    path: '/news-releases',
+  },
+  {
+    title: 'Listings (Stories)',
+    path: '/stories',
+  },
+];
+
+const getLovellUrl = (path, variation) =>
+  `${`/lovell-federal-${variation}-health-care`}${path}`;
+
+const lovellVariations = ['va', 'tricare'];
+
+lovellSwitchLinkPages.forEach(page => {
+  lovellVariations.forEach(variation => {
+    const oppositeVariation = variation === 'va' ? 'tricare' : 'va';
+
+    describe(`Lovell ${variation.toUpperCase()} ${page.title} page`, () => {
+      it('has the proper switch link', () => {
+        cy.visit(getLovellUrl(`${page.path}/`, variation));
+        cy.injectAxeThenAxeCheck();
+
+        cy.get('va-alert').should('have.length', 1);
+
+        cy.get('va-alert').contains(
+          `View this page as a ${oppositeVariation.toUpperCase()} beneficiary`,
+        );
+
+        cy.get('va-alert a').should(
+          'have.attr',
+          'href',
+          getLovellUrl(page.path, oppositeVariation),
+        );
+      });
+    });
+  });
+});


maxx1128:
The tests are failing in the pull request checks until the lovell pages are published and then available in the test environment. The tests all pass for me locally, so once the pages are in the test env, they should pass here. I'm making this as "do not merge" for now.

src/site/tests/lovell/switch-links.cypress.spec.js
```diff
@@ -0,0 +1,56 @@
+const lovellSwitchLinkPages = [
+  {
+    title: 'System Health Care Page',
+    path: '',
+  },
+  {
+    title: 'Locations',
+    path: '/locations',
```
swirtSJW, line 8: Would these run faster if they used the direct location rather than having to wait on the redirect from /path to /path/  ?
```suggestion
    path: '/locations/',
```

swirtSJW: Nevermind.  I see where you are setting the trailing slash down on line 39.

src/site/tests/lovell/switch-links.cypress.spec.js
```diff
@@ -0,0 +1,56 @@
...
+const lovellVariations = ['va', 'tricare'];
+
+lovellSwitchLinkPages.forEach(page => {
+  lovellVariations.forEach(variation => {
+    const oppositeVariation = variation === 'va' ? 'tricare' : 'va';
+
+    describe(`Lovell ${variation.toUpperCase()} ${page.title} page`, () => {
+      it('has the proper switch link', () => {
+        cy.visit(getLovellUrl(`${page.path}/`, variation));
+        cy.injectAxeThenAxeCheck();
```
swirtSJW, line 40: I'm not clear on the need for this.  Why do we have to make this run axe?

swirtSJW (commented):
This is all looking good @maxx1128   I just have the one question related to calling Axe.

src/site/tests/lovell/switch-links.cypress.spec.js
maxx1128: I was getting linting warnings in my code environment about needing an Axe check here. I added them in since the warnings had kept me from pushing commits to pull requests in the past. I think those warnings got added in the past to just enforce more accessibility testing.

swirtSJW: That seems odd though as it entangles a set of tests here that do not need to be entangled.   Also seems like it may be firing Axe tests on this content that may already be covered by other tests which would cause slower tests due to duplication of coverage.

I am hoping maybe @ryguyk can give me a better understanding of this.

ryguyk:
I actually have my own questions about this requirement. Linting rules seem to mandate axe checks on all E2E tests. I'm not sure how I feel about this yet.

swirtSJW:
> I actually have my own questions about this requirement. Linting rules seem to mandate axe checks on all E2E tests. I'm not sure how I feel about this yet.

Seems redundant, duplicative, and repetitious  ;)

swirtSJW:
I raised the Axe test question here. https://dsva.slack.com/archives/C023ZKDJJK1/p1672411047448619
I don't think this needs to block the current work. So I am going to merge this as-is into the integration branch and if we make some headway on removing the axe linting, we can remove it later.

PR #1416, opened 2022-12-27
VA-12069: Only show patient satisfaction scores for vha API endpoints

maxx1128:
## Description

Patient satisfaction scores aren't available for Tricare pages, but the widget for them is still on the pages. This causes an infinite spinner since it waits for data that will never come.

This pull request puts a conditional around the facility satisfaction scores so they only appear for appropriate endpoints. Tricare pages don't have the widget, and the functional VA ones have the widget (it doesn't load properly on my device only due to a local API issue, it won't affect the overall functionality).

closes [#12069](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12069)

### Tricare Page

Screen Shot 2022-12-28 at 12 31 59 PM

### VA Page

Screen Shot 2022-12-28 at 12 31 23 PM

## QA steps

**What needs to be checked to prove this works?** Lovell Tricare location pages need not to show the patient satisfaction scores widget.
**What needs to be checked to prove it didn't break any related things?** VA location pages still need to show it properly.
**What variations of circumstances (users, actions, values) need to be checked?** Just checking the Lovell pages.

## Acceptance criteria

- [ ] Lovell Tricare pages no longer show patient satisfaction scores
- [ ] Lovell VA pages still successfully show patient satisfaction scores

## Definition of done

### DIFF
--- a/src/site/layouts/health_care_local_facility.drupal.liquid
+++ b/src/site/layouts/health_care_local_facility.drupal.liquid
@@ -209,9 +209,11 @@
             </section>
           {% endif %}
 
-          <div data-widget-type="facility-patient-satisfaction-scores"
-               data-facility="{{ fieldFacilityLocatorApiId | widgetFacilityDetail | escape }}">
-          </div>
+          {% if fieldFacilityLocatorApiId contains "vha_" %}
+            <div data-widget-type="facility-patient-satisfaction-scores"
+                data-facility="{{ fieldFacilityLocatorApiId | widgetFacilityDetail | escape }}">
+            </div>
+          {% endif %}
 
           <!-- Social Links -->
           {% include "src/site/facilities/facility_social_links.drupal.liquid"


swirtSJW (approved):

ryguyk (approved):
This is a good quick solution that improves the experience, and we should merge it.

I do think we should explore whether this template is the right place for this logic. My conviction is that this sort of check is putting too much responsibility on this template, and creating too much logic for exceptions that is likely unnecessary. This could almost certainly be solved more flexibly by ensuring that the app itself handles things properly if the api call does not return a valid facility. After a brief look into that, it seems that the api returns a 200 even when the facility is invalid (the 200 result is an object that indicates there's an error) but the app does not handle this as an error condition. That piece of things is probably the place this should ultimately be fixed.
cc: @maxx1128 @swirtSJW

swirtSJW:
This is mergeable and is just on hold until code freeze lifts.  I closed the associated issue.

PR #1418, opened 2022-12-28
VACMS-11996: adds unit tests for Lovell listing pages

ryguyk:
## Description
closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11996

Listing pages (stories, events, news) for Lovell Health Care are created by combining the entities from the Lovell Federal listing pages with the entities from the Lovell Tricare and Lovell VA listing pages. Entities on Lovell Federal should ultimately be displayed on both Lovell Tricare and Lovell VA listing pages so that a listing page for Lovell Tricare, for example, is comprised of entities that are specific to Tricare _combined with_ entities that are defined at the Federal level.

This PR adds unit tests to confirm that the Lovell-specific processing correctly handles things as described above.

## Testing done & Screenshots
This PR is entirely focused on creating tests. The tests all pass with the created fixtures.

## QA steps
Locally
1. `yarn test:unit src/site/stages/build/drupal/tests/lovell/listingPages.unit.spec.js`
   - [ ] Validate that there are six (6) passing tests and no failing tests.

On review instance
1. Examine unit tests
   - [ ] Validate that the six (6) tests in question pass.

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -460,5 +460,7 @@ function processLovellPages(drupalData) {
 }
 
 module.exports = {
+  isLovellTricarePage,
+  isLovellVaPage,
   processLovellPages,
 };
--- a/src/site/stages/build/drupal/tests/lovell/fixtures/listing-pages/federal/events.json
+++ b/src/site/stages/build/drupal/tests/lovell/fixtures/listing-pages/federal/events.json
@@ -1,0 +1,970 @@
+{
+  "entityBundle": "event_listing",
+  "entityId": "36319",
+  "entityPublished": true,
+  "title": "Events",
+  "vid": 736012,
+  "entityUrl": {
+    "breadcrumb": [
+      {
+        "url": {
+          "path": "/",
+          "routed": true
+        },
+        "text": "Home"
+      },
+      {
+        "url": {
+          "path": "/lovell-federal-health-care",
+          "routed": true
+        },
+        "text": "Lovell Federal health care"
+      },
+      {
+        "url": {
+          "path": "",
+          "routed": true
+        },
+        "text": "NEWS AND EVENTS"
+      },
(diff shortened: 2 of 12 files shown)


apisandipas (approved):
Looks good to me.

maxx1128 (changes requested):
The tests technically pass here and the data is set up well, but I think the unit tests need to check more than the number of stories. The test checking that Federal and Tricare stories are properly merged could accidentally merge the Federal and VA stories together, but since the total number of stories is still two, the test would still pass. They should also check the correct groupings of stories/news/events.

The easiest way I think to do this is to check the titles of whatever is being filtered. Since it would be in each test, my thought is to add a function like this to the top:

```javascript
const checkEntitiesFromCategories = (entities, category) => {
  return entities
    .map(entity => entity.title)
    .every(
      title =>
        title.includes('Both') ||
        title.includes('Federal') ||
        title.includes(category),
    );
};
```

Then each test could have a check like this:

```javascript
const hasEventsFromProperCategories = checkEntitiesFromCategories(
  mergedEvents.pastEvents.entities,
  'TRICARE', // Or 'VA', depending on the test
);
expect(hasEventsFromProperCategories).to.be.true;
```

I tried this out on a few of the unit tests and it works as expected, it just may need some editing or renaming.

ryguyk:
@maxx1128 That's some very good feedback. I agree the tests would be much more robust with checks beyond simply entity counts, so I've added tests to check merged titles. 

Just a note: I did not implement exactly in the fashion you laid out above. Instead, similar to the existing tests checking the counts, I am comparing the exact merged titles to the exact initial data. I think that is a bit more robust than hard-coding in the test the expected strings.

Thanks for the thoughtful feedback!

maxx1128 (approved):
This looks good! The tests are great and check the titles, I played with them a bit locally and they check out 👍

swirtSJW (approved):
Nice work!

PR #1430, opened 2023-01-06
[VA-12013] Refacter `getModifiedLovellPage` function

maxx1128:
## Description

The Lovell `getModifiedLovellPage` function is big and confusing. This PR does a few things to simplify it and make things easier to manage.

* The code for changing each property (title, URL, switch link, etc) was extracted to individual functions so each one has a single, clear purpose.
* Each of these functions has been moved to a separate `lovell/update-page.js` file.
* These functions have been made into pure functions, so they only return values without changing them. This makes it more explicit in `process-lovell-pages.js` what properties are being changed and under what conditions. This also makes them easier to edit and understand.
* The page variables needed by all these functions are created by a single function, and are passed to all the above functions as a single object to be destructured. This keeps the code in `process-lovell-pages.js` from getting too noisy, and each pure function can be explicit in what it needs without code crowding in the functions either.
* The Lovell processing variables and helper functions (page IDs, variation strings, page type checks) have also been extracted to a new file, `lovell/helpers.js`. This is because they're used in multiple files, and having them in a separate, single source of truth makes them easier to use and understand.

closes [#12013](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12013)

## Testing done & Screenshots

Check local Lovell build pages, automated tests.

## QA steps

**What needs to be checked to prove this works?**
Lovell pages still work as normal and the automated tests still pass.

**What needs to be checked to prove it didn't break any related things?**
Lovell pages and checks.

## Acceptance criteria

- [ ] Lovell pages all still function as expected
- [ ] All current tests pass

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/lovell/helpers.js
+++ b/src/site/stages/build/drupal/lovell/helpers.js
@@ -45,24 +45,15 @@ function getLovellTitleVariation(variant) {
     : LOVELL_TRICARE_TITLE_VARIATION;
 }
 
-function getLovellUrl(urlVar) {
-  return `/lovell-federal-${urlVar}-health-care`;
+function getLovellUrl(linkVar) {
+  return `/lovell-federal-${linkVar}-health-care`;
 }
 
-function getLovellFormOfExistingUrl(url, urlVar) {
-  return url.replace(LOVELL_BASE_URL, getLovellUrl(urlVar));
-}
-
-function resetToFederalUrlIfNeeded(path, variant) {
-  const oppositeVariant =
-    variant === 'va' ? LOVELL_TRICARE_LINK_VARIATION : LOVELL_VA_LINK_VARIATION;
-  const reverseUrl = getLovellUrl(oppositeVariant);
-
-  if (path.includes(reverseUrl)) {
-    return path.replace(reverseUrl, LOVELL_BASE_URL);
-  }
-
-  return path;
+function getLovellVariantOfUrl(path, linkVar) {
+  return path.replace(
+    /\/lovell-federal(?:(?:-va|-tricare)?)-health-care/i,
+    getLovellUrl(linkVar),
+  );
 }
 
 module.exports = {
@@ -82,6 +73,6 @@ module.exports = {
   isListingPage,
   getLovellTitle,
   getLovellTitleVariation,
-  getLovellFormOfExistingUrl,
-  resetToFederalUrlIfNeeded,
+  getLovellUrl,
+  getLovellVariantOfUrl,
 };
--- a/src/site/stages/build/drupal/lovell/helpers.js
+++ b/src/site/stages/build/drupal/lovell/helpers.js
@@ -7,6 +7,7 @@ const LOVELL_VA_TITLE_VARIATION = 'VA';
 const LOVELL_TRICARE_TITLE_VARIATION = 'TRICARE';
 const LOVELL_VA_LINK_VARIATION = 'va';
 const LOVELL_TRICARE_LINK_VARIATION = 'tricare';
+const LOVELL_BASE_URL = '/lovell-federal-health-care';
 
 function isLovellFederalPage(page) {
   return (
@@ -34,13 +35,21 @@ function isListingPage(page) {
   return listingPageTypes.includes(page.entityBundle);
 }
 
+function getLovellUrl(urlVar) {
+  return `/lovell-federal-${urlVar}-health-care`;
+}
+
+function getLovellFormOfUrl(url, urlVar) {
+  return url.replace(LOVELL_BASE_URL, getLovellUrl(urlVar));
+}
+
 function resetToFederalUrlIfNeeded(path, variant) {
   const oppositeVariant =
     variant === 'va' ? LOVELL_TRICARE_LINK_VARIATION : LOVELL_VA_LINK_VARIATION;
-  const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;
+  const reverseUrl = getLovellUrl(oppositeVariant);
 
   if (path.includes(reverseUrl)) {
-    return path.replace(reverseUrl, `/lovell-federal-health-care`);
+    return path.replace(reverseUrl, LOVELL_BASE_URL);
   }
 
   return path;
@@ -56,9 +65,11 @@ module.exports = {
   LOVELL_TRICARE_TITLE_VARIATION,
   LOVELL_VA_LINK_VARIATION,
   LOVELL_TRICARE_LINK_VARIATION,
+  LOVELL_BASE_URL,
   isLovellFederalPage,
   isLovellTricarePage,
   isLovellVaPage,
   isListingPage,
+  getLovellFormOfUrl,
   resetToFederalUrlIfNeeded,
 };
--- a/src/site/stages/build/drupal/lovell/helpers.js
+++ b/src/site/stages/build/drupal/lovell/helpers.js
@@ -43,7 +43,7 @@ function getLovellUrl(urlVar) {
   return `/lovell-federal-${urlVar}-health-care`;
 }
 
-function getLovellFormOfUrl(url, urlVar) {
+function getLovellFormOfExistingUrl(url, urlVar) {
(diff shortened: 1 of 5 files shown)


src/site/stages/build/drupal/lovell/update-page.js
```diff
@@ -0,0 +1,125 @@
...
+
+  if (isLovellFederalPage(page)) {
+    pagePath = resetToFederalUrlIfNeeded(page.entityUrl.path, variant);
+  }
+
+  return pagePath.replace(
+    '/lovell-federal-health-care',
+    `/lovell-federal-${linkVar}-health-care`,
+  );
+}
```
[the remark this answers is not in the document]
ryguyk: **CONSIDER**
I also want to use this function as an example for another change that can likely apply across the board. I really like the thought behind streamlining and DRYing some of the computations of values needed in many of these functions. I'm not sure that passing all of them into every function via `pageVars` is the most clear way of accomplishing that, though. It's not immediately clear, in this case, for example, what needs to be passed into this function.

I think it would be more clear if we made it more explicit that the output of this function relies only on `variant`. If there are other values that are needed that can be computed from that modification variant, we can abstract that to its own function and call it from this function (rather than pass in all the related values).

```
function getLovellVariantPath(variant) {
	
  const linkVar = getLovellLinkVar(variant);

  return pagePath.replace(
    /lovell-federal(?:(?:-va|-tricare)?)-health-care/i, 
    `lovell-federal-${linkVar}-health-care`
  );
}
```
(Though, as asked above, in this case I'm not sure if `linkVar` would be any different than `variant`.)

ryguyk (commented):
This looks like some great improvement of the code. The main function is _much_ cleaner as a result of pulling out logic into smaller, focused functions. That is a pattern that should definitely be kept. 

I have added some comments in a few places about things that I think could go even further to  clean things up with this refactor. So far, these changes help _organize_ the code better, but I think there are places where the code can still be _simplified_ and _clarified_. That distinction between organization and clarity is subtle at times, to be certain, but I think it's captured here by looking at the example I point at in my commentary around the function `getLovellVariantPath`. In that case, the code is quite organized. There is a function defined whose purpose is simple and clear. However, the code within the function is not necessarily clear. It's difficult to follow and can be simplified.

You're doing some great work with this. This task is not easy. There is no clear recipe for making these sorts of changes. There's also a significant component of personal preference involved. My commentary is certainly a mix of best practices and individual preferences. My hope is that my thoughts provide some useful feedback for a bit more iteration.

I should also note that we now have the start of a testing suite for Lovell processing. As you're working on this refactor, please bear in mind that you can run the unit test for listing pages to help add some confidence that the refactor does not result in functional changes.
`yarn test:unit src/site/stages/build/drupal/tests/lovell/listingPages.unit.spec.js`

src/site/stages/build/drupal/lovell/update-page.js
```diff
@@ -0,0 +1,125 @@
...
+
+  return {
+    pageClone,
+    variant,
+    variantName,
+    linkVar:
+      variant === 'va'
+        ? LOVELL_VA_LINK_VARIATION
+        : LOVELL_TRICARE_LINK_VARIATION,
+    fieldOfficeMod:
```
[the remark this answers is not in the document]
maxx1128: `fieldOfficeMod` and `variantName` are the same, good catch. `variant` and `linkVar` are different due to the casing, and I think it's better to keep them separate for a separation of responsibilities (titles VS URLs).

src/site/stages/build/drupal/lovell/update-page.js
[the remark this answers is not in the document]
maxx1128: I agree the simplified version with the regex is more straightforward and cleaner and removes the need for the extra federal URL check. However, that function is still needed to handle the general Lovell quirkiness. Some federal pages have variant URLs that need to be replaced, or else not all Lovell pages get properly rendered. A regex can't take those cases into account since it requires looking at other page properties like `fieldAdministration`. That's why it would also need the full `page` variable passed to it and not just the URL.

I think one way to simplify this function could be to move more of the logic to the `resetToFederalUrlIfNeeded` function, since it at least moves the complexity to a file where it is read less often and can include more clarifying contents without making the rest of the code too busy. So it could be something like this:

```javascript
// update-page.js

function getLovellVariantPath(vars) {
  const { page, linkVar, variant } = vars;
  const pagePath = resetToFederalUrlIfNeeded(page, variant);

  return pagePath.replace(
    '/lovell-federal-health-care',
    `/lovell-federal-${linkVar}-health-care`,
  );
}
```

```javascript
// helpers.js

// Some federal pages already have variant paths
// but they're the wrong ones variants.
// This resets those cases back to their federal URL
// so they can get the correct variant path
// All other pages have their original path returned
function resetToFederalUrlIfNeeded(page, variant) {
  const { path } = page.entityUrl;

    const oppositeVariant =
      variant === 'va'
        ? LOVELL_TRICARE_LINK_VARIATION
        : LOVELL_VA_LINK_VARIATION;
    const reverseUrl = `/lovell-federal-${oppositeVariant}-health-care`;

    if (path.includes(reverseUrl)) {
      return path.replace(reverseUrl, `/lovell-federal-health-care`);
    }
  }

  return path;
}
```

This way `getLovellVariantPath` is easier to read, and the big function is at least elsewhere and has some comment documentation.

rmessina1010 (approved):
LGTM

src/site/stages/build/drupal/lovell/helpers.js
```diff
@@ -0,0 +1,78 @@
...
+    : LOVELL_TRICARE_TITLE_VARIATION;
+}
+
+function getLovellUrl(linkVar) {
+  return `/lovell-federal-health-care-${linkVar}`;
+}
+
+function getLovellVariantOfUrl(path, linkVar) {
+  return path.replace(
+    /\/lovell-federal(?:(?:-va|-tricare)?)-health-care/i,
```
ryguyk, line 54: This regex is going to need to change to accommodate the new naming:

I think this should work, though I have not yet tested.
`/\/lovell-federal-health-care(?:(?:-va|-tricare)?)/i`

ryguyk (changes requested):
Made one discovery. Still looking through the rest.

src/site/stages/build/drupal/lovell/update-page.js
```diff
@@ -0,0 +1,87 @@
...
+  return getLovellVariantOfUrl(page.entityUrl.path, oppositeVariant);
+}
+
+function getLovellBreadcrumbs(vars) {
+  const { page, variantName, linkVar } = vars;
+  // Modify Breadcrumb
+  return page.entityUrl.breadcrumb.map(crumb => {
+    // eslint-disable-next-line no-param-reassign
+    crumb.text = crumb.text.replace(
+      /Lovell Federal (VA )?health care/,
```
ryguyk, line 53: I'm not sure if this pattern with VA in the middle actually exists in the breadcrumb content coming from Drupal, but, if it does, it's likely changing based on the naming change. It'd likely be at the end now (i.e. `Lovell Federal health care - VA`).

We should either remove this VA piece of the regex (if it's not needed), or update it to what it needs to be.

ryguyk (changes requested):
A couple other thoughts.

src/site/stages/build/drupal/lovell/helpers.js
ryguyk: Further explanation:

New url example: `lovell-federal-health-care-tricare/some-page`

So `getLovellVariantOfUrl` is going to ultimately do this:
```
'lovell-federal-health-care-tricare/some-page'.replace(
    /\/lovell-federal(?:(?:-va|-tricare)?)-health-care/i,
    getLovellUrl(linkVar),
  );

```
The regex is going to match the first part of the string `lovell-federal-health-care` and replace that with `lovell-federal-health-care-tricare` so the resulting url will be `lovell-federal-health-care-tricare-tricare/...`

ryguyk: Tested here a bit: https://jsfiddle.net/x95d1nu2/

ryguyk (approved):

PR #1431, opened 2023-01-07
VACMS-11997: Adds unit tests for lovell sidebar locations

ryguyk:
## Description
Adds unit tests to confirm existence and proper sorting of facility-location links in the sidebar menus on Lovell pages.

The structure here is as follows:
- Added some fixture code to generate sidebar data for various cases (i.e. edge cases).
- Generate sidebar data to plug into an otherwise empty `drupalData` object to test only this sidebar data.
- Call `processLovellPages` for each test case and confirm the generated sidebar is as expected (essentially, the initial data moved into new sidebar properties under `drupalData.data`.

closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11997
relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11650

## Testing done & Screenshots
This PR introduces some new tests. They pass locally and need to pass in the review instance once this is built.

## QA steps

1. Pull branch and run this spec locally:
   - [ ] `yarn test:unit src/site/stages/build/drupal/tests/lovell/sidebarLocations.unit.spec.js`
1. Some small changes were made to move existing code to a shared location, so run this spec locally and confirm it passes as well: 
   - [ ] `yarn test:unit src/site/stages/build/drupal/tests/lovell/listingPages.unit.spec.js`
1. Confirm tests pass on CI
   - [ ] All tests should pass, but specifically confirm the unit tests added here do not fail in CI.

### DIFF
--- a/src/site/stages/build/drupal/tests/lovell/fixtures/generateSidebar.js
+++ b/src/site/stages/build/drupal/tests/lovell/fixtures/generateSidebar.js
@@ -1,0 +1,55 @@
+import cloneDeep from 'lodash/cloneDeep';
+import sidebarTemplate from './sidebar.json';
+import { findSidebarMenuLinkBySectionAndOptionalLabel } from '../utils';
+
+/**
+ * Generates an individual location menu link
+ */
+const generateSidebarLocation = (lovellVariant, label, urlString) => ({
+  label: `${lovellVariant.toUpperCase()} Location ${label}`,
+  url: {
+    path: `/lovell-federal-${lovellVariant}-health-care/locations/${urlString}`,
+  },
+  entity: {
+    fieldMenuSection: lovellVariant,
+  },
+  links: [],
+});
+
+/**
+ * Generates a sidebar menu tree for testing
+ */
+export default (vaLocationCount, tricareLocationCount) => {
+  const sidebar = cloneDeep(sidebarTemplate);
+
+  const lovellFederal = findSidebarMenuLinkBySectionAndOptionalLabel(
+    sidebar.links,
+    'both',
+  );
+
+  const servicesAndLocations = findSidebarMenuLinkBySectionAndOptionalLabel(
+    lovellFederal.links,
+    'both',
+    'Services and Locations',
+  );
+
+  const vaLocations = findSidebarMenuLinkBySectionAndOptionalLabel(
+    servicesAndLocations.links,
+    'va',
+    'Locations',
+  );
+  vaLocations.links = [...Array(vaLocationCount)].map((item, i) =>
+    generateSidebarLocation('va', i + 1, `va-location-${i + 1}`),
+  );
+
+  const tricareLocations = findSidebarMenuLinkBySectionAndOptionalLabel(
+    servicesAndLocations.links,
+    'tricare',
+    'Locations',
+  );
+  tricareLocations.links = [...Array(tricareLocationCount)].map((item, i) =>
+    generateSidebarLocation('tricare', i + 1, `tricare-location-${i + 1}`),
+  );
+
+  return sidebar;
+};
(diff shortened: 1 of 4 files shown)


src/site/stages/build/drupal/tests/lovell/fixtures/generateSidebar.js
```diff
@@ -0,0 +1,55 @@
...
+  );
+
+  const tricareLocations = findSidebarMenuLinkBySectionAndOptionalLabel(
+    servicesAndLocations.links,
+    'tricare',
+    'Locations',
+  );
+  tricareLocations.links = [...Array(tricareLocationCount)].map((item, i) =>
+    generateSidebarLocation('tricare', i + 1, `tricare-location-${i + 1}`),
+  );
```
maxx1128, line 52: **Suggestion:** could the logic that makes `vaLocations` and `tricareLocations` be separated into another util function? Or the map function used here could be written separately in this function.

src/site/stages/build/drupal/tests/lovell/sidebarLocations.unit.spec.js
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable @department-of-veterans-affairs/axe-check-required */
+import { expect } from 'chai';
+import { processLovellPages } from '../../process-lovell-pages';
+import {
+  stringArraysContainSameElements,
+  findSidebarMenuLinkBySectionAndOptionalLabel,
+} from './utils';
+import generateSidebar from './fixtures/generateSidebar';
+
+const getSidebarProperty = (drupalData, lovellVariant) =>
```
maxx1128, line 10: **Question:** should this be renamed `getSidebarQuery`?

src/site/stages/build/drupal/tests/lovell/fixtures/generateSidebar.js
maxx1128, line 52: **Suggestion:** could the logic that makes vaLocations and tricareLocations be separated into another util function? Or the map function used here could be written separately in this function.

maxx1128 (commented):
The tests look good and pass for me locally, I just have a few small questions/suggestions.

src/site/stages/build/drupal/tests/lovell/fixtures/generateSidebar.js
ryguyk: Thanks for the suggestion. I added a small abstraction. It should be noted that I ended up with a structure wherein `.map()` is still passed an inline function.
```
tricareLocations.links = [...Array(tricareLocationCount)].map((_item, i) =>
   generateTricareSidebarLocation(i + 1),
);
```

That newly created function then takes only a `label` as a parameter.
```
const generateTricareSidebarLocation = label =>
  generateSidebarLocation('tricare', label, `tricare-location-${label}`);
```

It _could_ have been accomplished by defining `generateTricareSidebarLocation` to accept `(item, i)`, which would then allow us to simply pass that function to the `.map()` call. I do not personally care for that approach in this instance because I think that would make the `generateTricareSidebarLocation` function less clear. The function's purpose is to generate a Tricare sidebar location with a specific label, and it should be able to do that outside of the context of being called from a `.map()` function. It only needs to be passed the label, and should be able to do its work as long as that is provided. If we defined it as considered here, it would require a signature that was more specific to being called from the `.map()` (i.e. `(item, i)`), and that feels less clear and useful to me, personally.

src/site/stages/build/drupal/tests/lovell/sidebarLocations.unit.spec.js
ryguyk: Great idea! I considered this when I first wrote it, but I opted away from it because it's not actually a query. It's a generated _property_ on the `drupalData.data` object. That said, the property has a name that includes `query`, and I understand that is done to match the initial federal-level property name. Given the idea/confusion here, I find that to be a strong argument for naming it as you suggest. I've made this change.

maxx1128 (approved):
Thank you for making the changes and I appreciate the extra discussion for the additional changes. This pull request looks good to go 👍

PR #1432, opened 2023-01-10
Updates CMS CA cert references.

ndouglas:
See department-of-veterans-affairs/va.gov-cms#12206.  This is intended to fix issues in CI, so if it passes CI it should be safe to merge.

Replaced every reference to the old CA with a reference to the new CA.  Let's see if that works 🤔 

This isn't sufficient; more work will be required.


.devcontainer/Dockerfile
```diff
@@ -42,7 +42,8 @@ RUN aws --version # Verify AWS CLI installation.
 ENV AWS_CA_BUNDLE /etc/ssl/certs/ca-certificates.crt
 
 # Add VA Root CA to Docker Certificate Authority (CA) Store so that NODE can use it for requests.
-ADD http://crl.pki.va.gov/PKI/AIA/VA/VA-Internal-S2-RCA1-v1.cer /usr/local/share/ca-certificates/
+ADD https://raw.githubusercontent.com/department-of-veterans-affairs/platform-va-ca-certificate/main/VA-Internal-S2-RCA1-v1.cer /usr/local/share/ca-certificates/
```
timcosgrove, line 45: Is there a reason the specification of the destination is different here?

ndouglas: Destination?

```
ADD https://raw.githubusercontent.com/department-of-veterans-affairs/platform-va-ca-certificate/main/VA-Internal-S2-RCA2.cer /usr/local/share/ca-certificates/VA-Internal-S2-RCA2.crt
```

is thus because it's also changing the file extension, if that's whatcha mean.

PR #1440, opened 2023-01-19
Update self-signed certs pointed to by various content build workflows.

timcosgrove:
## Description
Many of the workflows in content-build use private certs to connect to VA-internal servers, specifically to the CMS servers. Examples of these servers include:

* https://prod.cms.va.gov/
* https://main-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov/ ('Tugboat' server).

We make these certs available to self-hosted runners we use for Github Actions primarily so that we can specify the private cert to Node.js applications, so that those Node.js can make requests to our VA-internal servers without throwing self-signed cert errors (via the `NODE_EXTRA_CA_CERTS` env variable).

The certs used on Tugboat instances (CMS preview environments) expired recently and were updated. This requires pointing the workflows at that new private cert. However, often a given workflow will point at a different internal server which is still using an older cert.

The AMIs that build our GHA runners have been updated with the new cert, and existing runners have also been updated. What remains is to point the workflows at the updated certs.

### What the PR does
In most cases the Node scripts were pointing at a specific cert. However, on all these instances there is a general combined cert at `/etc/ssl/certs/ca-certificates.crt` that is a combined cert consolidating all private certs. In almost all cases we pointed at that.

The case for the Continuous Integration Cypress testing job is unique. That job is run on a Docker container sourced from Cypress.io. This container cannot contain our private certs, obviously. Moreover, because the individual job runs on the Docker container, it doesn't have access to the private certs on our EC2 instances. Therefore, the repo contains a copy of the necessary cert, and moves that cert into place in the Docker container so that it can be used there. This is also why the updated cert file is added to this PR (bottom of file list).

Finally, 2 weeks ago we pointed some of the workflows at https://staging.cms.va.gov/ for their CMS data endpoint, because the Tugboat certs had changed, and the existing certs still worked with Staging. This PR also shifts those workflows back to using the Tugboat-based endpoint (https://main-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov).

## Testing done
These changes working when accessing the CMS was tested via a testing workflow, viewable here: https://github.com/department-of-veterans-affairs/content-build/actions/workflows/content-release-test.yml

Despite them being marked as failed, what we are looking for here is that the Build step was able to complete successfully, as that is the step that uses the CMS endpoint and therefore needs the private cert. An example of that is here: https://github.com/department-of-veterans-affairs/content-build/actions/runs/3962035298

Additionally, the changes working for the CI workflow and specifically for the Cypress testing job can be validated here, as this is the CI job for this PR, which completed successfully: https://github.com/department-of-veterans-affairs/content-build/actions/runs/3962314238


.github/workflows/a11y-heading-order.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     env:
       DRUPAL_ADDRESS: https://cms-content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov
-      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/VA-Internal-S2-RCA1-v1.cer.pem
+      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/ca-certificates.crt
```
timcosgrove, line 24: On instances derived from the AMI, `/etc/ssl/certs/ca-certificates.crt` combines all certs. Pointing almost all jobs at the combined cert.

.github/workflows/continuous-integration.yml
```diff
@@ -34,19 +34,20 @@ jobs:
 
     env:
       # Sandbox Drupal address, username, and password is used on branches other than main.
-      DRUPAL_ADDRESS: https://staging.cms.va.gov
+      DRUPAL_ADDRESS: https://main-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov
```
timcosgrove, line 37: This undoes the change made last week to allow CI to pass. With the new certs in place, this should be ok now.

.github/workflows/continuous-integration.yml
```diff
@@ -361,7 +362,7 @@ jobs:
         - /etc/ssl/certs
 
     env:
-      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/VA-Internal-S2-RCA1-v1.pem
+      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/VA-Internal-S2-RCA-combined.pem
```
timcosgrove, line 365: This is only for Cypress tests. We cannot use ca-certificates.crt here because the job does not have access to the AMI's filesystem.

olivereri (approved):

holdenhinkle (approved):
No one on the QAS team has the expertise to review this PR properly. Everything _looks_ right. I feel more confident approving because Eric Oliver approved. Approved 👍

rmessina1010 (approved):
LGTM

PR #1441, opened 2023-01-20
Makes CMS ready-check use same host address as build.

ndouglas:
## Description

Closes department-of-veterans-affairs/va.gov-cms#12289.

### Before

Screen Shot 2023-01-24 at 2 02 45 PM

### After

Screen Shot 2023-01-24 at 1 56 39 PM

## QA Steps
Verify that on CI builds that the CMS-ready-check hostname matches that used in the build.


rmessina1010 (approved):
LGTM

timcosgrove (approved):
I would get a revert ready for this one just in case, but, I'm pretty sure the logic here checks out.

PR #1443, opened 2023-01-23
Revert "Update self-signed certs pointed to by various content build workflows."

timcosgrove:
Reverts department-of-veterans-affairs/content-build#1440.

This is a 'back-pocket' revert; we are keeping this ready in the event that something is not working as anticipated and we need to roll back.


rmessina1010 (approved):
LGTM

PR #1452, opened 2023-01-27
Specify ASG label for all self-hosted that is not on-demand.

timcosgrove:
## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11402

We are looking to isolate the runners which run the Continuous Integration:Cypress Tests job from all other jobs.

#1450 moved the Cypress Tests job to use its own on-demand runner. Previously it had specified the `asg` label, picking a runner from the ASG pool.

This PR moves any job running self-hosted to use the `asg` label. This should prevent all these jobs from ever using the same runner as the Cypress Types job.

## Acceptance Criteria
- [ ] All workflows/jobs in this PR run without error. Many may need to be tested after merge, unfortunately.

### DIFF
--- a/.github/workflows/a11y-heading-order.yml
+++ b/.github/workflows/a11y-heading-order.yml
@@ -122,7 +122,7 @@ jobs:
     name: Accessibility Tests
     needs: build
     timeout-minutes: 120
-    runs-on: self-hosted
+    runs-on: [self-hosted, asg]
     container:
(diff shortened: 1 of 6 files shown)


olivereri (approved):

rmessina1010 (approved):
LGTM

rbeckwith-oddball (approved):
Pretty straight-forward changes. Looks good to go.

PR #1453, opened 2023-01-27
Revert "Makes CMS ready-check use same host address as build."

ndouglas:
Reverts department-of-veterans-affairs/content-build#1441

Just in case.


timcosgrove (approved):

ndouglas:
Don't think this is needed.

PR #1475, opened 2023-02-14
VACMS-12068: Updates header levels on service-location template.

ryguyk:
## Description
Subheadings on VAMC facility service locations are inconsistent. "Contact Information" is an h4, and everything underneath that can logically be considered as a subheading beneath that h4, so they should all be h5.

Importantly, the template used here is also used elsewhere. The other place it is currently being used is to render non-clinical services (i.e. Billing and Insurance). On the non-clinical services pages, this information is presented slightly differently, so the header levels are not the same. 

For this PR to be valid, we need to ensure that the health-service accordions render correctly **and** that there is no regression in the presentation of non-clinical services.

### IMPORTANT
This PR introduces a slight (desired) change to some conditional display of headers. Previously, there was a condition under which some of these subheaders would display differently than others because it was implemented in a way that treated some of these subheaders as having another level of nesting (hours, phone). This was removed, as the desired behavior in both places is to have all subheaders as children of either "Contact Information" (health services) or of the VAMC Facility name (non-clinical services). This change makes the code significantly cleaner.

closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12068

## Testing done & Screenshots
Local testing.

### Assures h5 subheadings on VAMC Facility health service accordions:
http://039c47a435072b98659a25b04edf11f7.review.vetsgov-internal/boston-health-care/locations/jamaica-plain-va-medical-center/
image

### Assures h4 subheadings on VAMC top-task non-clinical service pages:
http://039c47a435072b98659a25b04edf11f7.review.vetsgov-internal/boston-health-care/billing-and-insurance/
image

## QA steps
- [ ] Verify [here ](http://039c47a435072b98659a25b04edf11f7.review.vetsgov-internal/boston-health-care/locations/jamaica-plain-va-medical-center/)that VAMC Facility health-service accordions have h5 subheadings under "Contact Information".

- [ ] Verify [here](http://039c47a435072b98659a25b04edf11f7.review.vetsgov-internal/boston-health-care/billing-and-insurance/) that VAMC System top-task pages (i.e. Billing and Insurance) pages have h4 subheadings under the h3 header showing the VAMC facility name where the service is offered.

### DIFF
--- a/src/site/components/phone-number.drupal.liquid
+++ b/src/site/components/phone-number.drupal.liquid
@@ -4,11 +4,7 @@
 {% if phoneLabel == empty %}
   {% assign phoneLabel = 'Phone' %}
 {% endif %}
-{% if haveLocationName != empty %}
-  <h4>{{ phoneLabel }}</h4>
-{% else %}
-  <h3 class="force-small-header">{{ phoneLabel }}</h3>
-{% endif %}
+<h{{ phoneHeaderLevel }}>{{ phoneLabel }}</h{{ phoneHeaderLevel }}>
 <a class="vads-u-margin-bottom--1" aria-label="{{ number.fieldPhoneNumber | accessibleNumber }}{% if number.fieldPhoneExtension %} extension: {{ number.fieldPhoneExtension | accessibleNumber }}{% endif %}"
    href="tel:{{ number.fieldPhoneNumber }}{% if number.fieldPhoneExtension %}p{{ number.fieldPhoneExtension }}{% endif %}">
   {{ number.fieldPhoneNumber }}{% if number.fieldPhoneExtension %}x {{ number.fieldPhoneExtension }}{% endif %}
--- a/src/site/paragraphs/service_location.drupal.liquid
+++ b/src/site/paragraphs/service_location.drupal.liquid
@@ -1,11 +1,7 @@
-{% assign serviceLocationAddress = single.fieldServiceLocationAddress.entity %}
-{% if serviceLocationAddress.fieldClinicName %}
-  {% assign haveLocationName = true %}
-{% else %}
-  {% assign haveLocationName = false %}
-{% endif %}
-
-{% include "src/site/facilities/service_address.drupal.liquid" %}
+{% include "src/site/facilities/service_address.drupal.liquid" with
+  serviceLocationAddress = single.fieldServiceLocationAddress.entity
+  serviceLocationAddressHeaderLevel = serviceLocationSubHeaderLevel
+%}
 
 {% comment %}
   single.fieldHours values:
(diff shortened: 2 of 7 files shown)


src/site/components/phone-number.drupal.liquid
```diff
@@ -4,11 +4,7 @@
 {% if phoneLabel == empty %}
   {% assign phoneLabel = 'Phone' %}
 {% endif %}
-{% if haveLocationName != empty %}
-  <h4>{{ phoneLabel }}</h4>
-{% else %}
-  <h3 class="force-small-header">{{ phoneLabel }}</h3>
-{% endif %}
+<h{{ phoneHeaderLevel }}>{{ phoneLabel }}</h{{ phoneHeaderLevel }}>
```
ryguyk, line 7: Explanation:

A conditional difference in header levels is not correct. This should always be the same, and should be determined by the calling template.

src/site/paragraphs/service_location.drupal.liquid
```diff
@@ -15,11 +11,7 @@
 {% endcomment %}
 
 {% if single.fieldHours != "1" %}
-  {% if haveLocationName %}
-    <h4 data-template="paragraphs/service_location">Hours</h4>
-  {% else %}
-    <h3 class="force-small-header" data-template="paragraphs/service_location">Hours</h3>
-  {% endif %}
+  <h{{ serviceLocationSubHeaderLevel }} data-template="paragraphs/service_location">Hours</h{{ serviceLocationSubHeaderLevel }}>
```
ryguyk, line 14: Explanation:

Similar to other comment of the same nature, a conditional difference in header levels is not correct here. This should always be the same, and should be determined by the calling template. This should not be considered a subheading beneath the Location name, so its heading level should not be dependent on it.

laflannery:
Confirmed VAMC Facility health-service accordions (both with and without Facility names) have h5 subheadings under "Contact Information".

Confirmed no regressions on VAMC System top-task pages - Checked Billing and Insurance & Register for care (Policies I believe is a Top Task page but doesn't follow this template from what I saw)

maxx1128 (approved):
I checked the code and the live review pages, the header levels all check out 👍

timcosgrove (approved):
Approved; I'll defer to @ryguyk and @ndouglas re: what sort of failure state we would want in the event that say `phoneHeaderLevel` is not provided.

ryguyk:
Some [nice engineering discussion here](https://dsva.slack.com/archives/CT4GZBM8F/p1676392340276029?thread_ts=1676391997.915989&cid=CT4GZBM8F) around whether or not these templates should provide default values for the heading levels. Ultimately decided no, so merging as is.

PR #1479, opened 2023-02-17
[VA-12570] Update COVID status field with facility-specific supplemental info

maxx1128:
## Description

Facilities can have COVID info specific to them that should be displayed instead. This pull request adds the extra GraphQL query to the ones with COVID precaution info and shows it in place of the regular COVID info if it's present.

Closes [#12570](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12570)
Blocked by [#12403](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/12403) since it relies on new CMS data that the GraphQL queries need.

## Testing done & Screenshots

Checking some facilities locally and making sure that the facility COVID data still shows up as expected.

Screen Shot 2023-02-17 at 5 03 28 PM

## QA steps

Build the repo locally with `yarn build --pull-drupal --drupal-address=https://pr12403-yjq5zexhavmhassuj4kctou3kauqjue2.ci.cms.va.gov/`

1. The new COVID status data is queried in needed facilities pages
   - [ ] Check that the data is present on the page locally in the console
2. Confirm that it uses the new COVID status data when available
   - [ ] Check that the new data is used when it's present on a page
   - [ ] Check that when the new data isn't present, the old COVID status data is used

## Acceptance criteria

- [ ] All QA steps are met and passed

## Definition of done

### DIFF
--- a/src/site/includes/covid-status.drupal.liquid
+++ b/src/site/includes/covid-status.drupal.liquid
@@ -10,6 +10,10 @@ class="vads-u-margin-top--{{topMargin}} vads-u-margin-bottom--{{bottomMargin}} v
 			'covid-status-heading':  '{{ fieldSupplementalStatus.entity.fieldStatusId }}',
 		});"
 	>
-		{{ fieldSupplementalStatus.entity.description.processed }}
+		{% if fieldSupplementalStatusMoreI.processed %}
+			{{ fieldSupplementalStatusMoreI.processed }}
+		{% else %}
+			{{ fieldSupplementalStatus.entity.description.processed }}
+		{% endif %}
 	</va-alert-expandable>
 {% endif %}
--- a/src/site/stages/build/drupal/graphql/facilities-fragments/healthCareLocalFacility.node.graphql.js
+++ b/src/site/stages/build/drupal/graphql/facilities-fragments/healthCareLocalFacility.node.graphql.js
@@ -30,6 +30,11 @@ const FACILITIES_RESULTS = `
           }
         }
       }
+      fieldSupplementalStatusMoreI {
+        value
+        format
+        processed
+      }
       fieldOperatingStatusFacility
       fieldFacilityLocatorApiId
       fieldIntroText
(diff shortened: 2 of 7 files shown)


src/site/includes/covid-status.drupal.liquid
```diff
@@ -10,6 +10,10 @@ class="vads-u-margin-top--{{topMargin}} vads-u-margin-bottom--{{bottomMargin}} v
 			'covid-status-heading':  '{{ fieldSupplementalStatus.entity.fieldStatusId }}',
 		});"
 	>
-		{{ fieldSupplementalStatus.entity.description.processed }}
+		{% if fieldSupplementalStatusMoreI.processed %}
```
ryguyk, line 13: **SHOULD**

I think this might be more accurate if we explicitly check for an empty string:
```
{% if fieldSupplementalStatusMoreI.processed != empty %}
```
As is, the fallback would be used for an undefined `processed` property, but the fallback would not be used for a `processed` property that is an empty string. The CMS requires this field, so it's almost certainly not going to be empty, but I still think we should be a bit more defensive.

ryguyk (changes requested):

src/site/includes/covid-status.drupal.liquid
ryguyk: @maxx1128 **UPDATE: What I wrote above is incorrect.**

I just spent some time studying this, and I clarified my understanding. The version of Liquid that we're using (TinyLiquid) does not handle truthy/falsey values in the same way as the more well-known Shopify version of Liquid.  In the latter, an empty string is truthy. In our version, an empty string is falsey, so simply checking `{% if fieldSupplementalStatusMoreI.processed %}` as you originally had it is correct. 

`empty` equates to `null`, so explicitly checking for `{% if fieldSupplementalStatusMoreI.processed != empty %}` would actually have the effect that I was trying to avoid. Since an empty string is _not_ null, this check would be true, and it would print the empty string rather than default to the fallback.

As mentioned, in this case, since the value is required in the CMS, we should always get a value that is defined and non-empty, so both checks would be equivalent. But they aren't equivalent in all cases, and that's something we should be aware of. 

My apologies for steering us in the wrong direction. I wanted to follow up here for two reasons:
1. To share my findings and be transparent about my error.
2. **I think we should probably change this back before merging, just to be thorough.**

FWIW, I'm not sure this is implemented correctly in every place in our codebase. I'm not sure how often it would cause errors, but it's interesting to keep an eye out for.

PR #1480, opened 2023-02-21
Stop Workflows From Using Root Permissions in Cypress Container

olivereri:
## Description

Fixes the need to run Cypress containers as root in the below GHA workflows;

* Accessiblity Tests
* Accessibility Tests - Heading Order
* Continuous Integration

When these containers are run as root it changes the runner user (UID 1001) home directory files to root ownership.  Other workflow jobs that use these runners fail to execute because they fail due to permission denied errors. Cypress containers DO NOT need to run as root. Adding the option to run as the runner user (UID 1001) `-u 1001:1001` ensures that container mounted volumes retain the host user file permissions.

Continuous Integration GHA workflow has been migrated away from using On-demand GHA runners and is directed to use AWS Auto-Scaling Group GHA runners. On-demand runners were configured to run as root and switching to ASG GHA runners obviates that.

Raw Slack discussion here:
https://dsva.slack.com/archives/CT4GZBM8F/p1677089769027479

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/50148  
relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12435

## Testing done & Screenshots

image
image

## QA steps

Continuous Integration tests pass
Continuous Integration tests pass, Accessibility Tests and Heading Order tests continue to function normally.

## Acceptance criteria

- [ ] Continuous Integration, Accessibility Tests, and Accessibility Tests Heading Order Cypress containers do NOT run as root.
- [ ] Continuous Integration, Accessibility Tests, and Accessibility Tests Heading Order Tests function normally.

## Definition of done
- [ ] This PR is merged.


.github/workflows/a11y-testing.yml
```diff
@@ -0,0 +1,260 @@
+#name: Accessibility Tests
+#
+#on:
+#  workflow_dispatch:
+#  schedule:
+#    - cron: '0 12 * * 1-5'
+
+name: ZZ Accessibilty Tests
```
ndouglas, line 8:
```suggestion
name: Accessibility Tests
```

olivereri: Using Tim's method of testing content-build related workflows:
https://dsva.slack.com/archives/CT4GZBM8F/p1674156682111419?thread_ts=1674149076.806359&cid=CT4GZBM8F

I know we've discussed creating a `content-build-test` repository and all the GHA infrastructure that goes along with it, but that comes with a host of it's own problems.  Namely protecting Test workflows from releasing content to production S3 buckets. Thing is we want mirror Content-Build, so periodically overwrite Content-Build-Test with the former's current state. That's all fine and good for CMS and CMS-Test but for CB destination buckets are set as ENV vars. Also Slack Notifications become a huge issue.  Until we can find a safe way to resolve that we'll need to rely on these janky ZZ workflows that live in their own feature branches.

pic unrelated
image

.github/workflows/continuous-integration.yml
```diff
@@ -409,29 +352,28 @@ jobs:
...
       CHROMEDRIVER_FILEPATH: /share/chrome_driver/chromedriver
 
     steps:
       - name: Checkout content-build
         uses: actions/checkout@v3
 
       # Required for Docker
-      - name: Move VA cert to /etc/ssl/certs
-        run: mv certs/VA-Internal-S2-RCA-combined.pem /etc/ssl/certs/
+      #- name: Move VA cert to /etc/ssl/certs
```
timcosgrove, line 375: If we don't need these, should we remove them?

timcosgrove (commented):
One comment

.github/workflows/continuous-integration.yml
olivereri: You're right, will do.

timcosgrove (approved):

.github/workflows/a11y.yml
```diff
@@ -159,7 +159,7 @@ jobs:
     runs-on: [self-hosted, asg]
     container:
       image: public.ecr.aws/cypress-io/cypress/browsers:node16.13.2-chrome100-ff98
-      options: -u 1001:1001 -v /usr/local/share:/share --user root
+      options: -u 1001:1001 -v /usr/local/share:/share
```
ndouglas, line 162: might be nice to change the flags here to `--user` and `--volume` here, because we really should've noticed earlier that we were specifying the same argument twice

ndouglas: /channeling inner Elijah

olivereri:
```
diff --git a/.github/workflows/a11y-heading-order.yml b/.github/workflows/a11y-heading-order.yml
index 5c0fda9ac..f058f2699 100644
--- a/.github/workflows/a11y-heading-order.yml
+++ b/.github/workflows/a11y-heading-order.yml
@@ -125,7 +125,7 @@ jobs:
     runs-on: [self-hosted, asg]
     container:
       image: public.ecr.aws/cypress-io/cypress/browsers:node16.13.2-chrome100-ff98
-      options: -u 1001:1001 -v /usr/local/share:/share
+      options: --user 1001:1001 --volume /usr/local/share:/share
     strategy:
       fail-fast: false
       max-parallel: 32
(1/1) Stage this hunk [y,n,q,a,d,e,?]? y

diff --git a/.github/workflows/a11y.yml b/.github/workflows/a11y.yml
index 4121b6b15..be9adeb0d 100644
--- a/.github/workflows/a11y.yml
+++ b/.github/workflows/a11y.yml
@@ -159,7 +159,7 @@ jobs:
     runs-on: [self-hosted, asg]
     container:
       image: public.ecr.aws/cypress-io/cypress/browsers:node16.13.2-chrome100-ff98
-      options: -u 1001:1001 -v /usr/local/share:/share
+      options: --user 1001:1001 --volume /usr/local/share:/share
     strategy:
       fail-fast: false
       max-parallel: 32
(1/1) Stage this hunk [y,n,q,a,d,e,?]? y

diff --git a/.github/workflows/continuous-integration.yml b/.github/workflows/continuous-integration.yml
index 1591db5a6..819494cd6 100644
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -358,7 +358,7 @@ jobs:
     timeout-minutes: 30
     container:
       image: public.ecr.aws/cypress-io/cypress/browsers:node16.13.2-chrome100-ff98
-      options: -u 1001:1001
+      options: --user 1001:1001
       volumes:
         - /usr/local/share:/share
         - /etc/ssl/certs:/etc/ssl/certs
(1/1) Stage this hunk [y,n,q,a,d,e,?]? y
```

image

ndouglas: gladiator-thumbsup

ndouglas (approved):

PR #1486, opened 2023-02-27
VA-12686: Hide "Manage Your Health" for Tricare pages

maxx1128:
## Description

On Lovell, the "Manage your Health Online" section was removed from Tricare pages, but it still appears on Tricare pages that were formerly federal pages. This updates the check so it also excludes any pages with `tricare` in the URL, which is always the case for them due to the process that makes federal pages into VA and Tricare pages.

closes [#12686](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12686)

## Testing done & Screenshots

Locally, see below for VA and Tricare page comparisons.

Screen Shot 2023-02-27 at 12 54 17 PM

Screen Shot 2023-02-27 at 12 54 08 PM

## QA steps

1. Check the main Tricare page
   - [ ] "Manage Your Health Online" does not appear
1. Check the main VA page
   - [ ] "Manage Your Health Online" does appear

## Acceptance criteria
- [ ] "Manage Your Health Care" no longer appears on any Tricare pages.

## Definition of done

### DIFF
--- a/src/site/layouts/health_care_region_page.drupal.liquid
+++ b/src/site/layouts/health_care_region_page.drupal.liquid
@@ -52,89 +52,89 @@
 
           <!-- "Manage your health online" section -->
           {% comment %} Hide this section for Lovell TRICARE {% endcomment %}
-          {% if fieldAdministration.entity.entityId != '1039' %}
-          <section>
-            {% if fieldAdministration.entity.entityId == '1040' %}
-              <h3>Manage your VA health online</h3>
-            {% else %}
-              <h3>Manage your health online</h3>
-            {% endif %}
-            <div
-              class="vads-u-display--flex medium-screen:vads-u-flex-direction--row vads-u-flex-direction--column">
+          {% unless fieldAdministration.entity.entityId == '1039' or entityUrl.path contains '-tricare' %}
+            <section>
+              {% if fieldAdministration.entity.entityId == '1040' %}
+                <h3>Manage your VA health online</h3>
+              {% else %}
+                <h3>Manage your health online</h3>
+              {% endif %}
               <div
-                class="vads-u-margin-right--0 medium-screen:vads-u-margin-right--3">
-                <div class="vads-facility-hub-cta">
-                  <a onClick="recordEvent({ event: 'cta-hub-link-click' });"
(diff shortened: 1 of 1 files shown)


src/site/layouts/health_care_region_page.drupal.liquid
```diff
@@ -52,89 +52,89 @@
...
-          {% if fieldAdministration.entity.entityId != '1039' %}
-          <section>
-            {% if fieldAdministration.entity.entityId == '1040' %}
-              <h3>Manage your VA health online</h3>
-            {% else %}
-              <h3>Manage your health online</h3>
-            {% endif %}
-            <div
-              class="vads-u-display--flex medium-screen:vads-u-flex-direction--row vads-u-flex-direction--column">
+          {% unless fieldAdministration.entity.entityId == '1039' or entityUrl.path contains '-tricare' %}
```
ryguyk, line 55: #### Good news
I think this new logic to include `entityUrl.path contains '-tricare'` will likely fix this issue. 

#### Not so good news
I'm not sure this is my preferred way to handle this, but it might be the only way without changing some things in the CMS.  It seems that checking the `entityId` might not be a solution if these pages are set to have a `fieldAdministration` reflecting the Federal level (that is to say, both the Tricare _and_ VA pages are coming over with a value of `347` for that property) . 

So, I'm not sure we need to keep the part that checks for `fieldAdministration.entity.entityId == '1039'`. It seems these pages are not set up that way, and this check will always be false.

Additionally, though not specifically within the scope of the ticket in question, the check for `1040` below will also do nothing. Please see comment on that line.

ryguyk (changes requested):

src/site/layouts/health_care_region_page.drupal.liquid
maxx1128: This is a good catch, and the lack of a more reliable way to tell variant pages apart from Federal ones has made things difficult in the past.  To solve this, maybe as part of the Lovell page processing, we could add another property that specifically marks it as a Lovell or VA variant? Something along these lines in the `getModifiedLovellPage` function:

```javascript
page.variant = {
    va: variant === "va",
    tricare: variant === "tricare"
}
```

Then the liquid template could just be something like this:

```liquid
{% if variant.va %}
    ...
{% endif %}
```

PR #1487, opened 2023-02-27
Rebuilt header plugin for accordion headers

jtmst:
## Description

closes 12342

## Testing done & Screenshots
Local testing on various headers.  

## QA steps

- Pull down branch
- Rebuild content-build
- verify various links from the accordion heading QA dock: https://docs.google.com/spreadsheets/d/1j1FgW1D78gFpVjPYTjsd0DpiDfCATYUf9xG_09-BeOQ/edit#gid=132464158

2. Then validate Acceptance Criteria from issue

## Acceptance criteria

- [ ] Heading levels are more accurate in most cases, and have a fallback for edge cases where predicting the level isnt possible

## Definition of done

### DIFF
--- a/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
+++ b/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
@@ -29,9 +29,9 @@ module.exports = {
 
     const findPreviousHeaderLevel = node => {
       // Returns the header level when the last header is in a wysiwyg div
-      const previousHeaderLevelParent = dom(node).prev(
-        'div[data-template="paragraphs/wysiwyg"]',
-      );
+      const previousHeaderLevelParent = dom(node)
+        .prevAll('div[data-template="paragraphs/wysiwyg"]:has(:header)')
+        .first();
 
       const lastHeader = previousHeaderLevelParent.find(':header').last()[0];
       if (lastHeader) {
--- a/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
+++ b/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
@@ -62,13 +62,17 @@ module.exports = {
       let incrementedHeaderLevel =
         parseInt(previousHeaderLevel?.substring(1), 10) + 1;
       const accordionHeaders = dom(node).find(':header');
+      // Ensures that only the main header is affected and not headers in the accordion content
+      const topLevelAccordionHeaders = accordionHeaders.filter(function() {
+        return dom(this).parent()[0].tagName === 'va-accordion-item';
+      });
 
       // Prevents attempted to make a header larger than an H6
       if (incrementedHeaderLevel > 6) incrementedHeaderLevel = 6;
 
       // only reassign header level if incrementedHeaderLevel is actually a number
       if (incrementedHeaderLevel && typeof incrementedHeaderLevel === 'number')
-        accordionHeaders.each((_, header) => {
+        topLevelAccordionHeaders.each((_, header) => {
           dom(header).prop('tagName', `h${incrementedHeaderLevel}`);
           didModifyHeaders = true;
         });
--- a/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
+++ b/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
@@ -64,7 +64,7 @@ module.exports = {
       const accordionHeaders = dom(node).find(':header');
 
       // Prevents attempted to make a header larger than an H6
-      if (incrementedHeaderLevel === 6) incrementedHeaderLevel = 5;
+      if (incrementedHeaderLevel > 6) incrementedHeaderLevel = 6;
 
       // only reassign header level if incrementedHeaderLevel is actually a number
       if (incrementedHeaderLevel && typeof incrementedHeaderLevel === 'number')
--- a/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
+++ b/src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
@@ -1,0 +1,81 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable no-console */
+
+/**
+ * The intention of this file is to check the DOM for accordions generated with incorrect
+ * Heading Hierachies and correct this issue by modifying the DOM.
+ * */
+
+module.exports = {
+  initialize() {
+    console.time('Force Correct Accordion Level');
+  },
+  conclude() {
+    console.timeEnd('Force Correct Accordion Level');
+  },
+  modifyFile(fileName, file) {
+    let didModifyHeaders = false;
+
+    if (!fileName.endsWith('html')) {
+      return;
+    }
+
+    const { dom } = file;
+
+    // Bail early if page doesnt have an Accordion
+    if (dom("[data-template='paragraphs/collapsible_panel']").length < 1) {
+      return;
+    }
+
+    const findPreviousHeaderLevel = node => {
+      // Returns the header level when the last header is in a wysiwyg div
+      const previousHeaderLevelParent = dom(node).prev(
+        'div[data-template="paragraphs/wysiwyg"]',
+      );
+
+      const lastHeader = previousHeaderLevelParent.find(':header').last()[0];
+      if (lastHeader) {
+        return lastHeader?.tagName;
+      }
+
+      // Returns the last header level for when there is no header nested in a wysiwyg
+      const fallbackPreviousHeader = dom(node)
+        .prevAll('h1, h2, h3, h4, h5, h6')
+        .first()[0];
+
+      if (fallbackPreviousHeader) {
+        return fallbackPreviousHeader?.tagName;
+      }
+
+      // Fallback for when no proper headers are provided
+      return 'h2';
+    };
+
+    const collapsiblePanelNodes = dom(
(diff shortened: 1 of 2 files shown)


jtmst:
Performance note,  These is adding 2 minutes 21 seconds to the contant build (tested locally)

laflannery:
**Notes from most recent testing:**
(Refer to testing sheet above)
- Only 1 issue found in "Correct accordions" tab
- No issues found in "Incorrect accordions tab"

**Clarifications:**
- It does not that the first scenario discussed in [this previous comment](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12352#issuecomment-1428406403) is actually impacted by this plugin - though we originally thought it would be
  - There are 2 examples of this in the "Incorrect accordions" sheet, lines 94 and 95 you can look at for reference: http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/wilmington-health-care/work-with-us/internships-and-fellowships/optometry/ and http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/richmond-health-care/work-with-us/internships-and-fellowships/occupational-therapy-fellowship-program/
  - I am actually good with these not being included now that I see it > _Can you confirm that this is in fact a change and this scenario has not been impacted?_
- We had previous discussed and wanted to keep in mind what would happen when accordions became H2s, would these be included in the "On this page" ToC?
  - I believe the answer is no, looking at line 203 in the "Incorrect accordions" sheet but can you confirm this is correct and what will always happen for the plugin behavior? http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/washington-dc-health-care/work-with-us/internships-and-fellowships/social-work-internship-program/

**Concern**
- I still have a concern that the plugin is changing header levels WITHIN accordions and not just the top level accordion headers itself. I found 2 of these instances so far:
  - Line 150 of the "Correct Accordion" sheet > inside the accordions were originally H3s and now they are H4s: http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/indiana-health-care/contact-us/
  - Line 27 of the "Correct Accordion" sheet > some of the headings inside the accordions where H3s and now they are H4s. The headings that were originally H4s, did not get changed: http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/phoenix-health-care/work-with-us/internships-and-fellowships/endocrinology-and-metabolism-fellowship/

jtmst:
> **Notes from most recent testing:** (Refer to testing sheet above)
> 
> * Only 1 issue found in "Correct accordions" tab (I do not consider this launch-blocking)
> * No issues found in "Incorrect accordions tab"
> 
> **Clarifications:**
> 
> * It does not look like the first scenario discussed in [this previous comment](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12352#issuecomment-1428406403) is actually impacted by this plugin - though we originally thought it would be
>   
>   * There are 2 examples of this in the "Incorrect accordions" sheet, lines 94 and 95 you can look at for reference: http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/wilmington-health-care/work-with-us/internships-and-fellowships/optometry/ and http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/richmond-health-care/work-with-us/internships-and-fellowships/occupational-therapy-fellowship-program/
>   * I am actually good with these not being included now that I see it > _Can you confirm that this is in fact a change and this scenario has not been impacted?_
> * We had previous discussed and wanted to keep in mind what would happen when accordions became H2s, would these be included in the "On this page" ToC?
>   
>   * I believe the answer is no, looking at line 203 in the "Incorrect accordions" sheet but can you confirm this is correct and what will always happen for the plugin behavior? http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/washington-dc-health-care/work-with-us/internships-and-fellowships/social-work-internship-program/
> 
> **Concern**
> 
> * I still have a concern that the plugin is changing header levels WITHIN accordions and not just the top level accordion headers itself. I found 2 of these instances so far:
>   
>   * Line 150 of the "Correct Accordion" sheet > inside the accordions were originally H3s and now they are H4s: http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/indiana-health-care/contact-us/
>   * Line 27 of the "Correct Accordion" sheet > some of the headings inside the accordions where H3s and now they are H4s. The headings that were originally H4s, did not get changed: http://18e7a3a4ed9f170fc62d1231cd065c41.review.vetsgov-internal/phoenix-health-care/work-with-us/internships-and-fellowships/endocrinology-and-metabolism-fellowship/

@laflannery 
I did not include the feature div logic when I rebuilt the plugin, but we could add that later if we feel its necessary.

The accordion headers should be excluded from the "On this page" component, and I have not seen any examples where that is not true

Changing the headers inside the accordion content was initially an error, but it did raise a question for me as to whether or not making them the same level as the main accordion header would help prevent skipped header levels if the editor were to throw h6s in there?  It has the potential to change styling though, if an editor were to create h5s inside an accordion whos headers get corrected to H2s. 

The one error for the /st-louis-health/ page happens because that header is nested inside a list_of_links_teasers component, kind of like how the headers inside the feature blocks are.  I think the two of those could be bundled together in a future bug squashing ticket.

laflannery (approved):

laflannery:
@jtmst Just to circle back on this:

> Changing the headers inside the accordion content was initially an error, but it did raise a question for me as to whether or not making them the same level as the main accordion header would help prevent skipped header levels if the editor were to throw h6s in there? It has the potential to change styling though, if an editor were to create h5s inside an accordion whos headers get corrected to H2s.

You are correct in that if we forced them to be the same level as the accordions we would most likely be removing even more violations, however this is editor created content and we are going to approach this differently (probably slower and multiple approaches) but we still don't want to take a wholesale approach necessarily to editor content which is why it was better to do what we did and just leave it fo now.

src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
```diff
@@ -0,0 +1,85 @@
...
+    const collapsiblePanelNodes = dom(
+      `[data-template="paragraphs/collapsible_panel"]`,
+    );
+
+    // Starts at each collapsible panel, steps back to the most recent wysiwyg header, and reassigns based on that
+    collapsiblePanelNodes.each((index, node) => {
+      const previousHeaderLevel = findPreviousHeaderLevel(node);
+      // in the case that there is a known previous header level in a wysiwyg div
+      let incrementedHeaderLevel =
+        parseInt(previousHeaderLevel?.substring(1), 10) + 1;
```
ryguyk, line 63: This works well, but might be a place for improved speed. Since we have a very specific set of possible headers, it's not necessary to calculate a substring and then parse an integer from it. Instead, we could just calculate this from a hash:
```
const getIncrementedHeaderLevel = previousHeaderLevel => ({
    'h1': 2,
    'h2': 3,
    'h3': 4,
    'h4': 5,
  })[previousHeaderLevel] || 6;
```

This would have the added benefit of incorporating the check to ensure the maximum incremented header level is 6.

src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
```diff
@@ -0,0 +1,85 @@
...
+      const accordionHeaders = dom(node).find(':header');
+      // Ensures that only the main header is affected and not headers in the accordion content
+      const topLevelAccordionHeaders = accordionHeaders.filter(function() {
+        return dom(this).parent()[0].tagName === 'va-accordion-item';
+      });
```
ryguyk, line 68: This works just fine, but it feels a bit verbose to find all the items and then write a separate filter function. Instead, I think we could probably accomplish this all in a single selector.
```
dom(node).find('va-accordion-item > :header')
```

Perhaps there are some edge cases I'm missing, though.

src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
```diff
@@ -0,0 +1,85 @@
...
+
+      // Prevents attempted to make a header larger than an H6
+      if (incrementedHeaderLevel > 6) incrementedHeaderLevel = 6;
+
+      // only reassign header level if incrementedHeaderLevel is actually a number
+      if (incrementedHeaderLevel && typeof incrementedHeaderLevel === 'number')
+        topLevelAccordionHeaders.each((_, header) => {
+          dom(header).prop('tagName', `h${incrementedHeaderLevel}`);
+          didModifyHeaders = true;
+        });
```
ryguyk, line 78: Another possible approach for improved speed might be to not loop through all of these and set the headers if they are already set to the desired level. Since all the top-level headers on the accordions should be set the same (because of design-system implementation), we'd likely only need to check the first one to know if we should update all of them.

ryguyk (approved):
This looks very good. I offered a few suggestions for possible speed improvements, but I'm approving because it's likely not critical to implement them immediately.

rmessina1010 (approved):
LGTM

src/site/stages/build/plugins/modify-dom/force-correct-accordion-headers.js
ryguyk: I should note here for thoroughness that we should perhaps call `.toLowerCase()` on `previousHeaderLevel` because when I was testing with jQuery (which is effectively the same implementation as Cheerio, as I understand it), the header `tagname` was returned with a capital `H` (e.g. `H3`).

PR #1488, opened 2023-02-28
[VA-12776] Lovell Federal System Page Cloning Exception

maxx1128:
## Description

closes [#12776](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12776) and closes [#12690](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12690)

## Testing done & Screenshots

Visual, see below. Titles, content, and breadcrumb are updated/corrected.

Screen Shot 2023-02-27 at 7 50 04 PM

## QA steps

1. Check that the Tricare and VA pages have the right information
   - [ ] The titles are correct
   - [ ] The breadcrumbs are correct
   - [ ] The content reflects the correct title and the info in the CMS
   - [ ] The Tricare page doesn't have a "Manage Your Health Online" section

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/lovell/update-page.js
+++ b/src/site/stages/build/drupal/lovell/update-page.js
@@ -43,22 +43,6 @@ function getLovellSwitchPath(vars) {
   return getLovellVariantOfUrl(page.entityUrl.path, oppositeVariant);
 }
 
-function getLovellBreadcrumbs(vars) {
-  const { page, variantName, linkVar } = vars;
-  // Modify Breadcrumb
-  return page.entityUrl.breadcrumb.map(crumb => {
-    // eslint-disable-next-line no-param-reassign
-    crumb.text = crumb.text.replace(
-      /Lovell Federal health care/,
-      getLovellTitle(variantName),
-    );
-    // eslint-disable-next-line no-param-reassign
-    crumb.url.path = getLovellVariantOfUrl(crumb.url.path, linkVar);
-
-    return crumb;
-  });
-}
-
 function getLovellVariantTitle(title, vars) {
   const { variantName } = vars;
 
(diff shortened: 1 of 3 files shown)


src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -234,6 +235,11 @@ function processLovellPages(drupalData) {
   } = drupalData.data.nodeQuery.entities.reduce(
     (acc, page) => {
       if (isLovellFederalPage(page)) {
+        // Federal Region Homepages should not be cloned
```
ryguyk, line 238: **SHOULD**
Very small fix that might feel pedantic, but I think it'll be more clear to change this comment text to `Homepage` (singular). There is one node (the Federal region/system page) that we want to ignore. It would become two separate pages (one for VA, one for TRICARE) if we were to let it go through, but we want to stop it _before_ it turns into two.

Very nice solution to just ignore it when that condition is true! Well done!

src/site/stages/build/drupal/lovell/update-page.js
```diff
@@ -47,11 +47,13 @@ function getLovellBreadcrumbs(vars) {
   const { page, variantName, linkVar } = vars;
   // Modify Breadcrumb
   return page.entityUrl.breadcrumb.map(crumb => {
-    // eslint-disable-next-line no-param-reassign
-    crumb.text = crumb.text.replace(
-      /Lovell Federal health care/,
-      getLovellTitle(variantName),
-    );
+    if (!crumb.text.includes(getLovellTitle(variantName))) {
```
ryguyk, line 50: **MUST**
I think we need to handle this like we handle the replacements in other places:
```
  crumb.text = crumb.text.replace(
    new RegExp(
      `${LOVELL_TITLE_STRING} health care(?:(?: - VA| - TRICARE)?)`,
      'i',
    ),
    getLovellTitle(variantName),
  );
```
The breadcrumb issue goes deeper than just the system pages. Currently, this breadcrumb appears on a TRICARE page, because it initially was set to `Lovell Federal health care - VA` and then the existing logic replaced `Lovell Federal health care` with `Lovell Federal health care - TRICARE` and then the final string became `Lovell Federal health care - TRICARE - VA`.
image

With the proposed fix, the condition would evaluate to false because the variant is TRICARE and this breadcrumb does not include TRICARE (but it does include VA), and the result would be the exact same as in the screenshot.

ryguyk (changes requested):

src/site/stages/build/drupal/lovell/update-page.js
maxx1128: This is a good catch, and it looks like `getLovellVariantTitle` would work here too 👍

ryguyk (approved):
Nice quick fixes!

PR #1493, opened 2023-03-03
VA-12806: Update Lovell listing links for variants

maxx1128:
## Description

Lovell listing pages have a combination of Tricare/VA and federal events. The Federal event links need the appropriate Tricare/VA variations in their URLs to keep users in a consistent section. But links that should be Tricare have VA in the URL. This pull request adds an extra adjustment to listing page data so the URLs are correct.

closes [#12806](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12806) 

## Testing done & Screenshots

Visual, see below.

Screen Shot 2023-03-03 at 3 57 26 PM

## QA steps

1. Go to a VA listing page
   - [ ] Validate that all URLs use the `va` variant
2. Go to a Tricare listing page
   - [ ] Validate that all URLs use the `tricare` variant

## Acceptance criteria

- [ ] All QA steps are successful

## Definition of done

### DIFF
--- a/src/site/stages/build/drupal/process-lovell-pages.js
+++ b/src/site/stages/build/drupal/process-lovell-pages.js
@@ -167,7 +167,7 @@ function updateLovellSwitchLinks(page, pages) {
  * @param {*} federalPages Listing pages to merge into tricareOrVaPages
  * @returns
  */
-function combineLovellListingPages(tricareOrVaPages, federalPages) {
+function combineLovellListingPages(tricareOrVaPages, federalPages, variant) {
   return tricareOrVaPages.map(listingPage => {
     const typePastMap = {
       event_listing: 'pastEvents',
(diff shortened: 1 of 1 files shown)


src/site/stages/build/drupal/process-lovell-pages.js
```diff
@@ -198,14 +198,24 @@ function combineLovellListingPages(tricareOrVaPages, federalPages) {
...
     const combinedPastListItemEntities = [
       ...pastListItemEntities,
       ...pastListItemEntitiesToCombine,
-    ];
+    ].map(updateEntityUrlForVariant);
     const combinedAllListItemEntities = [
       ...allListItemEntities,
       ...allListItemEntitiesToCombine,
-    ];
+    ].map(updateEntityUrlForVariant);
```
ryguyk, line 218: **CONSIDER**
I don't think we'd need to map over _all_ the entities. It should suffice to map over only the entities coming from the federal listing page. The entities on the VA/TRICARE listing pages should already have the correct URL. It's not technically incorrect as written, as the URLs will still be correct. It's just likely doing unnecessary processing.

ryguyk (approved):
Approved with one consideration for possible improvement.

src/site/stages/build/drupal/process-lovell-pages.js
maxx1128: Good catch, will move the function up a little and see if it still builds correctly.

maxx1128: It still works, will push the update now.

PR #1521, opened 2023-04-03
VA-12717: Facility health services use phone number component

maxx1128:
## Description

Facility health services have phone numbers that are improperly formatted. This uses our existing `va-telephone` component which fixes these formatting issues and gets it set up for better maintainability.

closes [#12717](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12717)

## Testing done & Screenshots

Visual, see below

Screen Shot 2023-04-03 at 11 35 10 AM

## QA steps

1. Do a new build with this pull request
2. Check that the [Jamaica Plain VA Medical Center](http://localhost:3002/boston-health-care/locations/jamaica-plain-va-medical-center/) "Blind and Low Vision Rehabilitation" section phone number is correct
   - [ ] The `aria-label` includes the extension
   - [ ] The anchor tag's `href` includes a "+1" at the start
   - [ ] The anchor tag's `href` has the extension after a comma

## Definition of done

### DIFF
--- a/src/site/facilities/health_care_local_health_service.drupal.liquid
+++ b/src/site/facilities/health_care_local_health_service.drupal.liquid
@@ -40,23 +40,19 @@
       <h5>Phone</h5>
       {% if locationEntity.fieldPhoneNumbersParagraph %}
         {% for number in locationEntity.fieldPhoneNumbersParagraph %}
-          <a aria-label="{{ number.entity.fieldPhoneNumber | accessibleNumber }}"
-             href="tel:{{ number.entity.fieldPhoneNumber }}{% if number.entity.fieldPhoneExtension %}p{{ number.entity.fieldPhoneExtension }}{% endif %}">
-            {{ number.entity.fieldPhoneNumber }}
-            {% if number.entity.fieldPhoneExtension %}
-              x {{ number.entity.fieldPhoneExtension }}
-            {% endif %}
-          </a>
+          <va-telephone
+            contact="{{ number.entity.fieldPhoneNumber | removeDashes }}"
+            extension="{{ number.entity.fieldPhoneExtension | default: '' }}"
+          />
         {% endfor %}
       {% endif %}
     </div>
   {% else %}
     <div class="vads-u-display--flex vads-u-flex-direction--column vads-u-margin-bottom--1" data-template="facilities/health_care_local_health_service">
       <h5>Phone</h5>
-      <a aria-label="{{ fieldPhoneNumber | accessibleNumber }}"
-          href="tel:{{ fieldPhoneNumber }}">
-        {{ fieldPhoneNumber }}
-      </a>
+      <va-telephone
+        contact="{{ fieldPhoneNumber | removeDashes }}"
+      />
     </div>
   {% endif %}
 
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -543,6 +543,10 @@ module.exports = function registerFilters() {
     return null;
   };
 
+  liquid.filters.removeDashes = data => {
+    return data?.replace?.(/-/g, '') || null;
+  };
+
   liquid.filters.deriveLastBreadcrumbFromPath = (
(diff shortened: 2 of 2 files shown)


jtmst (approved):

src/site/facilities/health_care_local_health_service.drupal.liquid
```diff
@@ -40,23 +40,19 @@
...
   {% else %}
     <div class="vads-u-display--flex vads-u-flex-direction--column vads-u-margin-bottom--1" data-template="facilities/health_care_local_health_service">
       <h5>Phone</h5>
-      <a aria-label="{{ fieldPhoneNumber | accessibleNumber }}"
-          href="tel:{{ fieldPhoneNumber }}">
-        {{ fieldPhoneNumber }}
-      </a>
+      <va-telephone
+        contact="{{ number.entity.fieldPhoneNumber }}"
+      />
```
ryguyk, line 55: This doesn't look right to me. This `else` block is the fallback for when there is no phone number available on the service itself. In this `else` block, as was previously the case, it should fall back to the facility phone number. In fact, `number.entity.fieldPhoneNumber` should never be defined here. `number` is scoped to the `for` loop in the `if` block.

ryguyk (changes requested):

laflannery:
@maxx1128 I just looked at this - why does this not look like the component in the design system?

**Design system:**
image

- No spaces in the href
- periods in the aria-label

**Review instance:**
image

- Spaces in the href
- dashes in the aria-label

Why wouldn't this be identical to the DS component?

src/site/facilities/health_care_local_health_service.drupal.liquid
maxx1128: This is a good catch. I'll try to get in a fix for this early tomorrow.

maxx1128:
@laflannery I got it working with the dashes filtered from the numbers and the code is lining up better with what's needed now. This component also includes the extension in the visible phone number as it does in the component library.

Screen Shot 2023-04-06 at 11 00 11 AM

maxx1128:
@ryguyk This is ready for re-review.

src/site/filters/liquid.js
```diff
@@ -543,6 +543,13 @@ module.exports = function registerFilters() {
     return null;
   };
 
+  liquid.filters.numberWithoutDashes = data => {
+    if (data) {
+      return data.replace(/-/g, '');
+    }
+    return null;
+  };
```
ryguyk, line 551: If `data` is not a string, this will throw an error. I think there are two options:
1. `if (data && data.replace)`
   - This will return null if it's undefined or if it's not a string (doesn't have the `replace` method).
2. `return data.replace?.(/-/g, '')`
   - This will return `undefined` if `replace` is not available. You could return `null` by instead doing `return data.replace?.(/-/g, '') || null`, if returning `null` is actually necessary.

Further, I wonder about the naming/usage of this function. Right now, it's really just `removeDashes`. The passed in value does not need to be a string representing a number. So, this function could theoretically be used elsewhere. This might be a good thing. If it is, I'd suggest renaming to reflect that. 

If not, and we instead want to limit this function call to strings representing phone numbers, we could perhaps ensure that the passed in data is actually of the format `nnn-nnn-nnnn` or whatever format we're expecting. If the passed-in value is not of that format, the function could return either `undefined` or the original value (I'm not sure which makes more sense).

ryguyk (changes requested):

src/site/filters/liquid.js
maxx1128: This is good feedback, I'm going with this for the code and name changes:

```javascript
liquid.filters.removeDashes = data => {
  return data.replace?.(/-/g, '') || null;
};
```

maxx1128: Got a build error, went back to the conditional check.

ryguyk (approved):
This looks great. Nice job on the fixes.

PR #1532, opened 2023-04-19
VA-13252: E2E test for presence of Lovell Menus

maxx1128:
## Description

Lovell menus vanished without notice until the change was in production. This adds End To End tests to ensure the menus are still present in Lovell. They specifically check the "About Us" pages for VA and Tricare for the following elements:

* There are three menu "sections" and each has a header for a label
* There are top level menu links present
* There are at sublevel menu links present (for "About Us")

This amount of testing ensures that the menu is present along with some of the key elements to make sure everything is there, not just the wrapper.

closes [#13252](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13252)
relates to [#13244](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13244)

## Testing done & Screenshots

Successfully ran the new tests locally

Screen Shot 2023-04-19 at 1 04 35 PM

## QA steps

1. Ensure that all the tests pass
   - [ ] Run the tests locally with `yarn cy:run`
   - [ ] See that all automated tests for this PR pass
2. Check that the tests also fail when expected
   - [ ] Make sure tests fail when the menu isn't present (I confirmed this myself in an old local build)
   - [ ] Change the tests to purposefully make them fail, and run them again to see if they actually fail

## Acceptance criteria

- [ ] Tests pass when the Lovell menus are present, and fail when the Lovell menus are not present

## Definition of done

### DIFF
--- a/src/site/facilities/tests/lovell/lovell-pages.cypress.spec.js
+++ b/src/site/facilities/tests/lovell/lovell-pages.cypress.spec.js
@@ -1,0 +1,45 @@
+const testForLovellSidebar = (section, url = '/') => {
+  const pageUrl = `/lovell-federal-health-care-${section}${url}`;
+
+  it('has a sidebar with visible sections with headers', () => {
+    cy.visit(pageUrl);
+    cy.injectAxeThenAxeCheck();
+
+    const numberOfMenuSections = 3;
+
+    cy.get('.va-sidenav')
+      .find('.va-sidenav-level-1')
+      .should('be.visible')
+      .should('have.length', numberOfMenuSections)
+      .find('h2')
+      .should('be.visible')
+      .should('have.length', numberOfMenuSections);
+  });
+
+  it('has a sidebar with clickable links', () => {
+    cy.visit(pageUrl);
+    cy.injectAxeThenAxeCheck();
+
+    // Top level sidebar links
+    cy.get('.va-sidenav')
+      .find('.va-sidenav-level-2 a.va-sidenav-item-label')
+      .should('be.visible')
+      .should('have.length.greaterThan', 0);
+
+    // 2nd level sidebar links for "About Us"
+    if (url === '/about-us/') {
+      cy.get('.va-sidenav')
+        .find('.va-sidenav-level-3 a.va-sidenav-item-label')
+        .should('be.visible')
+        .should('have.length.greaterThan', 0);
+    }
+  });
+};
+
+describe('Lovell Tricare - Main Page', () => testForLovellSidebar('tricare'));
+describe('Lovell Tricare - About Page', () =>
+  testForLovellSidebar('tricare', '/about-us/'));
+
+describe('Lovell VA - Main Page', () => testForLovellSidebar('va'));
+describe('Lovell VA - About Page', () =>
+  testForLovellSidebar('va', '/about-us/'));


src/site/tests/lovell/lovell-pages.cypress.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      .find('.va-sidenav-level-1')
+      .should('be.visible')
+      .should('have.length', numberOfMenuSections)
+      .find('h2')
+      .should('be.visible')
+      .should('have.length', numberOfMenuSections);
+  });
+
+  it('has a sidebar with clickable links', () => {
+    cy.visit(pageUrl);
```
ryguyk, line 20: Do we need to separate these into two separate page visits? If we want two separate `it` blocks, we could perhaps put the `cy.visit` into a `before` hook.

src/site/tests/lovell/lovell-pages.cypress.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      .find('.va-sidenav-level-2 a.va-sidenav-item-label')
+      .should('be.visible')
+      .should('have.length.greaterThan', 9);
+
+    // 2nd level sidebar links for "About Us"
+    if (url === '/about-us/') {
+      cy.get('.va-sidenav')
+        .find('.va-sidenav-level-3 a.va-sidenav-item-label')
+        .should('be.visible')
+        .should('have.length.greaterThan', 2);
```
ryguyk, line 34: I'm not sure we want to get this specific. The content here could change. I'd be more comfortable checking that there's at least _a_ link. Maybe `.should('have.length.greaterThan', 0)`

src/site/tests/lovell/lovell-pages.cypress.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('has a sidebar with clickable links', () => {
+    cy.visit(pageUrl);
+    cy.injectAxeThenAxeCheck();
+
+    // Top level sidebar links
+    cy.get('.va-sidenav')
+      .find('.va-sidenav-level-2 a.va-sidenav-item-label')
+      .should('be.visible')
+      .should('have.length.greaterThan', 9);
```
ryguyk, line 27: Same comment here as with the About Us link count. I'm not sure we want to get this specific. It's not likely that many of these main links would change or go away, but it's conceivable that they could. I don't think we'd want to specifically ensure that there are 10+ of these. If we are checking for three sections, then we might better check that each of the sections has at least one link. Those three sections would be less likely to change, so I'm comfortable keeping that at three exactly if we want.

ryguyk (changes requested):
I mentioned a few changes I think might help. Most importantly, though, we need to move this file to a directory owned by Facilities.

src/site/tests/lovell/lovell-pages.cypress.spec.js
maxx1128: I had them separate since they're each testing different parts of the side nav (headers, links) but the `before` and beforeEach` hooks aren't working when I run the test. Going to try a little bit more on this.

ryguyk (approved):
This looks better. I'm approving.

It does still feel like we shouldn't need to visit the page twice to test all these things. If `before` hooks are providing challenges, perhaps merging these into a single `it` block is another option. I do understand your approach to separating them as well, though, so I'll leave that up to you.

PR #1539, opened 2023-04-26
Revert "Update actions that are outdated."

timcosgrove:
Back pocket revert in the event of issues:

Reverts department-of-veterans-affairs/content-build#1538


timcosgrove:
Uneeded

PR #1549, opened 2023-05-09
Vacms 13456 graphql3 d10 test

timcosgrove:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13456

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that

## Acceptance criteria

- [ ]

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -1556,7 +1556,7 @@ module.exports = function registerFilters() {
         return '00:00:00';
       }
     }
-    if (time === null) {
+    if (time === null || time == -1) {
       return '';
     }
     return moment(time, 'Hmm').format('HH:mm:ss');
--- a/src/site/includes/hours-item.liquid
+++ b/src/site/includes/hours-item.liquid
@@ -8,11 +8,11 @@
 	{% if !hours.starthours && !hours.endhours && !hours.comment %}
         Closed
 	{% else %}
-		{% if hours.starthours %}
+		{% if hours.starthours && hours.starthours != -1 %}
 			{{ hours.starthours | officeHoursTimeFormatter}}
 			to
 		{% endif %}
-		{% if hours.endhours %}
+		{% if hours.endhours && hours.endhours != -1 %}
 			{{ hours.endhours | officeHoursTimeFormatter}}
 		{% endif %}
     {% endif %}
--- a/src/site/stages/build/drupal/graphql/locationsListingPage.graphql.js
+++ b/src/site/stages/build/drupal/graphql/locationsListingPage.graphql.js
@@ -10,7 +10,6 @@ const locationListingPage = `
     ${entityElementsFromPages}
     title
     entityId
-    fieldMetaTags
     fieldOffice {
       targetId
       entity {
--- a/src/platform/site-wide/tests/accessible-modal.cypress.spec.js
+++ b/src/platform/site-wide/tests/accessible-modal.cypress.spec.js
@@ -12,9 +12,9 @@ describe('Accessible Modal Test', () => {
     cy.injectAxeThenAxeCheck();
 
     // Open modal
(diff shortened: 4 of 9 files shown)


src/site/filters/liquid.js
```diff
@@ -1556,7 +1556,7 @@ module.exports = function registerFilters() {
         return '00:00:00';
       }
     }
-    if (time === null) {
+    if (time === null || time == -1) {
```
timcosgrove, line 1559: `==` is intentional. I am not 100% sure whether `-1` would come down as an integer or as a string here, and wanted to catch both cases.

src/site/stages/build/drupal/graphql/locationsListingPage.graphql.js
```diff
@@ -10,7 +10,6 @@ const locationListingPage = `
     ${entityElementsFromPages}
     title
     entityId
-    fieldMetaTags
```
timcosgrove, line 13: These do nothing in our current setup, and under Drupal10/GraphQL updated libraries, they cause problems.

src/site/includes/hours-item.liquid
```diff
@@ -8,11 +8,11 @@
 	{% if !hours.starthours && !hours.endhours && !hours.comment %}
         Closed
 	{% else %}
-		{% if hours.starthours %}
+		{% if hours.starthours && hours.starthours != -1 %}
```
timcosgrove, line 11: This handles a frankly weird recent decision by the office_hours maintainer: https://www.drupal.org/project/office_hours/issues/3333963

Summary: previously the maintainer would use -1 for an empty value, thinking that `null` was not possible; and then, they handled the translation from -1 to `null` at the field formatter level. This worked fine, even for API data retrieval like we do.

With the above issue, the maintainer realized they could actually enter `null` as the value in the DB legitimately, so they started doing this. They also removed the parts of the field formatter that did the translation from -1 to `null`. 

What they did not do is provide any update to existing field values of -1. So, the API is now outputting -1 where it previously output null.

src/site/filters/liquid.js
```diff
@@ -1556,7 +1556,7 @@ module.exports = function registerFilters() {
         return '00:00:00';
       }
     }
-    if (time === null) {
+    if (time === null || parseInt(time, 10) === -1) {
```
timcosgrove, line 1559: Jslint is insisting on the radix parameter (the 10, for base 10) so that we do not assume that the numbers we're looking at are base 8 or base 17.

timcosgrove, line 1559: Please see the comment below for why this change is here. The short answer is, the Drupal 'Office Hours' module can return both `null` and `-1` for a day where there are no hours, and we need to accommodate that.

src/site/includes/hours-item.liquid
```diff
@@ -8,11 +8,11 @@
...
 			{{ hours.starthours | officeHoursTimeFormatter}}
 			to
 		{% endif %}
-		{% if hours.endhours %}
+		{% if hours.endhours && hours.endhours != -1 %}
```
timcosgrove, line 15: More of the same.

src/site/stages/build/drupal/graphql/locationsListingPage.graphql.js
timcosgrove: 1. `fieldMetaTags` returns `null` for us currently; no data is being returned to be used.
2. Under the current GraphQL php library, `fieldMetaTags` cannot have any subrequest items.
3. Under the forthcoming GraphQL php library, `fieldMetaTags` *must* have its subrequest items explicitly queried.

Removing it seemed the easiest option here.

maxx1128 (approved):
I was able to do a successful `content-build` off of this branch 👍

PR #1564, opened 2023-05-16
VACMS-11482 Drupalize footer

randimays:
## Note
In draft so it can wait to merge until after Monday (5/22) production deploy.

## Description
We are now pulling in the first 3 columns and the bottom rail of footer data from Drupal. The 4th column will still be hardcoded in `footer-links.json`.
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11482

[Tugboat instance](https://web-s9xdtgyk5oytztzajslqb0ywpl8utssp.ci.cms.va.gov/)

## Testing done & Screenshots
Tested locally with `vets-website` and `content-build` both running, both in the injected footer scenario and the regular va.gov sites.

Before - Desktop 
Screenshot 2023-05-08 at 12 10 45 PM

After - Desktop
Screenshot 2023-05-08 at 12 11 01 PM

Before - Mobile
Screenshot 2023-05-08 at 12 11 56 PM
Screenshot 2023-05-08 at 12 11 48 PM
Screenshot 2023-05-08 at 12 11 42 PM
Screenshot 2023-05-08 at 12 11 35 PM
Screenshot 2023-05-08 at 12 11 28 PM
Screenshot 2023-05-08 at 12 11 22 PM

After - Mobile
Screenshot 2023-05-08 at 12 12 39 PM
Screenshot 2023-05-08 at 12 12 34 PM
Screenshot 2023-05-08 at 12 12 26 PM
Screenshot 2023-05-08 at 12 12 20 PM
Screenshot 2023-05-08 at 12 12 14 PM
Screenshot 2023-05-08 at 12 12 07 PM

## QA steps
What needs to be checked to prove this works? All footer links and functionality should be checked on the va.gov side and the injected footer side to be sure no functionality has gone awry.

## Acceptance criteria
- [ ] All links work as expected in the current tab – this is different from production but consistent with design system guidelines
- [ ] disabled menu links and unpublished nodes are not included in the footer
- [ ] rel attributes do not include "noopener" (see https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11482#issuecomment-1319030535)
- [ ] rel attributes do not include "noreferrer"
- [ ] Column layout of footer matches current production, using Drupal va-gov-footer menu (Col 4 of top section not drupalized)
- [ ] Bottom rail links of footer match current production, using Drupal footer-bottom-rail menu
- [ ] Language selector links are rendered via existing React component

Mobile:
- [ ] Columns turn into accordions as they do in Prod, in same order
- [ ] Veteran Crisis Line is not part of Contact Us (which is where the rest of its column goes); instead it appears at top of footer as in Prod
- [ ] Bottom rail links of footer match current production, using Drupal footer-bottom-rail menu
- [ ] Language selector links behave as in Prod (accordion), look uniform with other accordions

Injection:
- [ ] Footer presents correctly in injected scenario

## Definition of done


src/site/stages/build/plugins/create-header-footer.js
```diff
@@ -6,15 +6,70 @@ const {
   formatHeaderData: convertDrupalHeaderData,
 } = require('../drupal/menus');
 
+const FOOTER_SECTIONS = {
+  BOTTOM_RAIL: 'bottomRailFooterData',
+  FOOTER_COLUMNS: 'footerColumnsData',
+};
```
ryguyk, line 12: **DISCUSSION**

Using an ENUM-like object like this feels odd to me in this case because we have complete control over the property names. Beyond that, if we wanted to change one of these, we'd still have to change it in other places (e.g. If we change line 10, we'd still have to change it on line 63 _and_ line 69 as well).

This leads to a more general discussion of how I make sense of this problem at large.

So, **Disclaimer:** what follows is for discussion and not any sort of urging/suggesting/requesting changes. I tend to think about this problem slightly differently, and it seems it might be useful to share my approach.

I think the best place to start is to say that I'm not sure I think it's necessary to introduce this sort of switching on property names because I'm not sure it's necessary to introduce the property names in the first place. If each section requires different handling, then we can handle those separately by directly passing each section to whichever handler it needs rather than collectively passing it to a function that then needs to figure out what kind of section it is to then handle it accordingly. Creating an object with property names to then switch on those property names feels like unnecessary indirection:
```
const processSectionType1 = () => {/*process Type 1 section*/};
const processSectionType2 = () => {/*process Type 2 section*/};
const processSectionType3 = () => {/*process Type 3 section*/};

const processAllSections = sections => {
	return flatten(Object.keys(sections).map(sectionName => {
  	if (sectionName === 'sectionType1') {
    	return processSectionType1(sections[sectionName]);
    } else if (sectionName === 'sectionType2') {
    	return processSectionType2(sections[sectionName]);
    } else {
    	return processSectionType3(sections[sectionName]);
    }
  }));
}
const output = processAllSections({
  sectionType1,
  sectionType2,
  sectionType3
});

/**
 * 1. Avoids indirection of processAllSections function.
 * 2. More clear that output is an array.
 */
const output2 = [
  ...processSectionType1(sectionType1),
  ...processSectionType2(sectionType2),
  ...processSectionType3(sectionType3),
];
```

So, now looking more directly at this problem, this is how I think of the problem and how to solve it:
```
const formattedLink = (link, linkIndex, columnNumber = null) => ({
  column: columnNumber,
  href: link?.url?.path,
  order: linkIndex + 1,
  target: null,
  title: link?.description,
});
const formattedColumn = (column, columnId) => column?.links?.map?.((link, linkIndex) => formattedLink(link, linkIndex, columnId)) || [];
const formattedColumns = columns => columns?.links?.reduce?.((acc, column, columnIndex) => [...acc, ...formattedColumn(column, columnIndex + 1)], []) || [];

const footerData = [
  ...formattedColumn(bottomRailFooterData, 'bottom_rail'),
  ...formattedColumns(footerColumnsData),
  ...columnFourData,
];
```

src/site/stages/build/plugins/tests/create-header-footer.spec.js
```diff
@@ -0,0 +1,120 @@
...
+      ],
+    },
+    footerData: [
+      {
+        column: 4,
+        href: 'https://www.va.gov/resources',
+        order: 1,
+        title: 'Resources and support',
+      },
+    ],
```
ryguyk, line 69: **CONSIDER**

It could probably be argued that since this data is hard-coded in the repo already, that we could just import that rather than reconstruct test data.

ryguyk (commented):
Leaving some comments. Nothing here directly leads to disapproval. There are some things I think should be considered for change, and then a lot of other things that are for discussion. I'm not yet approving because I need to do additional QA. I wanted to get these comments back to you sooner than later.

src/site/stages/build/plugins/tests/create-header-footer.spec.js
ryguyk: I can argue this both ways, probably. This probably isn't the biggest thing to worry about.

src/site/stages/build/plugins/create-header-footer.js
```diff
@@ -6,15 +6,70 @@ const {
...
+
+const formatLink = (link, linkIndex, columnNumber = null) => {
```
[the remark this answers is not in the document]
randimays: Re: the `null` default. Thanks for pointing this out. I'd added it before realizing `bottom_rail` was a column ID. I'll update.

src/site/stages/build/plugins/create-header-footer.js
randimays: I like this suggestion! Seems a bit cleaner and easier to follow. Implementing

dsasser (approved):
Tested on Tugboat:
Screenshot 2023-05-18 at 2 09 35 PM
Screenshot 2023-05-18 at 2 09 18 PM

ryguyk (approved):
This looks great!

rsmithadhoc (approved):

PR #1575, opened 2023-05-23
Adds preview support for most Lovell pages.

ryguyk:
## Description
Adds preview support for Lovell pages, with some exceptions:
- Listing pages will not work. These require merging with Federal listing pages, which will not be fetched when building only a single page.
- Homepages will work but will not include umbrella featured items. Getting those items is done by grabbing them from the listing pages, which will not be built when building only the single page.

Note: There's a [related draft PR](https://github.com/department-of-veterans-affairs/content-build/pull/1576) that gets most or all of the way to adding support for these other pages. I've opted for the simpler solution to satisfy this issue, as the larger solution feels like some scope creep. The more robust solution can be considered for a later date if there is a strong need for this support.

closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13289

## Testing done & QA steps
### Ensure Lovell pages can be previewed
1. `yarn preview`
2. http://localhost:3002/preview?nodeId=56142
   - [x] Ensure page loads
   - [ ] Ensure TRICARE version (default)
3. http://localhost:3002/preview?nodeId=56142&lovellVariant=tricare
   - [x] Ensure page loads
   - [x] Ensure TRICARE version (explicit)
4. http://localhost:3002/preview?nodeId=56142&lovellVariant=va
   - [x] Ensure page loads
   - [x] Ensure VA version (explicit)
   
### Ensure non-Lovell pages can _still_ be previewed.
1. http://localhost:3002/preview?nodeId=8127
   - [x] Ensure page loads

## Acceptance criteria
See original ticket.

### DIFF
--- a/script/preview.js
+++ b/script/preview.js
@@ -23,6 +23,9 @@ const HOSTNAMES = require('../src/site/constants/hostnames');
 const bucketsContent = require('../src/site/constants/buckets-content');
 const singlePageDiff = require('./preview-routes/single-page-diff');
 const createMetalSmithSymlink = require('../src/site/stages/build/plugins/create-symlink');
+const {
+  processLovellPages,
+} = require('../src/site/stages/build/drupal/process-lovell-pages');
 
 const defaultBuildtype = ENVIRONMENTS.LOCALHOST;
 const defaultHost = HOSTNAMES[defaultBuildtype];
@@ -280,7 +283,11 @@ app.get('/preview', async (req, res, next) => {
 
     Object.assign(drupalData.data, nonNodeContent.content.data);
 
-    const drupalPage = drupalData.data.nodes.entities[0];
+    drupalData.data.nodeQuery = drupalData.data.nodes;
+    processLovellPages(drupalData);
+    const pageIndex = req.query?.lovellVariant === 'va' ? 1 : 0;
+
+    const drupalPage = drupalData.data.nodes.entities[pageIndex];
     const drupalPath = `${req.path.substring(1)}/index.html`;
 
     if (!drupalPage.entityBundle) {


script/preview.js
```diff
@@ -280,7 +283,11 @@ app.get('/preview', async (req, res, next) => {
 
     Object.assign(drupalData.data, nonNodeContent.content.data);
 
-    const drupalPage = drupalData.data.nodes.entities[0];
+    drupalData.data.nodeQuery = drupalData.data.nodes;
+    processLovellPages(drupalData);
+    const pageIndex = req.query?.lovellVariant === 'va' ? 1 : 0;
```
maxx1128, line 288: **Suggestion:** if there isn't a `lovellVariant` query, then it looks like this defaults to `0` and it goes to Tricare. I got it working by putting in a fallback value like this (or a version of it).

```javascript
let pageIndex = 0;

if (req.query) {
  pageIndex = req.query?.lovellVariant === 'va' ? 0 : 1;
}
```

maxx1128 (changes requested):
The only success criteria not being met right now is the default Lovell variant for the page is Tricare, not VA.

script/preview.js
ryguyk: @maxx1128 Thanks for the thorough QA. In this case, I think the QA steps are wrong rather than the code. You're right that the default is TRICARE. I wrote it as such because that led to the simplest code.

I do think that it's arguable that VA _should_ be the default. Your suggested code will fix this VA vs. TRICARE default issue, but - importantly - it will break every other situation because in every other situation there will only be one page. If we try to set the default `pageIndex` to `1`, that will fail in most cases.

So, we have two options:
1. Update the code to make VA the default while not breaking everything else.
2. Don't change anything and be ok with TRICARE being the default.


Given the stakes, my vote is for the simpler code (option 2). While it feels like VA is a better default for things on va.gov, I'm not sure there is really any value here. We can still render both pages by providing the query parameter. So, I'm updating the QA steps to expect TRICARE as the default.

timcosgrove (approved):

PR #1587, opened 2023-06-01
Notify Slack when there is a failure.

timcosgrove:
## Description
Add Slack notification to the failure step.

Addresses https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13964

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -488,6 +488,15 @@ jobs:
           url: ${{ env.DRUPAL_ADDRESS }}/api/content_release/error
           method: GET
 
+      - name: Notify Slack
+        uses: department-of-veterans-affairs/platform-release-tools-actions/slack-notify@8c496a4b0c9158d18edcd9be8722ed0f79e8c5b4 #
+        continue-on-error: true
+        with:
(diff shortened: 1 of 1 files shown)


ndouglas (approved):

.github/workflows/content-release.yml
```diff
@@ -488,6 +488,15 @@ jobs:
           url: ${{ env.DRUPAL_ADDRESS }}/api/content_release/error
           method: GET
 
+      - name: Notify Slack
+        uses: department-of-veterans-affairs/platform-release-tools-actions/slack-notify@8c496a4b0c9158d18edcd9be8722ed0f79e8c5b4 #
+        continue-on-error: true
```
ndouglas, line 493: **nonblocking**: Should `Notify CMS - Error` also have `continue-on-error: true`? It's a bit concerning that the first notification might fail silently and prevent the rest from firing.

timcosgrove: `continue-on-error` is there for the step, not the workflow. I believe the thinking is, Slack notification breaking should not break content build; but, notify CMS should, because notifying CMS of success/failure is critical for CMS ownership of content release state.

timcosgrove: IOW, notifying slack is informational; notifying CMS is actually structural to the whole content release flow.

timcosgrove: We could switch the order of the CMS and Slack notification steps so that Slack notification of the failure happens regardless of a failure to notify the CMS of the failure. Would that work?

I think that first sentence needs 'failure' in it some more.

PR #1640, opened 2023-07-21
Adds certs for newer RCAs to Dockerfiles.

ndouglas:
I'm not sure that this is pressing. I'm not sure the root Dockerfile is even in use. I don't know how the other one is used.

So this PR might be stupid. It's a risk I'm willing to take, to stand up for what is good and true and consistent in Dockerfiles.


Dockerfile
```diff
@@ -36,6 +36,7 @@ ENV AWS_CA_BUNDLE /etc/ssl/certs/ca-certificates.crt
 # Add VA Root CA to Docker Certificate Authority (CA) Store so that NODE can use it for requests.
 ADD https://raw.githubusercontent.com/department-of-veterans-affairs/platform-va-ca-certificate/main/VA-Internal-S2-RCA1-v1.cer /usr/local/share/ca-certificates/
 RUN openssl x509 -inform DER -in /usr/local/share/ca-certificates/VA-Internal-S2-RCA1-v1.cer -out /usr/local/share/ca-certificates/VA-Internal-S2-RCA1-v1.crt
+ADD https://raw.githubusercontent.com/department-of-veterans-affairs/platform-va-ca-certificate/main/VA-Internal-S2-RCA2.cer /usr/local/share/ca-certificates/VA-Internal-S2-RCA2.cer.crt
```
ndouglas, line 39: This is already in PEM format, so it shouldn't need to be converted.

powellkerry (approved):

PR #1650, opened 2023-08-03
VACMS-11354 Add IDs to accordions

randimays:
## Description
Ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/11354

When linking to an accordion using a fragment (e.g. `/health-care#what-are-my-benefit-options`), the page was not jumping to the correct content and opening the accordion. This was partly due to a Design System bug that is now fixed; the accordion was missing the `open` property when this scenario happened. But the reason the page wasn't jumping was because the `<va-accordion-item>`s were not labeled with the correct IDs.

## Testing done & Screenshots
`campaign_landing_page.drupal.liquid`
`health_care_local_facility.drupal.liquid`
`landing_page.drupal.liquid`
`faq_multiple_q_a.drupal.liquid`
`q_a.collapsible_panel.drupal.liquid`

Screenshot 2023-08-03 at 12 58 47 PM
Screenshot 2023-08-03 at 12 45 08 PM
Screenshot 2023-08-03 at 12 37 30 PM
Screenshot 2023-08-03 at 12 36 10 PM
Screenshot 2023-08-03 at 1 56 56 PM

## QA steps

What needs to be checked to prove this works?  Go to a page that uses one of these templates, copy an ID off the `va-accordion-item`, and go to the URL with that appended (with a `#`), and make sure it jumps to the accordion and opens it.

## Acceptance criteria
- [ ] Audit codebase for templates that render accordions, note which do not have ids
- [ ] All templates from audit open when url includes the appropriate fragment

### DIFF
--- a/src/site/facilities/facility_health_service.drupal.liquid
+++ b/src/site/facilities/facility_health_service.drupal.liquid
@@ -7,6 +7,7 @@
       data-label="{{ serviceTaxonomy.name }}"
       data-childlabel="{{ serviceTaxonomy.fieldAlsoKnownAs }}"
       data-template="facilities/facilities_health_service"
+      id="{{serviceTaxonomy.name | hashReference: 30 }}"
     >
       <h3 slot="headline">
         {{ serviceTaxonomy.name }}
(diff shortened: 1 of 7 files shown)


chriskim2311 (approved):
LGTM!

maxx1128 (approved):
I confirmed it locally that the pages jump to the accordions and open them as expected 👍

rmessina1010 (approved):
LGTM

PR #1653, opened 2023-08-08
VACMS-14670: Fix centralized content Alert and Expander Text 

maxx1128:
## Description

The Vet Center centralized content alert and expandable text content aren't rendering the content from the CMS properly. These changes to the front-end template should get the content rendering properly.

relates to [#14647](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14647)

#### With Errors

Screen Shot 2023-08-08 at 10 38 42 AM

#### Without Errors

Screen Shot 2023-08-11 at 12 11 42 PM

## QA steps

**What needs to be checked to prove this works?** That the alert and expandable text render properly with the expected content.
**What needs to be checked to prove it didn't break any related things?** The entire page where this content is rendered.  
**What variations of circumstances (users, actions, values) need to be checked?** None 

### Check Centralized Content
1. Go to the [White Oak Vet Center page](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/brooklyn-vet-center/) and check that the alert and expansion content renders properly (2nd dropdown in FAQ section Click "Am I eligible for Vet Center services as a Veteran?" to expand).
   - [x] Validate that the content renders  as a blue alert properly
   
image

   - [x] validate that the alert can be expanded
   
image

2. Check that other Centralized Content on Vet Center is unaffected Links: [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/brooklyn-vet-center/) -> [Prod](https://www.va.gov/brooklyn-vet-center/)
   - [x] Validate that the content and display of the expanded content on PR and Prod match
4. Check Centralized content on VAMC policy page is unaffected Links: [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/black-hills-health-care/policies/) -> [Prod](https://www.va.gov/black-hills-health-care/policies/)
    - [x] Validate that PR and Prod match
5. Check Centralized content on VAMC medical records page is unaffected Links: [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/black-hills-health-care/medical-records-office/) -> [Prod](https://www.va.gov/black-hills-health-care/medical-records-office/)
   - [x] Validate that PR and Prod match
6. Check Centralized content on VAMC register for care page is unaffected Links: [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/black-hills-health-care/register-for-care/) -> [Prod](https://www.va.gov/black-hills-health-care/register-for-care/)
   - [x] Validate that PR and Prod match
7. Check Centralized content on VAMC billing and insurance page is unaffected Links: [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/black-hills-health-care/billing-and-insurance/) -> [Prod](https://www.va.gov/black-hills-health-care/billing-and-insurance/)
    - [x] Validate that PR and Prod match
### Check Alerts
8. Check a page that has an alert that is NOT coming from centralized content and is a reusable alert  [CMS](https://prod.cms.va.gov/education/other-va-education-benefits)  -> [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/education/other-va-education-benefits/) -> [Prod](https://www.va.gov/education/other-va-education-benefits/)
   - [x] Validate that PR and Prod match for the appearance of the light blue alert block for "The Veteran Rapid Retraining Assistance Program (VRRAP) has closed"
9. Check a page that has an alert that is NOT coming from centralized content and is NOT reusuable (comes with its own heading and text)  CMS  -> [PR](https://web-spvbjheoe12isfu1maky120fjrnm7ivg.demo.cms.va.gov/disability/eligibility/hazardous-materials-exposure/agent-orange/) -> [Prod](https://www.va.gov/disability/eligibility/hazardous-materials-exposure/agent-orange/) 
    - [x] Validate that PR and Prod match for the light blue alert titled "New PACT Act-related presumptive conditions and locations"

## Acceptance criteria

- [x] All QA steps pass
- [x] No other elements in the page render improperly

## Definition of done

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -811,11 +811,30 @@ module.exports = function registerFilters() {
 
     // Converts all complex key/value pairs in obj to simple strings
     // e.g. key: [{ value: 'foo' }] => key: 'foo'
-    const flattenArrayValues = obj => {
+    // Recursion is used to flatten pairs in entity objects deeper in the data
+    const flattenArrayValues = (obj, flattenContentType) => {
+      const isEntityArray = a => {
+        const uniqKeyValues = a.filter((item, pos) => a.indexOf(item) === pos);
+        return uniqKeyValues.length === 1 && uniqKeyValues[0] === 'entity';
+      };
+
       const newObj = {};
       for (const [key] of Object.entries(obj)) {
-        if (Array.isArray(obj[key]) && obj[key][0]?.value) {
+        if (
+          Array.isArray(obj[key]) &&
+          obj[key][0]?.value &&
+          (Object.keys(obj[key][0]).length === 1 ||
+            flattenContentType === 'react_widget')
+        ) {
           newObj[key] = obj[key][0].value;
+        } else if (
+          Array.isArray(obj[key]) &&
+          isEntityArray(obj[key].map(objKey => Object.keys(objKey)[0]))
+        ) {
+          // Recursively flattens nested entity arrays
+          newObj[key] = obj[key].map(nestedObj => {
+            return { entity: flattenArrayValues(nestedObj.entity) };
+          });
         } else {
           newObj[key] = obj[key];
         }
@@ -867,7 +886,7 @@ module.exports = function registerFilters() {
         };
       }
       case 'react_widget': {
-        const normalizedData = flattenArrayValues(entity);
+        const normalizedData = flattenArrayValues(entity, 'react_widget');
         if (!normalizedData.fieldErrorMessage.value) {
           return {
             ...normalizedData,
--- a/src/site/paragraphs/expandable_text.drupal.liquid
+++ b/src/site/paragraphs/expandable_text.drupal.liquid
@@ -16,7 +16,13 @@ Example data:
         <div class="additional-info-title">{{ entity.fieldTextExpander }}</div>
 
         {% if entity.fieldWysiwyg %}
-            <div class="additional-info-content usa-alert-text" hidden>{{ entity.fieldWysiwyg.processed }}</div>
+            <div class="additional-info-content usa-alert-text" hidden>
+                {% if entity.fieldWysiwyg.processed %}
+                    {{ entity.fieldWysiwyg.processed }}
+                {% elsif entity.fieldWysiwyg.0.processed %}
+                    {{ entity.fieldWysiwyg.0.processed }}
+                {% endif %}
+            </div>
         {% endif %}
     </div>
 {% endif %}
--- a/src/site/filters/liquid.unit.spec.js
+++ b/src/site/filters/liquid.unit.spec.js
@@ -1685,6 +1685,61 @@ describe('processCentralizedContent', () => {
                   ],
                 },
               },
+              {
+                entity: {
+                  targetId: '135322',
+                  targetRevisionId: '1025879',
+                  entityType: 'paragraph',
+                  entityBundle: 'alert',
+                  pid: '135322',
+                  label:
+                    'National Vet Center content > Content > Questions > Answer',
+                  status: true,
+                  langcode: 'en',
+                  fieldAlertBlockReference: [],
+                  fieldAlertHeading: [
+                    {
+                      value:
+                        'What are the covered educational assistance benefits?',
+                    },
+                  ],
+                  fieldAlertType: [
+                    {
+                      value: 'information',
+                    },
+                  ],
+                  fieldVaParagraphs: [
+                    {
+                      entity: {
+                        targetId: '135321',
+                        targetRevisionId: '1025878',
+                        entityType: 'paragraph',
+                        entityBundle: 'expandable_text',
+                        pid: '135321',
+                        label:
+                          'National Vet Center content > Content > Questions > Answer > Alert content',
+                        status: true,
+                        langcode: 'en',
+                        fieldTextExpander: [
+                          {
+                            value:
+                              'What are the covered educational assistance benefits?',
+                          },
+                        ],
+                        fieldWysiwyg: [
+                          {
+                            value:
+                              '<p>The covered educational assistance benefits are benefits from any of these programs:&nbsp;</p>\r\n\r\n<ul>\r\n\t<li>Montgomery GI Bill Active Duty</li>\r\n\t<li>Montgomery GI Bill Selected Reserve</li>\r\n\t<li>Post-9/11 GI Bill</li>\r\n\t<li>Reserve Educational Assistance Program (REAP)</li>\r\n\t<li>Veteran Rapid Retraining Assistance Program (VRRAP)</li>\r\n\t<li>Veteran Readiness and Employment (VR&amp;E)</li>\r\n\t<li>Veterans’ Educational Assistance Program (VEAP)</li>\r\n\t<li>Veteran Employment Through Technology Education Courses (VET TEC)</li>\r\n</ul>\r\n',
+                            format: 'rich_text',
+                            processed:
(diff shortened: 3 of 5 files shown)


src/site/paragraphs/expandable_text.drupal.liquid
```diff
@@ -16,7 +16,13 @@ Example data:
         <div class="additional-info-title">{{ entity.fieldTextExpander }}</div>
 
         {% if entity.fieldWysiwyg %}
-            <div class="additional-info-content usa-alert-text" hidden>{{ entity.fieldWysiwyg.processed }}</div>
+            <div class="additional-info-content usa-alert-text" hidden>
+                {% if entity.fieldWysiwyg.processed %}
+                    {{ entity.fieldWysiwyg.processed }}
+                {% elsif entity.fieldWysiwyg.0.processed %}
```
swirtSJW, line 22: Shouldn't the call to  liquid.filters.processCentralizedContent  have fixed this so that it does not have to be handled on the template?   I think the idea of that filter is to make centralized content (fetched content) look like normal content so we don't need special exceptions on every template.

maxx1128: The function flattens out any arrays with a "value" object, but it doesn't hit the WYSIWYG content since it has several other key/value pairs attached. There aren't any filters for this content, and the same approach was used on `src/site/paragraphs/wysiwyg.drupal.liquid` so it looks like that's the solution we already went with in the codebase.

eselkin: It doesn't seem to add to bloat elsewhere to put in this condition to match. I think it's ok to leave in. But, I agree, the existing logic from that other files seems out of place (event there), especially if we have a function that should theoretically pull objects out of being nested in arrays.

eselkin (commented):
Review comments, and I will approve. Logically everything looks good, just would like to know the tradeoffs of adding to the processCentralizedContent to flatten more or keeping things as they are.

src/site/filters/liquid.js
```diff
@@ -811,11 +811,29 @@ module.exports = function registerFilters() {
...
+        } else if (
+          Array.isArray(obj[key]) &&
+          isEntityArray(obj[key].map(objKey => Object.keys(objKey)[0]))
+        ) {
+          // Recursively flattens nested entity arrays
+          newObj[key] = obj[key].map(nestedObj => {
+            return { entity: flattenArrayValues(nestedObj.entity) };
+          });
         } else {
           newObj[key] = obj[key];
```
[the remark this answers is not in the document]
maxx1128: The need for this never came up before since no one nested an expandable text component inside an alert before, so the data never needed to be flattened more than one level deep. Limiting this additional check to the single-key `entity` instances makes sure that the flattening only happens in the specific instances of nested components happening in nested queries. The function did the flattening in an earlier version without that extra check, and it removed all the WYSIWYG content inside the expandable text component by flattening away the data that the component needed.

Screen Shot 2023-08-11 at 10 59 30 AM

src/site/paragraphs/expandable_text.drupal.liquid
swirtSJW: Apologies, I must be missing something.   We have this case in 
image

in `liquid.filters.processCentralizedContent`  That looks like it is doing the same thing as you are doing here on the template.  The point of processCentralizedContent is to make it look like other non-cetnralized instances of the same data so it does not need to be re-adjusted on a template by template basis.

maxx1128: This `wysiwyg` case is technically separate from the `q_a_section` case in how the centralized content is being processed, so it's not always called.

1. If a `wysiwyg` component is used outside of a QA section, then this case is activated and it can be referenced as `{{ entity.fieldWysiwyg.processed }}`.
2. If a `wysiwyg` component is used inside of a QA section then the `q_a_section` case is activated instead and this `wysiwyg` flattening does not occur. It then needs to be referenced as `{{ entity.fieldWysiwyg.0.processed }}`.

This is how it was originally set up before, hence the same conditional used elsewhere in the code before this pull request. The `wysiwyg` being inconsistently flattened I think is more of a refactor than a bug fix, so it's a good follow-up ticket for this one.

eselkin: I think that's appropriate, refactors for this should probably go in another ticket, since this ticket is really about the bug. Can we bring this up to get a new ticket for refactoring to simplify and centralize logic? @jilladams

eselkin (approved):
I'm approving this PR because logically it fixes the bug. Would love to open another ticket to handle the refactor of logic to make things more uniform across the uses of the entity wysiwyg components.

src/site/paragraphs/expandable_text.drupal.liquid
swirtSJW: @eselkin @maxx1128    Makes sense to move this to a new ticket.   I think though that this split is one we caused with this PR.  
In this PR we made flatten recursive, rather than making processCentralizedContent recursive.  So it recurses all the instances in the fetched content, but only runs them through the switch case once.   It should run each iteration through the switch case.  That way it would not matter what Drupal Paragraph is inside of what other Paragraph, it would process them each the same.

swirtSJW:
This makes  me a bit nervous here with the low number of QA steps.   We have altered code that affects:
- Centralized content in many places throughout the site, not just on VetCenter pages but we are only QAing Vet Center Centralized content.
- Alert display across the site and its use in Centralized content, but we aren't QAing it beyond Centralized Content.

swirtSJW:
Also do we have an explanation for why the review instance is failing?
image

swirtSJW:
Actually the inability to get too the review instance may be on me.  I get this error in my SOCKS terminal every time I go to load that instance, but I can get to other contet-build instances.

image

@maxx1128 @eselkin Are you able to get to it?

jilladams:
Good calls out from both Eli / Steve. I see that Centralized Content is available on the following products:

Screenshot 2023-08-14 at 3 42 10 PM

@eselkin @swirtSJW @maxx1128 let's discuss in Scrum re: how we might be able to limit scope of this change to only Vet Centers, or decide about that requested ticket. @xiongjaneg FYI

swirtSJW:
@jilladams  wrote
> how we might be able to limit scope of this change to only Vet Centers,

There is not a nice way to do that.   The fix for this is in a function that gets applied to all Centralized content in the system.  That is the right place to fix it, we just need broader coverage on our QA steps to make sure we did not break existing.  The same is true of the alert template change.  It is used in other places.  We just have to proceed cautiously with a complete set of QA steps.

jilladams:
@maxx1128 is now on PTO for 3 days, which I didn't remember was coming. We need to regroup fairly urgently on what our team can do to get this over the line in his absence. @swirtSJW @eselkin I've asked Jane to queue that convo for the early end of planning tomorrow.

swirtSJW:
This review instance is still not available http://f62a7e4177de030d6417c8deab19435e.review.vetsgov-internal/

swirtSJW:
I am adding QA step now.

swirtSJW:
QA steps are added, but they have no links, because the PR environment does not seem to be available. 
I am not sure what the content-build repo does to rebuild the environment so I am going to pull this PR into a tugboat build

swirtSJW:
@eselkin and @jilladams QA steps are updated with links and are ready for a run-through.

swirtSJW:
I just triggered a content-build from the tugboat to make sure it was using the latest code from this PR.  It should be complete within 30 min.

eselkin:
> This review instance is still not available http://f62a7e4177de030d6417c8deab19435e.review.vetsgov-internal/

Still not available today.

src/site/paragraphs/expandable_text.drupal.liquid
eselkin: @swirtSJW I think that seems like a better alternative.

swirtSJW:
The links in the QA steps at the top of the page are to Max's Tugboat which I rebuilt with this PR, so those should be ready to use for your QA @eselkin   As each one validates, feel free to check off each box as you go.

eselkin:
All QA steps pass.

eselkin (approved):
LGTM

powellkerry (approved):

swirtSJW (approved):
Tested is working correctly.   Concerns about handling the recursion a bit higher rather than the flattener, up will be handled in another ticket.

PR #1663, opened 2023-08-14
VACMS-10944 Add Reusuable QAs to benefit detail pages

chriskim2311:
## Description
Add new QA group content component to be used in benefit detail pages

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that

## Acceptance criteria

- [ ]

## Definition of done

### DIFF
--- a/src/site/layouts/page.drupal.liquid
+++ b/src/site/layouts/page.drupal.liquid
@@ -50,9 +50,13 @@
           {% endif %}
 
           {% for block in fieldContentBlock %}
+            {% if block.entity.entityBundle == "q_a_group" %}
+            {% include "src/site/paragraphs/q_a_group_content.drupal.liquid" with entity = block.entity %}
+          {% else %}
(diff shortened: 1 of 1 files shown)


laflannery:
@chriskim2311 I see no current a11y issues however I wanted to check the accordion display so I was going to change it in the CMS and I noticed that [this particular page](https://web-lbedteddjnjsnvayhblwi4xdcua7g1mk.ci.cms.va.gov/qa-group-benefit-detail-page-first-in-order/) was already set up to be displayed as accordions in the CMS:
image


Is this something that we need to make functional here on this new content type or is this perhaps an existing issue and this checkbox does not work on any content type where we have it?

chriskim2311:
@laflannery Good catch, I looked into this and its not currently implemented to use this field. I will follow-up with Wes and Jill about cutting a new ticket to address or if this is launch blocking.

randimays (approved):
LGTM

PR #1671, opened 2023-08-18
Example of how to implement custom header and footer

Andrew565:
- Add keys to registry
- Look for key in header
- Create a custom-header.html file
- Look for key in footer
- Create a custom-footer.html file

## Description

relates to #[1828](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1828)

## Testing done & Screenshots

Tested locally

### DIFF
--- a/src/applications/registry.json
+++ b/src/applications/registry.json
@@ -1551,7 +1551,10 @@
     "productId": "63fa0a80-e79d-47e7-9cce-e03a8edb2535",
     "template": {
       "vagovprod": false,
-      "layout": "page-react.html"
+      "layout": "page-react.html",
+      "noHeader": true,
+      "customHeader": "custom-header",
+      "customFooter": "custom-footer"
     }
    },
    {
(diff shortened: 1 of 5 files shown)


micahchiang (commented):
I think a challenge with this approach is the implication of allowing teams to provide their own custom headers and footers. In this example, it seems like if a team wants to use a custom header, they would put a new file in `site/includes` with the markup they wish to use. Some questions that would need to be addressed:

1. how do the `customHeader` and `customFooter` flags determine which team's custom header and/or footer should be used? This:
```
{% if customFooter === "custom-footer" %}
        {% include "src/site/includes/custom-footer.html" %}
{% endif %}
```
Would maybe need to do something like this (not sure if this is the right syntax):

```
{% if customFooter === "custom-footer" %}
        {% include "src/site/includes/{{ customFooter }}.html" %}
{% endif %}
```
2. How much flexibility would we want teams to have? 
3. Should this even be exposed to teams? 
  I think something we'd want to avoid is `site/includes` getting too bloated with header and footer files specific to each team. A different approach may be to somehow pass in some metadata from `registry.json` and use that as props to a predefined custom header template. Not sure. 

Either way, I think this is a great jumping off point for some further discussion.

PR #1686, opened 2023-08-28
Update content-release.yml

ndouglas:
A recent PR added some steps to the notify-cms-on-failure bit, but didn't add a checkout step. Consequently notifying the CMS of failure is not functioning correctly right now.


ariperez:
LGTM

ariperez (approved):

PR #1687, opened 2023-08-29
Schedule the prune-self-hosted-runner script.

ndouglas:
## Description

I made a tweak to the prune-self-hosted-runners script to retrieve up to 100 runners (bugfix!) and also to schedule it for regular execution. I've been using the script locally to prune old, idle runners for a few days and haven't encountered any unexpected issues.

I don't want to merge this just yet, though, as I'd like to see how stable the runners are in their current arrangement.

### DIFF
--- a/.github/workflows/prune-self-hosted-runners.yml
+++ b/.github/workflows/prune-self-hosted-runners.yml
@@ -3,7 +3,7 @@ name: Prune Self-Hosted Runners
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 * * * *' # Hourly
+    - cron: '27 2 * * *' # Daily, @ ~2:27AM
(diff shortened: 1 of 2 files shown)


.github/workflows/prune-self-hosted-runners.yml
```diff
@@ -2,6 +2,8 @@ name: Prune Self-Hosted Runners
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 * * * *' # Hourly
```
ndouglas, line 6:
```suggestion
    - cron: '27 2 * * *' # Daily, @ ~2:27AM
```

A little more random-ish.

powellkerry (approved):

PR #1691, opened 2023-08-29
VACMS-13222: Fixes styling with some issues being seen on prod

chriskim2311:
## Description

In some circumstances we would see some header line-height mismatches with the When and Cost fields. This PR adds some margin for the where field to space it between the cost field and makes sure that both headers are level.

## Testing done & Screenshots
Tested multiple events against new changes and made sure that events did not have mismatching header levels. Some examples are below:

Outreach_And_Events___Labor_Day_Celebration_-_Los_Angeles__CA___Veterans_Affairs

Outreach_And_Events___Brick_By_Brick_Weekly_Community_Stream___Veterans_Affairs

## Acceptance criteria
- [ ] Make sure styles are showing up correctly in UI

### DIFF
--- a/src/site/layouts/event.drupal.liquid
+++ b/src/site/layouts/event.drupal.liquid
@@ -280,62 +280,62 @@
 
   function onExpandRecurringEventsClick() {
   var expandRecurringEventsButton = document.getElementById('expand-recurring-events');
-                            var expandRecurringEventsIcon = document.getElementById('expand-recurring-events-icon');
-                            var recurringEvents = document.getElementById('recurring-events');
-                        
-                            // Escape early if we are not able to find the elements.
-                            if (!expandRecurringEventsButton || !expandRecurringEventsIcon || !recurringEvents) {
-                              return;
-                            }
-                        
-                            // Toggle the bottom borders of the button.
-                            if (expandRecurringEventsButton.style.borderBottomLeftRadius === '0px') {
-                              expandRecurringEventsButton.style.borderBottomLeftRadius = '5px';
-                              expandRecurringEventsButton.style.borderBottomRightRadius = '5px';
-                            } else {
-                              expandRecurringEventsButton.style.borderBottomLeftRadius = '0px';
-                              expandRecurringEventsButton.style.borderBottomRightRadius = '0px';
-                            }
-                        
-                            // Toggle the visibility of the recurring events.
-                            recurringEvents.classList.toggle('vads-u-display--none');
-                            recurringEvents.classList.toggle('vads-u-display--flex');
-                        
-                            // Toggle the icon.
-                            expandRecurringEventsIcon.classList.toggle('fa-plus');
-                            expandRecurringEventsIcon.classList.toggle('fa-minus');
-                          }
-                        
-                          function onShowAllRecurringEventsClick() {
-                            // Derive recurring event items.
-                            var recurringEventItems = document.querySelectorAll('.recurring-event');
-                        
-                            // Show all recurring events.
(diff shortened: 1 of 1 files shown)


src/site/layouts/event.drupal.liquid
```diff
@@ -280,62 +280,62 @@
 
   function onExpandRecurringEventsClick() {
   var expandRecurringEventsButton = document.getElementById('expand-recurring-events');
```
chriskim2311, line 282: Fixed weird spacing here, not a part of what needs to be reviewed.

randimays (commented):
This is what I'm seeing in the review instance 🤔 

Screenshot 2023-08-30 at 9 21 11 AM

Based on the design comp, these should not be in a flex parent. There's not a scenario where they are supposed to be side-by-side (correct?). It seems like we should remove the flex properties on the parent and allow these to flow naturally in the DOM (and remove any unnecessary margin).

src/site/layouts/event.drupal.liquid
```diff
@@ -107,7 +107,7 @@
                   </div>
 
                 {% elsif fieldLocationType == "online" %}
-                  <p class="vads-u-margin--0 vads-u-margin-bottom--2">
+                  <p class="vads-u-margin--0 vads-u-margin-bottom--2 vads-u-margin-right--0p5">
```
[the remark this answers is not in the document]
chriskim2311: This provides a `margin-right: 4px` for the Where field to provide some spacing between the Cost field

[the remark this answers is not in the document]
chriskim2311: With removal of the flex this is not needed for the online event case, removed

src/site/layouts/event.drupal.liquid
```diff
@@ -153,20 +153,21 @@
...
 
-            <!-- Cost -->
-            {% if fieldEventCost %}
-              <div class="vads-u-display--flex vads-u-flex-direction--row vads-u-margin-top--1 vads-u-margin-bottom--2">
-                <p class="vads-u-margin--0 vads-u-margin-right--0p5">
-                  <strong>Cost:</strong>
-                </p>
+          <!-- Cost -->
+          {% if fieldEventCost %}
+            <div class="vads-u-display--flex vads-u-flex-direction--row vads-u-margin-bottom--2">
```
randimays, line 160: Not sure why `Cost:` and the cost are separate paragraphs? This reads a bit strangely with a screen reader (because it focuses separately on each paragraph). Should these be contained within a single `<p>`?

@laflannery curious about your perspective here.

laflannery: @randimays I think this is ok the way we have it. Putting it into 1 paragraph definitely wouldn't be a terrible move but it's a bit dependent on screen readers and how folks browse a site. But for this - I think it does make sense to have "Cost" separate because this is set up as a sort of list with sort of headings for the event meta. So just as we have the "When" and "Where" separate from the date, using the same pattern for Cost make sense to me.

randimays: Thanks @laflannery. Disregard, @chriskim2311  :)

randimays (approved):

laflannery:
@chriskim2311 In looking at Randi's question above I found something odd in the sr-only text for the directions links, let me know if you want me to open a new ticket for this:

- On the detail pages, the sr-only text seems to pull the text of the event when the [address has no link](https://www.va.gov/outreach-and-events/events/52918/) but it will pull the name of there location when [there is a link](https://www.va.gov/outreach-and-events/events/59701/). Are we able to change this so that the location name is always used regardless?

PR #1697, opened 2023-09-05
VACMS-14931 PACT Act URL change

randimays:
## Description
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14931

Our previous URL (including `/test/`) was rejected by the VA servers, so we need to make a change to a new URL.


harshil1793 (approved):

jilladams:
Fully approved and Randi is AFK for a little bit, so merging.

PR #1726, opened 2023-09-29
Increase build timeouts to ~75~ 90 minutes.

ndouglas:
Seeing a lot of timeouts for builds targeting non-prod-CMS. I don't know why this is occurring -- it doesn't seem to be due to a change in the number of files synced or in latency vs. prod.cms.va.gov -- but I'm concerned this might hamper our ability to do the production release today. 

I'd like to increase the timeout for the build phase for both production and non-production builds by ~15~ 30 minutes to be more predictable while we figure this out.

spooky

The download-assets stage seems to take (from a single comparison between two randomly selected builds) ~30 minutes longer than just a couple of days ago:

Screenshot 2023-09-29 at 3 49 18 PM

Screenshot 2023-09-29 at 3 51 12 PM

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -88,7 +88,7 @@ jobs:
     runs-on: [self-hosted, asg]
     needs:
       - validate-build-status
-    timeout-minutes: 60
+    timeout-minutes: 75
     defaults:
       run:
         working-directory: content-build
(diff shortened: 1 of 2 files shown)


timcosgrove (approved):
Provided we roll it back shortly.

CBonade (approved):

ndouglas:
This might have resolved and consequently not need to be merged:

Screenshot 2023-09-29 at 5 40 27 PM

PR #1731, opened 2023-10-03
update vaos root url in prod

simiadebowale:
VAOS is going LIVE with our url changes tomorrow at 10am ET. This PR updates the root url to our new root url and removes the workaround we had in place for testing the new url in staging.


edmund-dunn (approved):
I don't see any issues here. Good to go.

CBonade (approved):

CBonade (approved):

PR #1755, opened 2023-10-21
Remove test of My VA link since the link was hidden.

timcosgrove:
## Description
the My VA link was removed here:

https://github.com/department-of-veterans-affairs/vets-website/pull/26068/files

But, we still had a test that targeted it. This removes that test.


eselkin (approved):
Necessary fix.

chriskim2311 (approved):

PR #1762, opened 2023-10-25
[VA-14940] Add front End template skeleton for Facility content type for VBA

maxx1128:
## Description

This pull request adds the template skeleton for the VBA Facility pages. "Template Skeleton" refers to getting as many elements from the query that are currently available (in this case, everything but the centralized content in accordions), and bringing them to their needed page elements.

relates to [#14940](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14940)

## Testing done & Screenshots

Visual, see below.

Screen Shot 2023-10-31 at 5 44 34 PM
Screen Shot 2023-11-01 at 7 29 49 PM

## QA steps

1. Set up and confirm local build works
   - [ ] Run `yarn build --pull-drupal --drupal-address=https://cms-txcrk8bym4a0ivw3xfr8xvdhezmls36f.demo.cms.va.gov/ username=max.antonucci@agile6.com password=drupal8` to build site with needed data. Confirm the build is successful.
   - [ ] Run `yarn serve` to view site locally.
   - [ ] Confirm VBA Facility page is active by visiting `http://localhost:3002/vba-facilities/locations/seattle-regional-office/`
   - [ ] Confirm that the existing page elements are styled properly and in the proper page layout.
2. Confirm that the correct data from the query is being used in the skeleton
   - [ ] Check that the correct information is currently being pulled from the query and used correctly in the template
   - [ ] Check if any additional data could be pulled from the query and used in the template

## Acceptance criteria

- [ ] All QA steps pass and are successfully checked off.
- [ ] Laura has completed an accessibility review.

## Definition of done


src/site/layouts/vba_facility.drupal.liquid
```diff
@@ -1,3 +1,207 @@
...
+
+          <h2 class="vads-u-line-height--1 vads-u-margin-bottom--3">
+            Other nearby VA locations
+          </h2>
+
+          {% include 'src/site/includes/vet_centers/nearby.liquid' with
+              mainVetCenterPhone = fieldPhoneNumber
+              mainVetCenterAddress = fieldAddress
+              mainVetCenterId = fieldFacilityLocatorApiId
+          %}
```
eselkin, line 189: In the Sketch we're supposed to have an "Other nearby VA locations" section but not a nearby Vet Center locations. I wonder if this is pulling in the Vet Centers, but should really be pulling in the more generic VAMC and other services locally.

src/site/layouts/vba_facility.drupal.liquid
```diff
@@ -1,3 +1,207 @@
...
+
+          <p>
+            Looking for a VA benefits location in another area?
+          </p>
+
+          <p>
+            <a href="#">
```
eselkin, line 196: I think we can make this href to "/find-locations", right?

eselkin (commented):
looks good except a couple of things. I'll work on them.

src/site/layouts/vba_facility.drupal.liquid
eselkin: removed for the time being - until NearbyVALocations is built

eselkin:
@laflannery You can view this site to test: https://web-txcrk8bym4a0ivw3xfr8xvdhezmls36f.demo.cms.va.gov/vba-facilities/locations/seattle-regional-office/

eselkin (approved):
Skeleton base is now correct. Will work on styling in next ticket

jilladams:
Marking DO NOT MERGE. Since this work is not behind a feature flag, we do not want to merge to Prod either until it is, or until we've completed Collab Cycle Staging review.  (Which, now that I say it: we'll need it in Staging in order to complete that review, so we maybe should go ahead and ticket that Feature flag? WE can confirm next week. )

jilladams:
I'm interested in the response to Laura's question -- the Skeleton ticket is being fast-followed by a styling ticket @eselkin has in flight, so if we know there are components we can use, we should be using them, and we could argue for that happening in the styling ticket since this isn't merging to prod right now anyway? 

Also just flagging: the note in the prior comment re: Spotlight links not coming through: that was edited in after the fact so didn't get included in the email notification (FYI @laflannery how that happens), in case we need to fix that in the styling ticket as well. Should be must feedback for merging this PR when it's time either way.

eselkin:
Interesting that that component is taking the uri not the url component of the spotlight. I'll look into it.

eselkin:
@laflannery I'm still learning the components. Which stands out to you as not using components?

laflannery:
@eselkin These elements should be using components:
- Breadcrumbs
- Main Phone number
- Link in Other nearby Locations section
- Get Directions links might be?
  - I'm not sure if the link component allows for the target or rel attibutes
  - I'm also not sure if it allows for the sr-only pattern we use to make unique link labels

Also, for reference, either now or in the next step, I made comments in [this sketch file](https://www.sketch.com/s/891d33ae-152d-471c-ab5d-9aedf89cf6ff/p/772C1EED-1ED9-4544-8DB1-FBB388534931/canvas) where all the components should be

eselkin:
Ok. I'll look there as well. Thank you so much.

eselkin:
@laflannery or @thejordanwood or whoever... The spotlight / featured sections taken from Vet Centers don't work here so I'm copying them and updating them to use the URLs (regular paths) rather than the URIs (nodes). But something struck me about the CTA links with the > right chevrons after them inside the spotlights. These are similar to the va-link active link components but not bold. Should they be using the va-link active links? Should we use the va-link active links, should we make the font-weight normal?

src/site/layouts/vba_facility.drupal.liquid
maxx1128: @eselkin Should we remove this section and heading entirely? Or maybe we can put a conditional around it so that when it gets added, it appears automatically on the next build.

```liquid
{% if fieldOtherVaLocations %}
  <h2 class="vads-u-line-height--1 vads-u-margin-bottom--3">
    Other nearby VA locations
  </h2>

  <div data-widget-type="other-facility-locations-list" data-facilities="{{ fieldOtherVaLocations }}"></div>
{% endif %}
```

eselkin: Do we have that field in the graqphql? From what it sounded like from Christian, they aren't going to add any nearby list to the VBA data. We're going to need to create something like NearbyVetCenters which just takes the id of the facility you're doing the search around, not the list of centers.

maxx1128: That field isn't on the GraphQL, I presumed it would be added later (or something similar to it). So are you referring to instead putting the `src/site/includes/vet_centers/nearby.liquid` back in there when we create the correct field for it?

maxx1128:
The only remaining element from the skeleton feedback is the Nearby Locations, and that will need first to be addressed with [this ticket.](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14929) I will do another sanity check on any other points, but this pull request should be good to merge when appropriate.

chriskim2311 (approved):
This looks good to me, great work!

powellkerry (approved):

maxx1128:
Due to the Thanksgiving holiday, this pull request will be merged on Monday, 11/27.

jilladams:
Per checking in with Max, merging!

PR #1811, opened 2023-11-28
VACMS-10176: Fix Featured events when no featured event is live on Health Care Region Pages

chriskim2311:
## Summary
Pulled filter logic from graphQL to the browser in drupal liquid filters. Now there is a `filterAndSortEvents` that we will use to take in all events for a health care region(up to 1000 events) and filter then sort the events based on most upcoming event. 

## Related issue(s)
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10716

## Testing done
Tested in tugboat env to check for event staleness. Was able to see all events are being passed through to the FE, not just 10 and most upcoming event is being shown.

### DIFF
--- a/src/site/filters/liquid.js
+++ b/src/site/filters/liquid.js
@@ -1109,6 +1109,32 @@ module.exports = function registerFilters() {
     });
   };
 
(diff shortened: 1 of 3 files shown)


randimays (approved):
LGTM

PR #1840, opened 2023-12-21
Codeowner updates for Facilities/PW ownership in layouts/tests folders

jilladams:
This started bc I realized that the va_police layout wasn't set for Facilities as codeowner. Give a mouse a cookie. 🐭 🍪 

Done: 
* Specified which layouts for which products belong to PW vs. Fac. That will be useful as AP team moves forward, for knowing who to coordinate with about what.
* Specified Test folders that are owned by Facilities.
* Grouped Facilities-owned items into product buckets (VAMC/VBA/Vet Centers)
* Added @department-of-veterans-affairs/ap-admins as codeowners of Events, so they'll get notified of any changes to the Events liquid template. Can remove this once PW owns the Next implementation and we have a path to Prod. 

Didn't do: Anything else, including but not limited to: 
* paragraphs
* navigation
* includes that aren't in a subfolder
* blocks

Looking at those helped me realize I was a mouse with a cookie. I'm interested in whether reviewers think the changes I _did_ make are at all useful / should merge, or create a burden and it's easier just to leave things co-owned and let everyone get the PR noise and decide where they want to / should weigh in. 

## Why do I care
1. PR noise. It's nice to get notified where you're needed, instead of everywhere every time.
2. Expedient reviews: less noise means the people who need to actually review will take the time to do so more readily.
3. Growing contributor base: for example, the IIR team will soon join in on the fun. We need to be able to spell out ownership in here, as easily as we do in vets-website.

And maybe a note to @timcosgrove that it would be so cool if it were easy to establish full codeownership of features / related files in Next. I don't know if that's real, but: consider it a request to consider.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -46,6 +46,7 @@ src/site/layouts/story.drupal.liquid @department-of-veterans-affairs/vfs-facilit
 src/site/layouts/tests/vamc* @department-of-veterans-affairs/vfs-facilities-frontend
 src/site/layouts/vamc*.drupal.liquid @department-of-veterans-affairs/vfs-facilities-frontend
 src/site/stages/build/drupal/static-data-files/vaPoliceData @department-of-veterans-affairs/vfs-facilities-frontend
+src/site/tests/police @department-of-veterans-affairs/vfs-facilities-frontend
 
 ## VBA
 src/site/includes/vba_facilities @department-of-veterans-affairs/vfs-facilities-frontend
@@ -71,6 +72,7 @@ src/site/layouts/publication_listing.drupal.liquid @department-of-veterans-affai
 src/site/layouts/q_a.drupal.liquid @department-of-veterans-affairs/vfs-public-websites-frontend
 src/site/layouts/support*.drupal.liquid @department-of-veterans-affairs/vfs-public-websites-frontend
 src/site/layouts/va_form.drupal.liquid @department-of-veterans-affairs/vfs-public-websites-frontend
+src/site/tests/home @department-of-veterans-affairs/vfs-public-websites-frontend
 
 
 # GraphQL Queries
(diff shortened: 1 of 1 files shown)


randimays (approved):
LGTM. Agree 100% with reducing PR noise. Thanks Jill!

timcosgrove (approved):
I think this is fine. It may remove ownership from the primary teams up top, but if it does we can rectify that later.

.github/CODEOWNERS
```diff
@@ -25,13 +25,55 @@ src/site/teasers @department-of-veterans-affairs/vfs-public-websites-frontend @d
...
+
+## Vet Center
 src/site/includes/vet_centers @department-of-veterans-affairs/vfs-facilities-frontend
+src/site/layouts/vet_center*.drupal.liquid @department-of-veterans-affairs/vfs-facilities-frontend
+src/site/layouts/tests/vet_center* @department-of-veterans-affairs/vfs-facilities-frontend
+
+# Public Websites
+src/site/layouts/campaign_landing_page.drupal.liquid @department-of-veterans-affairs/vfs-public-websites-frontend
+src/site/layouts/checklist.drupal.liquid @department-of-veterans-affairs/vfs-public-websites-frontend
+src/site/layouts/event* @department-of-veterans-affairs/vfs-public-websites-frontend @department-of-veterans-affairs/ap-admins
```
CBonade, line 64: It looks like one of the groups added here is flagged as not existing

CBonade (changes requested):

jilladams:
@department-of-veterans-affairs/ap-admins definitely exists. @timcosgrove or @tjheffner I think you both probably have repo level permissions. Could you add @ap-admins as a content-build contributor, here: https://github.com/department-of-veterans-affairs/content-build/settings/access.  I don't have perms to do that. Or: if @accelerated-publishing is already added, I could change the codeowner line to use that group instead. (I chose @ap-admins bc it's a smaller set of folks who actually need the code change awareness, vs. the bigger group with administrative interest but who wouldn't need to do actual code reviews.)

eselkin (approved):
Looks good to me as long as the group gets added.

tjheffner:
I've added the `@ap-admins` team to this repo.

tjheffner (approved):
Screenshot 2024-01-09 at 4 01 32 PM

PR #1869, opened 2024-01-15
VACMS-15318 Smart banners: add store icons locally

randimays:
Added the VA app icons (from the Google Play and App Store) in the repository assets so we can avoid Content Security Policy errors with bringing in images from a different domain. Removed the links from the meta data.

Ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15318


chriskim2311 (approved):
LGTM

CBonade (approved):

PR #1884, opened 2024-01-23
VACMS-15758: Remove feature toggle for single value link

chriskim2311:
**Note**: Delete the description statements, complete each step. **None are optional**, but can be justified as to why they cannot be completed as written. Provide known gaps to testing that may raise the risk of merging to production.

## Summary
Remove code associated with feature_single_value_field_link feature flag that is no longer in use.

## Related issue(s)
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15758


maxx1128 (approved):
I don't see any other references to remove, so 👍

randimays (approved):
LGTM

PR #1913, opened 2024-02-13
Disable form 20-10206 in production

ndsprinkle:
## Summary
As we work with stakeholders to investigate electronic signatures, we need to temporarily disable form 20-10206.

## Related issue(s)
department-of-veterans-affairs/va.gov-team-forms#1029


rmessina1010 (approved):
LGTM

PR #1955, opened 2024-03-12
VACMS-17280: benefit hub component upgrade

chriskim2311:
## Summary
Benefit hub landing template component upgrade v1 to v3

## Related issue(s)
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17280

## Testing done
Tested new components locally
`/health-care/`

## Screenshots
VA_Health_Care___Veterans_Affairs

VA_Health_Care___Veterans_Affairs

VA_Health_Care___Veterans_Affairs

VA_Health_Care___Veterans_Affairs

## What areas of the site does it impact?

Benefit Hubs

### DIFF
--- a/src/site/layouts/landing_page.drupal.liquid
+++ b/src/site/layouts/landing_page.drupal.liquid
@@ -98,14 +98,14 @@
                           href="tel:+1711"
                           onclick="recordEvent({ 'event': 'nav-hub-rail', 'nav-path': 'Ask questions' });"
                         >
-                        {{ service.title }}
+                        <span>{{ service.title }}</span><br>
                         </a>
                       {% elsif service.fieldLink.url.path %}
                         <a
                           href="{{ service.fieldLink.url.path }}" 
                           onclick="recordEvent({ 'event': 'nav-hub-rail', 'nav-path': 'Ask questions' });"
                         >
-                        {{ service.title }}
+                        <span>{{ service.title }}</span>
                         </a>
                       {% else %}
                         {{ service.title }}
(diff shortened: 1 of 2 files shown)


chriskim2311:
@laflannery 
Working on the telephone component upgrade on benefit hubs. Came across this where we are setting the title and the phone number. We have the same issue as before in the other PR where the component cannot take custom non-digit numbers from the CMS. We can make them va-links, or keep them a tags until DST accounts for custom strings.


@thejordanwood Also this would be the new look since the va-link cannot take the title like we did in the previous 

Old
VA_Health_Care___Veterans_Affairs

New
VA_Health_Care___Veterans_Affairs

laflannery:
@chriskim2311 totally my bad for not calling this out because when I did my investigation the first time you brought this up I knew this was used here too - so sorry! 

But that being said, [the discovery ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17326) I made to figure this all out will handle this eventually so at least that's covered for future.

For now, I'm not sure `<va-link>` is going to work because a couple don't seem to have an `href` that is formatted properly:
- The TTY number is `tel:1+771`. Having the + in the middle is going to be not ideal and I'm also questioning the 1, I feel like that might mess it up as well?
- The VETS number on the Health care page has `tel:1-877-222-VETS(8387)` the non-digit characters in the number I think are going to mess up using this properly

So with that - because we can't guarantee proper formatting for these numbers here, we may want to keep these as they are?

thejordanwood:
@chriskim2311 I agree with that approach!

randimays (approved):

eselkin (approved):
LGTM

jamigibbs (approved):

PR #1958, opened 2024-03-12
fix(): Increase the role-duration-seconds for AWS cred step in deploy

hgbarreto:
## Overview
Encountered the following error during a run of `continuous-integration` workflow
`An error occurred (ExpiredToken) when calling the PutObject operation: The provided token has expired.`

Adjusting the role session duration to verify if that is the cause for the token expiration mid-deploy

### Notes
- https://dsva.slack.com/archives/CT4GZBM8F/p1710255732049039


Hassantariq-MetroStar (approved):
Looks good to me.

CBonade (approved):

PR #1966, opened 2024-03-15
VACMS-17412 Analytics adjustments for components

randimays:
## Summary
Add `disable-analytics` to DST web components that have custom events on them.

## Related issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17412

### DIFF
--- a/src/site/blocks/promo.drupal.liquid
+++ b/src/site/blocks/promo.drupal.liquid
@@ -32,13 +32,16 @@
 {% endcomment %}
 {% assign image = entity.fieldImage.entity.image %}
 <div data-template="blocks/promo" data-entity-id="{{ entity.entityId }}" class="merger-r-rail hub-promo" id="promo">
-    <img src="{{ image.derivative.url }}" alt="{{ image.alt }}" title="{{ image.title }}" width="{{ image.derivative.width }}" height="{{ image.derivative.height }}">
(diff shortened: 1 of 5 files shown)


eselkin (approved):
I think we should do this everywhere. But I was curious, do all the recordEvents get captured? I noticed with AdSwerve that they didn't trigger locally with the GA id I expected.

randimays:
@eselkin 

> I think we should do this everywhere.

Jamie Stutt from the Platform Analytics team shared that she doesn't advise using `<va-link>`'s baked-in analytics because it doesn't log the destination URL, which would distinguish between a web address and a PDF file, for instance. 

Jami Gibbs from Design System asked me to file a bug for this last week ([ticket here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2600)) so they can add destination URL to the analytics.

I think our goal should be to eliminate as many custom events as possible and use the web components' baked-in analytics so I've been working to identify gaps in the web components. Dave C is also going to review all of PW's custom events this week and determine where we can tighten up our logging.

> But I was curious, do all the recordEvents get captured? I noticed with AdSwerve that they didn't trigger locally with the GA id I expected.

This is kind of a black box to me. I have noticed that some things don't get logged locally, but haven't had time to poke into it yet. Good callout, though - I definitely think it's something we need to follow up on. Do you want to take that as an action item for the Platform team(s) or should I?

ariperez (approved):
LGTM

PR #1973, opened 2024-03-19
VACMS-17187 Benefits Detail pages web component upgrades

randimays:
## Summary
Convert summary box in Benefit Detail pages to the web component.

## Related issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17187

## Testing done
Tested `/health-care/about-va-health-benefits/` locally.

## Screenshots
Screenshot 2024-03-19 at 11 04 27 AM

## Acceptance criteria
- [x] Featured content section is now using V3 Summary box component
- [x] other features display correctly
- [x] verify/review the E2E tests
- [x] check desktop and mobile
- [ ] a11y review

### DIFF
--- a/src/site/layouts/page.drupal.liquid
+++ b/src/site/layouts/page.drupal.liquid
@@ -40,13 +40,16 @@
 
           {% assign featureCount = fieldFeaturedContent | size  %}
           {% if featureCount > 0 %}
-            <div class="feature">
+            <va-summary-box>
               {% for block in fieldFeaturedContent %}
                 {% assign bundleComponent = "src/site/paragraphs/" | append: block.entity.entityBundle %}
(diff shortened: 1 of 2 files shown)


laflannery:
@randimays this new summary box has a role of region but it's not being found by the screen reader, [the storybook one](https://design.va.gov/storybook/iframe.html?id=uswds-va-summary-box--default&viewMode=story) is though. I also noticed that the storybook one has an aria-label, which is correct because we should be naming what the region is, and I'm not sure if that's something we need to add manually or if it gets auto generated based on the content inside. It should be the heading of the box

laflannery:
Sorry one more thing that I saw while not paying attention during this meeting lol:
- There isn't a space between the alert and the summary box anymore - [I looked at this page](http://771885f0f4077fff7181cbc4e1079014.review.vetsgov-internal/health-care/eligibility/)

PR #2074, opened 2024-05-09
VACMS-16964 Operating status alert link fix

randimays:
## Summary
On the Facility locations page, where operating statuses exist, we previously implemented an active link inside of a slim alert. Because the active link has animation on it (when it is hovered, the caret moves to the right), it expands the colored background of the alert behind it. We converted the link to a regular link with a bold text treatment instead.

## Related issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16964

## Testing done
Tested on `/charleston-health-care/locations/`

https://github.com/department-of-veterans-affairs/content-build/assets/19175324/5c1eb9a4-2cb1-4e7b-957b-61d4f50b8591


eselkin (approved):
What a bold PR. LGTM

PR #2086, opened 2024-05-15
VACMS-17817 Resources & Support va-icon cleanup

randimays:
## Summary
Update Resources & Support icons on the content-build side to `va-icon`.

## Related issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17817

## Testing done
Tested locally.

In order to test the `media_list_images` template, you need to run `yarn preview` locally and go to `localhost:3002/?preview?nodeId=6954`. There is no production content for this template, but it only exists on the R&S page and we do have some content in lower environments.

The icon in the `list_of_links` template was left behind when we converted the `Go to all articles` links to active va-links. The icon wasn't showing up visually, but it is in the DOM in prod. See screenshots in the below section. This can be tested at `/resources`.

The `browse_by_audience` template isn't used in prod, but when it is implemented, it looks just like the browse by topic section, so as long as that section is working ok and implemented identically in the `browse_by_audience` template, we can have confidence it will work if it is implemented.

### media_list_images (not production content)
Screenshot 2024-05-15 at 3 59 37 PM
Screenshot 2024-05-15 at 3 59 33 PM

#### In production
The icon is present next to the active link, but it does not appear
Screenshot 2024-05-15 at 4 02 34 PM

#### Locally
The icon that is not appearing has been removed and the active link still works

Screenshot 2024-05-15 at 4 02 10 PM

### DIFF
--- a/src/site/components/browse-by-audience.drupal.liquid
+++ b/src/site/components/browse-by-audience.drupal.liquid
@@ -8,9 +8,7 @@
     {% for audienceTag in audienceTags %}
       {% if forloop.index <= pageSize %}
         <li class="vads-u-margin-y--2">
-          <a class="vads-u-font-weight--bold vads-u-text-decoration--none" href="{{ audienceTag.entityUrl.path }}/">
-            {{ audienceTag.name }} <i role="presentation" aria-hidden="true" class="fa fa-chevron-right vads-u-margin-left--1"></i>
-          </a>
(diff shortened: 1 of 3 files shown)


laflannery (approved):
Looked at RI and confirmed that everything was still working properly after removing this icon

jtmst (approved):

micahchiang (approved):

PR #2096, opened 2024-05-21
VACMS-17824 Remove axe check smoke test

randimays:
## Summary
We had a ticket to convert all `<i>` icons to `va-icon`, including this template. But it turns out that this template is no longer used. It had a test folder in `src/site/layouts/tests` that used the template as a way to confirm the axe check was working properly. But in December 2021, this test folder was removed: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17824

Removing this template instead.

## Related issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17824


maxx1128 (approved):
This is good. I'm guessing the failing builds are due to some of the tugboat issues, so I'm going to approve it 👍

PR #2141, opened 2024-06-20
Give build an absurdly long timeout.

timcosgrove:
## Summary

- increases the timeout of the build step. This has been hitting its timeout limit recently.

This is a second attempt and it raises the timeout to an absurdly high level. 

## Related issue(s)

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/18362

## Testing done

There is no testing to do; it raises the limit.


7hunderbird (approved):
+1

PR #2294, opened 2024-09-27
Lift 50 minute timeout on build step and raise build job to 75 minutes.

timcosgrove:
## Summary
The build step is currently set to 50 minutes maximum and has timed out occasionally recently. It is not consistently over 50 minutes, but network conditions can lead it to rise above that timing.

This lifts the 50 minute timeout. The build job (all the surrounding work that goes along with the build) is set in this PR at 75 minutes. Again, this is for outliers runs where network conditions slow down the build.

## Testing done
No testing needed; these are simply timing changes.

### DIFF
--- a/.github/workflows/content-release.yml
+++ b/.github/workflows/content-release.yml
@@ -91,7 +91,7 @@ jobs:
     runs-on: [self-hosted, asg]
     needs:
       - validate-build-status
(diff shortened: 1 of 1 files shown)


7hunderbird (approved):
+1

PR #2301, opened 2024-10-03
VATEAM-92265: Normalize Address Digital Form pattern output

derekhouck:
## Summary

- Normalizes the Address Paragraph type created by department-of-veterans-affairs/va.gov-cms#19331

### Example output

```json
 {
        "id": 161347,
        "chapterTitle": "Veteran's mailing information",
        "type": "digital_form_address",
        "pageTitle": "Digital Form: Address",
        "additionalFields": {
          "militaryAddressCheckbox": true
        }
}
```

## Related issue(s)

- department-of-veterans-affairs/va.gov-team#92265

## Testing done

- Added unit tests to the Digital Form post processor, the Digital Form GraphQL fragment, and the `address` GraphQL fragment
- Pulled local Drupal data and confirmed the resulting output from content-build

## What areas of the site does it impact?

Only affects the output of `digital-forms.json`, which is not currently generated in production.

### Quality Assurance & Testing

- [x] I fixed|updated|added unit tests and integration tests for each feature (if applicable).
- [x] Linting warnings have been addressed
- [ ] Documentation has been updated ([link to documentation](#) \*if necessary)
- [ ] Screenshot of the developed feature is added
- [ ] [Accessibility testing](https://depo-platform-documentation.scrollhelp.site/developer-docs/wcag-2-1-success-criteria-and-foundational-testing) has been performed

### Error Handling

- [x] Browser console contains no warnings or errors.
- [x] Events are being sent to the appropriate logging solution
- [ ] Feature/bug has a monitor built into Datadog or Grafana (if applicable)

### Authentication

- [x] Did you login to a local build and verify all authenticated routes work as expected with a test user

## Requested Feedback

- Page titles currently have a "Digital Form: " prefix. That will be addressed by department-of-veterans-affairs/va.gov-team#94290

### DIFF
--- a/src/site/stages/build/drupal/static-data-files/digitalForm/postProcessDigitalForm.unit.spec.js
+++ b/src/site/stages/build/drupal/static-data-files/digitalForm/postProcessDigitalForm.unit.spec.js
@@ -119,6 +119,30 @@ describe('postProcessDigitalForm', () => {
       });
     });
 
+    context('with an Address step', () => {
+      it('includes appropriate fields', () => {
+        twoStepEntity.fieldChapters.push({
+          entity: {
+            entityId: '161344',
+            type: {
+              entity: {
+                entityId: 'digital_form_address',
+                entityLabel: 'Digital Form: Address',
+              },
+            },
+            fieldTitle: 'Generated Address',
+            fieldMilitaryAddressCheckbox: false,
+          },
+        });
+        const [, testForm] = postProcessDigitalForm(queryResult);
+        const [{ additionalFields }] = testForm.chapters.filter(
+          chapter => chapter.type === 'digital_form_address',
+        );
+
+        expect(additionalFields.militaryAddressCheckbox).to.eq(false);
+      });
+    });
+
     context('with an Identification Information step', () => {
       let additionalFields;
 
--- a/src/site/stages/build/drupal/static-data-files/digitalForm/fragments/address.graphql.js
+++ b/src/site/stages/build/drupal/static-data-files/digitalForm/fragments/address.graphql.js
@@ -1,0 +1,14 @@
+/*
+ *
+ * The "Address" Digital Form pattern.
+ *
+ * Pattern documentation:
+ * https://design.va.gov/patterns/ask-users-for/addresses
+ *
+ */
+module.exports = `
+  fragment address on ParagraphDigitalFormAddress {
+    fieldTitle
+    fieldMilitaryAddressCheckbox
+  }
+`;
--- a/src/site/stages/build/drupal/static-data-files/digitalForm/fragments/address.graphql.unit.spec.js
+++ b/src/site/stages/build/drupal/static-data-files/digitalForm/fragments/address.graphql.unit.spec.js
@@ -1,0 +1,13 @@
+/* eslint-disable @department-of-veterans-affairs/axe-check-required */
+
+import { expect } from 'chai';
+import address from './address.graphql';
+
+describe('address fragment', () => {
+  it('includes fieldTitle', () => {
+    expect(address).to.have.string('fieldTitle');
+  });
+  it('includes fieldMilitaryAddressCheckbox', () => {
+    expect(address).to.have.string('fieldMilitaryAddressCheckbox');
+  });
+});
--- a/src/site/stages/build/drupal/static-data-files/digitalForm/fragments/digitalForm.graphql.js
+++ b/src/site/stages/build/drupal/static-data-files/digitalForm/fragments/digitalForm.graphql.js
@@ -1,3 +1,4 @@
+const address = require('./address.graphql');
 const identificationInformation = require('./identificationInformation.graphql');
 const nameAndDateOfBirth = require('./nameAndDateOfBirth.graphql');
 
@@ -7,6 +8,7 @@ const nameAndDateOfBirth = require('./nameAndDateOfBirth.graphql');
  *
  */
 module.exports = `
+  ${address}
   ${identificationInformation}
   ${nameAndDateOfBirth}
 
@@ -28,6 +30,7 @@ module.exports = `
             entityLabel
           }
         }
+        ...address
         ...identificationInformation
         ...nameAndDateOfBirth
       }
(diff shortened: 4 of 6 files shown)


pjhill (approved):

src/site/stages/build/drupal/static-data-files/digitalForm/postProcessDigitalForm.unit.spec.js
```diff
@@ -119,6 +119,30 @@ describe('postProcessDigitalForm', () => {
...
+          },
+        });
+        const [, testForm] = postProcessDigitalForm(queryResult);
+        const [{ additionalFields }] = testForm.chapters.filter(
+          chapter => chapter.type === 'digital_form_address',
+        );
+
+        expect(additionalFields.militaryAddressCheckbox).to.eq(false);
+      });
+    });
```
ryguyk, line 144: **Food for thought, nothing more**

I'm generally a fan (in non-OOP contexts) of trying to keep code as _immutable_ as possible. Given that bias, when I first looked at this code, I wasn't immediately sure what was happening. After pushing a chapter onto `twoStepEntity` (line 124), the subsequent code never references `twoStepEntity`. It felt odd to me.

Seeing that line 137 acts on `queryResult`, I figured out that `queryResult` contains a _reference_ to `twoStepEntity`, and, despite `queryResult` being defined as a `const`, JavaScript's "features" do not actually make `const` objects immutable. So `queryResult` is actually changed by changing `twoStepEntity`, but my conviction is that this reality is not immediately obvious, and, importantly, a potential vector for bugs.

This is not a blocking issue, to be clear. I don't think this *needs* to be changed, especially in this context. This test code is fairly isolated and not overly complex. But it makes me recall an experience I had where this approach was taken in a feature that covered a much larger surface area, and it became very, very confusing. A lot of different code was updating the same object in memory, and some weird bugs resulted. 

Again, just food for thought, and perhaps a theoretical discussion about best practice about which I'm sure 10 engineers would have ten different opinions. 😅

ryguyk (approved):
It looks like you're starting to get into a groove with these now that we have a few behind us. Great to see.

I added some food for thought. No action needed.

src/site/stages/build/drupal/static-data-files/digitalForm/postProcessDigitalForm.unit.spec.js
derekhouck: I totally understand where the confusion comes from. `queryResult` is an array containing `twoStepEntity`, and it is used as the input to `postProcessDigitalForm` and all tests are run on its output. (An array containing two objects, of which we only test the second.)

It may make sense to move these tests to the `../tests` directory instead of living alongside the project files and extracting the test entities into fixture files. Do you have opinions one way or the other?

ryguyk: **Disclaimer:**
Again, I don't think there is any action _required_ here. This is simply an opportunity for some theoretical discussion. I don't want to bog down our work by being overly pedantic. First and foremost, I want code review to focus on whether things _work_. I think this code checks that box, so we should get it in and move on.

That said, I do think code review is a great opportunity for discussion around "best practices", etc., and this crossed me as some good food for thought/discussion. Take it for whatever you think it's worth.

## Test-file location
I don't have strong feelings about where these live, per se. In general, I prefer test files living alongside the files they test, but it does seem like there is some small semblance of an existing approach to put things in the `/tests` directory. I previously added some tests into `/tests/lovell`, for example, probably because the `tests` directory was already there, so I tried to match that existing practice. 

At this point, honestly, this isn't my top concern. It seems the test entities could be extracted into fixture files regardless of which directory they live in, but even that isn't a huge concern for me. This is relatively small still. We can move things around if they get unwieldy.

I'm not sure this is related to my confusion, though. I don't think I communicated well enough what I was attempting to communicate, so let me try again:

## Immutability
After thinking about this some more, I think the clearest way to describe my confusion is to start by saying that this would have been more clear to me:

```
queryResult.data.nodeQuery.entities.twoStepEntity.fieldChapters.push(...);
```

That's long, sure, but it's more clear that we are pushing something into the `queryResult` context. If that's what had been there, I probably would not have thought any further about this because the intention would have been more immediately clear.

BUT, importantly, that's still not the real point. The real point is that it's interesting (and potentially problematic) that these two things are even interchangeable to begin with:

```
// These two have the same effect
// 1.
queryResult.data.nodeQuery.entities.twoStepEntity.fieldChapters.push(...);

// 2.
twoStepEntity.fieldChapters.push(...);
```

The fact that they are interchangeable reveals that the data we're interested in is not well encapsulated in our `queryResult` object. We can easily affect the data inside that object without going "through" that object. This approach might be preferable in certain, limited scenarios, but, in general, it can be the source of some unintended side effects.

I have found in a lot of the things I've worked on that trying to keep things immutable has made it easier to reason about what's going on with my code. Here is an example of how I might tackle this problem with this sort of approach:

```
// Rather than pushing onto (and mutating) an existing object, we create a new one
const threeStepEntity = {
  fieldChapters: [
    ...twoStepEntity.fieldChapters,
    anotherChapter
  ],
};

// Same idea here
const newQueryResult = {
  data: {
    nodeQuery: {
      entities: [oneStepEntity, twoStepEntity, threeStepEntity],
      
      // Or, if all you need is the new one:
      // entities: [threeStepEntity],
    },
  },
};

postProcessDigitalForm(newQueryResult);
```

Now, this doesn't do everything we need to truly be immutable:
1. If we were doing some modifications of nested data, the spread operator might not be enough (it creates a shallow copy), and we'd maybe want to reach for `_.cloneDeep` (lodash) or something similar.
2. If we were to still write something like `twoStepEntity.fieldChapters.push(...)`, it would still affect _both_ `queryResult` and `newQueryResult` (unless we used `_.cloneDeep` when adding the entities to the query result object). But this point probably brings us back to the main point, which is that it's likely not the best approach, in general, to try to mutate things in this fashion. That's the line of code that caught my eye in the first place.

In this case, I think the above change would be a sufficient improvement in that it would remove all of the explicit mutations from the code while at the same time improving (in my opinion) the readability. Both blocks of code have the effect of creating an entity with three steps and attaching that to a query result that is passed to the post-process function. The more immutable approach makes that more clear (in my opinion), if for no other reason than that it explicitly creates an object that indicates it has three steps.

PR #2323, opened 2024-10-22
Revert "Remove build timeout until we know why content release is slow."

timcosgrove:
The conditions that required this change have passed, and we do want the timeout to alert us if content-release is taking too long.

Reverts department-of-veterans-affairs/content-build#2321


7hunderbird (approved):
+1

PR #2326, opened 2024-10-24
Disable connected devices app

acrollet:
## Summary

- This app is slated to be decommissioned by the health portfolio. It already gives a 404, but doing this to make things extra clear.


rmessina1010 (approved):
LGTM

PR #2361, opened 2024-11-20
remove conditional style.css

micahchiang:
## Description
This pull request removes the conditional style.css import in the header when `useLocalStyleAndComponents` is present. We're removing it because no apps are opting out of the global stylesheet.


pmclaren19 (approved):
Pulled down locally and looks good!

PR #2373, opened 2024-12-02
Source Cypress from Va-internal S3.

timcosgrove:
## Summary

- This brings the Cypress testing mechanism in line with what vets-website is doing
  - The main change is that it runs the container and Cypress testing steps on GHA-hosted infrastructure rather than VA-hosted
- This is being done in response to the Cypress executable no longer being downloadable on the VA network. This is a error we first noticed on Dec 2, 2024. An example: https://github.com/department-of-veterans-affairs/content-build/actions/runs/12120500910/job/33814202119#step:6:226

For reference, here is the vets-website implementation of this:
- https://github.com/department-of-veterans-affairs/vets-website/blob/main/.github/workflows/continuous-integration.yml#L663
- https://github.com/department-of-veterans-affairs/vets-website/blob/main/.github/workflows/install/action.yml

## Testing done

-  Continued runs of CI until it completed correctly
- Cypress test runs output successfully: https://github.com/department-of-veterans-affairs/content-build/actions/runs/12147390354/job/33877182392#step:8:312

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -393,20 +393,20 @@ jobs:
 
   cypress-tests:
     name: Cypress E2E Tests
-    runs-on: [self-hosted, asg]
+    runs-on: ubuntu-16-cores-latest
     needs:
       - login-to-amazon-ecr
       - build
     timeout-minutes: 30
     container:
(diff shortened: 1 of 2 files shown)


edmund-dunn (approved):
Nice work!

PR #2417, opened 2025-01-17
Log when Facilities cannot find a URL from its breadcrumb rather than erroring.

timcosgrove:
## Description
All this does is log when a Facilities page does not have a URL in its second breadcrumb component, rather than throwing an error. Please see discussion https://dsva.slack.com/archives/CT4GZBM8F/p1737124609714199

## Testing
1. _Facilities or helpdesk: Please fill in instructions for how to recreate the data situation that led to the error._
2. Run a content-build on this Tugboat instance with the data from step 1. 

Expected: You should see `WARNING: CMS error while building breadcrumbs` in the log output.
Expected: The build should still complete.

### DIFF
--- a/src/site/stages/build/drupal/health-care-region.js
+++ b/src/site/stages/build/drupal/health-care-region.js
@@ -3,6 +3,7 @@
 const _ = require('lodash');
 const moment = require('moment');
 const liquid = require('tinyliquid');
+const { logDrupal: log } = require('./utilities-drupal');
 const {
   createEntityUrlObj,
   createFileObj,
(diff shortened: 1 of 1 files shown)


jilladams:
Question: this change will allow future accidental publishes of Facilities without a parent to go through without breaking the build, right?  Wondering how we can make sure we know about them, to mitigate, if the build passes correctly.

jilladams:
(Not saying we shouldn't merge this. Just flagging: do we need to ticket on Sitewide to add some sort of thing for our own awareness if/when this happens again? To your point in Slack: could we programmatically prevent merge without parent? Saying that here now, am I just doubling the Slack thread? SOrry for noise. I'll just stub that ticket for us.)

timcosgrove:
@jilladams to that point, I would say that this PR is here to prevent Content Release from breaking for all content because of this situation. The thing that this particular function takes care of - adding links from the parent to this child page - isn't critical to page functioning as near as I can tell. If having a page or two missing these links would be a critical problem for Veteran experience, let me know that.

Preventing the underlying situation from happening would require handling on the CMS side of things I think and is Facilities- specific. That would be up to Facilities to figure out how best to handle.

timcosgrove:
But yeah, to this point:
> Question: this change will allow future accidental publishes of Facilities without a parent to go through without breaking the build, right? Wondering how we can make sure we know about them, to mitigate, if the build passes correctly.

I think breaking the entire build in order to give a warning about a Facility going out without a parent is a really high price to pay.

jilladams:
Got it! Thank you. We stubbed https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20300 for prevention.

randimays (approved):

brianseek (approved):
LGTM, @timcosgrove lets sync later on the testing steps

eselkin (approved):
LGTM

PR #2434, opened 2025-02-03
remove old 5490 and 1990 unauth urls from registry.json

s-caso:
## Summary

- remove old 5490 and 1990 unauth urls from registry.json


rmessina1010 (approved):
LGTM

PR #2527, opened 2025-05-08
add preconnect links for API urls

acrollet:
## Summary

This PR adds a [preconnect hint](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preconnect) for the vets-api URL for each environment. This should improve application load times at the margin, as most apps need to fetch at least the user & feature toggles endpoints before rendering. Downsides are unlikely since this is a separate origin and won't block TCP connection slots for the revproxy and S3 origins.

Screenshot 2025-05-08 at 11 05 33

## Are you removing or changing a registry.json `entryName` in this PR?
- [x] No, I'm not changing any `entryName`s (skip to Summary and delete the rest of this section)

### DIFF
--- a/src/site/includes/header.html
+++ b/src/site/includes/header.html
@@ -22,6 +22,13 @@
 
   <!-- Preconnecting required origins: https://web.dev/uses-rel-preconnect/ -->
   <link rel="preconnect" href="https://www.googletagmanager.com" />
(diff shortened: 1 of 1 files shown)


rmessina1010 (approved):
LGTM

it-harrison (approved):
looks good

PR #2926, opened 2026-04-27
Update `form-data` to 2.5.4

pwolfert:
## Summary

Forces `form-data` package version to resolve to a newer version to resolve a security alert (https://github.com/department-of-veterans-affairs/content-build/security/dependabot/150). This is an indirect dependency coming from `cypress`.

## Related issue(s)


brianseek (approved):
👍

PR #2928, opened 2026-04-27
Fix one more use of removed `sinon` dependency

pwolfert:
## Summary

This is a follow-up to https://github.com/department-of-veterans-affairs/content-build/pull/2906. I guess we were using `sinon` in one other place.

## Related issue(s)

## Testing done

Run `yarn test` locally after `yarn install-safe`


brianseek (approved):
👍

PR #2934, opened 2026-04-29
Remove the `socks-proxy-agent` and its `ip` dependency

pwolfert:
## Summary

We were getting a security alert for `ip` ([97](https://github.com/department-of-veterans-affairs/content-build/security/dependabot/97), and it's only used by `socks-proxy-agent`. We do not use SOCKS anymore, so we can remove it.

## Related issue(s)


pwolfert:
Looks like the build on GHA still works. Marking it as ready for review.

edmund-dunn (approved):

